PR #140, opened 2023-05-09
avoid static imports in Java sample code

rjayasinghe:
in case this code is copied to an IDE the imports are missing and you have a bit of a challenge to find the right import for the static imports (VS Code Java extension does not support it at all at the moment).

Alternatively we could also use the new 'focus' feature for code blocks and include the static import statements as 'unfocused' but have them included in the clipboard after clicking the copy button.


renejeglinsky:
Can this be merged? I had the feeling it's not final, that's why I'm asking.

rjayasinghe:
Let's at least update with the base branch. @agoerler merge or talk? ;)

agoerler:
> @agoerler merge or talk? ;)

talk :-)

renejeglinsky:
We could do sth similar to what we've done in the getting started in a Nutshell. There we have the small snippets that is needed and in addition a detail box that can be expanded if you want the full code example incl. imports.

image

PR #157, opened 2023-05-17
Move content from notebook header into docs

mnkiefer:
- Moved Notebook header content to documentation, which includes
  - Warning that all executable cells run in the user's native environment
  - Tip to use Magic cell commands for common tasks


chgeo:
I don't think we should present warnings at that this place.  The user might just browse the tools page and this says _Please don't use notebooks, they are dangerous_ :)
If at all, we could add these disclaimers in the notebook files, and we already had that discussion.

PR #186, opened 2023-06-01
Don't promote `provisioning_parameters` in for `cds.xt.DeploymentService`

swaldmann:
This is a legacy setting, we should use the flatter `hdi.create` instead.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -9,7 +9,7 @@
   ],
   "packageRules": [
     {
-      "matchPackageNames": [ "com.sap.cds:cds-services-api" ],
+      "matchPackageNames": [ "com.sap.cds:cds-services-api", "com.sap.cds:cds4j-api" ],
       "rangeStrategy": "bump",
       "groupName": "CAP Java SDK"
     }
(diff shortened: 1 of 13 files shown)


gregorwolf:
Hi @swaldmann ,

does this also apply for:

```
"provisioning_parameters": { "database_id": "<HANA DB GUID>" }
```

In https://cap.cloud.sap/docs/guides/multitenancy/old-mtx-apis#subscribe-tenant and https://cap.cloud.sap/docs/releases/changelog/#jan-23-changed ? 

@renejeglinsky: Is there any documentation for MTXS regarding the provisioning_parameters? Can't find anything at https://cap.cloud.sap/docs/guides/multitenancy/mtxs.

swaldmann:
Hi @gregorwolf,

https://cap.cloud.sap/docs/guides/multitenancy/old-mtx-apis#subscribe-tenant is for the old MTX package (`@sap/cds-mtx`), where the legacy setting still has to be used.

https://cap.cloud.sap/docs/releases/changelog/#jan-23-changed shows how this configuration is now done in the new `@sap/cds-mtxs` package, but for compatibility both settings are supported.

In CAPire we only document the new way now:
https://cap.cloud.sap/docs/guides/multitenancy/mtxs#deployment-config

We chose to go with this naming rather than the `provisioning_parameters` way because
- "`provisioning_parameters`" comes from `@sap/instance-manager`, which we don't use any more in `@sap/cds-mtxs`
- calling the keys `create` and `bind` conveniently maps to `cf create` and `cf bind`

Best regards,
Steffen

PR #229, opened 2023-06-15
Order by alias example in Java section

vmikhailenko:
Simple example of ordering by alias for Java


java/query-api.md
```diff
@@ -689,6 +689,16 @@ Select.from("bookshop.Books")
...
 
+You can order by the alias of a column of the select list or a column that is defined as a result of the function call. 
+
+```java
+Select.from("bookshop.Person")
+    .columns(p -> p.get("name").toUpper().as("aliasForName"))
+    .orderBy(p -> p.get("aliasForName").asc());
+```
+
+Aliases of columns have precedence over the element names when `orderBy` is evaluated. Aliases may shadow the respective elements. To avoid that shadowing, don't use aliases that equal existing element names.
```
agoerler:
```suggestion
Aliases of columns have precedence over the element names when `orderBy` is evaluated. 

::: warning
Aliases may shadow elements names. To avoid shadowing, don't use element names as aliases.
::::
```

smahati (approved):

PR #285, opened 2023-07-10
Use deployable file structure and dependencies

simonoswald:
- When the model is not in `db` or `srv`, `cds-deploy` does not find it.
- When we don't include the `@cap-js/postgres`, we can't deploy to postgres


patricebender (approved):

simonoswald:
Checked with @danjoa: Seems fine, check again with the CSVs and if all works fine to merge.

simonoswald:
@renejeglinsky @manjuX I double checked by following the guide and doing a CF deployment, seems to work! Good to merge after your review :)

renejeglinsky:
Is needed by customers. Will merge and release capire afterwards.

PR #313, opened 2023-07-21
make <> more prominent

johannes-vogel:
as discussed yesterday


guides/databases-sqlite.md
```diff
@@ -657,7 +657,13 @@ SELECT.from('Foo')         //> [{ foo:1 }, ...]
 SELECT('bar').from('Foo')  //> ERROR: no columns to read
 ```
 
+### <> operator {.impl .node}
 
+Operator `<>` works as specified in SQL standard, `name != 'John'` translates to `name <> 'John' OR name is null`.
```
patricebender, line 662: I think this makes it a little bit clearer
```suggestion
Operator `<>` works as specified in SQL standard, while `name != 'John'` translates to `name <> 'John' OR name is null`.
```

stewsk (approved):

danjoa (approved):

smahati (approved):

PR #314, opened 2023-07-21
default pool config not applicable for @cap-js databases

sjvans:


johannes-vogel:
Shouldn't we align instead of documenting?

sjvans:
> Shouldn't we align instead of documenting?

i think @BobdenOs wanted to see if we can use a different pooling strategy in the @cap-js databases... either way is fine for me

PR #323, opened 2023-07-25
Unrelease services.md

### LINKED ISSUE #293: Duplicate documentation for custom service implementation?
Hi,

This morning I was searching for some guidance for the custom service implementation and found:

https://cap.cloud.sap/docs/node.js/services#srv-impl-how

which shows the 3 options (subclass, plain function, arrow function) as equal. But this afternoon I've navigated via https://cap.cloud.sap/docs/node.js/ and found:

https://cap.cloud.sap/docs/node.js/core-services#how-to-provide-custom-service-implementations

where the subclass based implementation has the preference. I think /docs/node.js/services is an artefact that should be checked for removal or?

Best Regards
Gregor

renejeglinsky:
Hi,

pls check #293 and if this PR solves the reported issue. If yes, we can in a next step remove the file.

Closes #293


danjoa (approved):
Yes pls remove

smahati (approved):

PR #330, opened 2023-07-27
Update domain-modeling.md

ChiaraDeDominicis-Reg:
Wrong entity name associated to key project in Many-to-Many Association. It should be Projects :-)


swaldmann (approved):

swaldmann:
Thanks a lot, Chiara!

PR #334, opened 2023-07-28
Improve search ranking for title matches

swaldmann:
- Removed some search-relevant stop words, otherwise "Node.js" or "Java" yield no search results.
  - I'd consider getting rid of stop words altogether, it's weird when for e.g. "Reuse and Compose" the "and" is not matched:
Screenshot 2023-07-28 at 11 22 58

- Boost title matches with a factor of 10000. You pretty much never want many full text matches to overtake a title match.
  - Example: searching for CDS, you ideally get the CDS overview page as a first result, but now you get many pages with a lot of "cds" in them. This usually just favors long guides. The VitePress default boost of 4 for title matches is far too low.
  - In a follow-up: we should also include a `level` field for some matches, to make sure the "Multitenancy" guide is coming in the results before "(1) Enable Multitenancy", for example. This requires some search index changes in VitePress, though. I'd consider writing our own search plugin (looks like just ~250 LoC).


chgeo:
> otherwise "Node.js" or "Java" yield no search results

In the context with more terms, like `draft Java` they make sense, yes.

> it's weird when for e.g. "Reuse and Compose" the "and" is not matched:

Yes, but w/o the `and` stop word, you still see it's not what you want, i.e. no exact match:
image

I think exact matches would be the proper solution here.  As long as they aren't around, I'd stick with `and` being ignored.  It just doesn't make sense to have it in the index.

> searching for CDS, you ideally get the CDS overview page as a first result, but now you get many pages with a lot of "cds" in them

Yes, makes sense.

> In a follow-up: we should also include a level field for some matches

This might not be needed as I think the level is already there.  Let me try this..

>  I'd consider writing our own search plugin (looks like just ~250 LoC).

With a UI, it's probably a lot more.  Plus the search it not extensible atm, so that youalso have to wire the new UI on your own.

swaldmann:
It looked like it's enough to provide an alternative plugin to https://github.com/vuejs/vitepress/blob/main/src/node/plugins/localSearchPlugin.ts as long as we make the generated index object compatible (with `title`, `titles`, and `text` fields). Then https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/components/VPLocalSearchBox.vue should be able to handle it and requires no modification. Something worth exploring after the release.

chgeo (approved):

swaldmann:
Hmm, I'm fine with merging this now because overall it's better, but I think we still haven't quite nailed it. 

For example, for "Multitenancy" you now still don't get the "Multitenancy" page, even if spelled out completely. In old Capire, you just have to type "Mu".

I'll see that I port the (generic) optimizations from old Capire here as well in a follow-up. Also had some ideas I didn't have time to implement that I'd give 1-2 days of deep work.
___


For reference, in development, I extensively tested those search terms for old Capire:
```
    Multitenancy
    xs-security
    CDS
    CSN
    ~/.cdsrc
    i18
    i18n
    cds watch
    alert('Test')
```

### Further ideas

- More semantically enriched search index 
  - `url` proved helpful in ranking search terms in old Capire
  - `java/node` sections for `text` attribute
    -  could enable automatic Node.js/Java switching
    - already mitigated by Java/Node.js downrate
  - auto-expansion of `details` blocks 
- It's probably better to create a total score instead of `return`ing early in the `boostDocument`
- Eval compression techniques for search index
  - Would mitigate larger search index → shouldn't be a concern w/ async loading & proper caching

### Observations

- VitePress seems to be doing search index generation on HTML instead of MD
  - Certainly also has pros though to generate build on rendered page (cf. Notebook generation)
  - But: lots of regexing on non-regular language → take with grain of salt

chgeo:
Yes, we should improve here.
I have created an issue for follow-ups: https://github.tools.sap/cap/docs/issues/306

PR #338, opened 2023-07-28
Potential fix for Resolve Binding in Hybrid Mode

finkmanAtSap:
If I understand the first sentence of the warning correctly, cloud bindings are only resolved by cds watch and cds env. This means, the seconds sentence is missing a "not". If I understood the warning wrong, could you please provide a better explanation what information the second sentence is intended to add?


swaldmann:
Hi @finkmanAtSap,

Your edit is correct for older versions of `cds-dk`, but recently we also added (optional) binding resolution for `cds run/serve`. So if you just remove the entire warning block we can merge it :)

Best regards,
Steffen

finkmanAtSap:
Hi Steffen,

Thanks for merging. Sorry I did not get to it before you did.

PR #342, opened 2023-07-31
Fiori Draft: add info on switching on fiori-draft-enabled

stewsk:
Can only be switched on if texts entities don't contain entries.


hjboth (commented):
maybe a word of wisdom why it only works as long as there are no records in `_texts`?

stewsk:
@hjboth : feel free to polish

hjboth (approved):
👍

smahati (approved):

PR #343, opened 2023-08-01
Calculated elements are not searchable by default

vmikhailenko:
In Java runtime, the calculated elements with "on-read" (more precisely, ones that annotated with `@Core.Computed` ) semantics are excluded from the search by default. 
This is the case for Node.js as well, but the condition is a little different. 

Java excludes the virtual elements from the search even if they are explicitly annotated. Should we document that as well?


guides/providing-services.md
```diff
@@ -449,6 +449,8 @@ entity Books { ... }
 
 Searches all elements of type `String` excluding the element `isbn`, which leaves the `title` and `descr` elements to be searched.
 
+You can extend the search to calculated elements which are not searchable by default.
```
agoerler, line 452:
```suggestion
You can extend the search to calculated elements, which are not searchable by default.
```

guides/providing-services.md
```diff
@@ -416,7 +416,7 @@ That would basically search for occurrences of `"Heights"` in all text fields of
 
 #### Using the `@cds.search` Annotation {#using-cds-search-annotation}
 
-By default all elements of type `String` of an entity are searched. Yet, sometimes you may want to deviate from this default and specify a different set of searchable elements, or to extend the search to associated entities. Use the `@cds.search` annotation to do so. The general usage is:
+By default all concrete elements of type `String` of an entity are searched. Yet, sometimes you may want to deviate from this default and specify a different set of searchable elements, or to extend the search to associated entities. Use the `@cds.search` annotation to do so. The general usage is:
```
agoerler, line 419:
```suggestion
By default only elements of type `String` of an entity are searched, provided they are not [calculated](../cds/cdl#calculated-elements) or [virtual](../cds/cdl#virtual-elements). Yet, sometimes you may want to deviate from this default and specify a different set of searchable elements, or to extend the search to associated entities. Use the `@cds.search` annotation to do so. The general usage is:
```

agoerler (approved):

agoerler:
@johannes-vogel - please review and check wether is also applies to CAP Node.js

danjoa (approved):
👍 looks good

agoerler (approved):

smahati (approved):

PR #345, opened 2023-08-01
Align code fences

swaldmann:
The predominant style seems to use the bare code fence language title for the segmented control.

I'd also be fine with the `Mac/Linux`, `Windows`, `Powershell` variant, but it should be aligned across CAPire.

@chgeo any opinion?


chgeo:
The bare language style is fine with me as well.

Let's wait for @renejeglinsky's feedback as he might remember why we have the Windows/Mac/Linux style.

renejeglinsky:
Hi,
I missed this one. Sorry!
I think we used it like this, because we did it the same way in the old capire. There it was just replaced automatically: https://github.tools.sap/cap/docs-old/blob/3e4722bd3d532b26f5b47ad3f248c896ed71f544/web/scripts.js#L154
And I think it looks nicer compared to sh, powershell and cmd.

renejeglinsky:
Hi,
I missed this one. Sorry!
I think we used it like this, because we did it the same way in the old capire. There it was just replaced automatically.
And I think it looks nicer compared to sh, powershell and cmd.

renejeglinsky (approved):

PR #351, opened 2023-08-03
Default pool config not applicable for @cap js databases

smahati:
carrying over #314 

@johannes-vogel asked "Shouldn't we align instead of documenting?"
we can work on this PR once we get more clarification.


sjvans (approved):
i'd publish now and remove again once we use the same defaults in @cap-js databases

danjoa:
> i'd publish now and remove again once we use the same defaults in @cap-js databases

How do we track this to fix asap?

sjvans:
> > i'd publish now and remove again once we use the same defaults in @cap-js databases
> 
> How do we track this to fix asap?

@BobdenOs is thinking about using a different approach to pooling. if we use generic pool there as well, then we'll use the same defaults. we have a BLI re hana connections in @cap-js impl.

PR #360, opened 2023-08-09
Fix markdownlinting errors

MatthiasAtSAP:
Adds some missing languages tags. 

Also found some markdown links without link value. For some I have found a matching link e.g.: `[cds. resolve()]()` &#8594; `[cds. resolve()](./cds-compile.md#cds-resolve)`, but for some there is no own section like `[tl;dr]()` in `about/glossary.md` or `[cds. load()]()` in `node.js/cds-facade.md`. In these cases, the link simply points to the current page.

### DIFF
--- a/about/glossary.md
+++ b/about/glossary.md
@@ -12,13 +12,13 @@ CAP-related Terms and Acronyms {.subtitle}
 <!-- {{$frontmatter?.synopsis}} -->
 <br>
 
-[CAP]() — shorthand for "SAP Cloud Application Programming Model"
+[CAP](./glossary.md) — shorthand for "SAP Cloud Application Programming Model"
 : Not an official name, though.
 
-[capire]() — Italian for 'understand'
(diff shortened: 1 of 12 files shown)


chgeo (approved):

about/glossary.md
```diff
@@ -12,13 +12,13 @@ CAP-related Terms and Acronyms {.subtitle}
 <!-- {{$frontmatter?.synopsis}} -->
 <br>
 
-[CAP]() — shorthand for "SAP Cloud Application Programming Model"
+[CAP](./glossary.md) — shorthand for "SAP Cloud Application Programming Model"
```
chgeo, line 15: Can we just use `.` here to denote the current page?  This way, we don't need to repeat the page name.

MatthiasAtSAP: `.` points to the current directory, so `[CAP](./glossary.md)` (located in `/about/glossary.md`) would refer to `/about` instead of `/about/glossary.md`

PR #363, opened 2023-08-14
Pure CSS solution for impl-variants

### LINKED ISSUE #296: Links with anchor do not lead to correct section
for example <https://cap.cloud.sap/docs/guides/databases#native-db-queries> does not lead me to the `native-db-queries` section, but to a section which is below the desired one. Tested on the newest Chrome Version as well as in Safari.

swaldmann:
Similar to `.dark`, just adding the language as a class to the `html` element. This is also how `vuejs.org` seems to do it.

### DIFF
--- a/.github/etc/test/impl-variants/test.md
+++ b/.github/etc/test/impl-variants/test.md
@@ -1,0 +1,44 @@
+---
+impl-variants: true
+outline: [2,4]
+---
+
+[[toc]]
+
+## H2 Node {.impl .node}
+
+### H3 Node 1
+
+<div>
+Text
+</div>
+
+### H3 Node 2
+
+Node text
+
+```sh
+code
+```
+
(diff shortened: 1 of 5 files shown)


danjoa (approved):

chgeo (commented):
Looks good and is a step forward.

But this doesn't solve our scrolling issue, right?
Our JS code is still is in conflict w/ the scroll positioning.

In [this Vue page](https://vuejs.org/guide/essentials/reactivity-fundamentals.html), you can see that they change the DOM, also have flickering effects, but their scroll positions are correct 🤔

swaldmann:
No, this doesn't fix the scroll position issue unfortunately, even with `onBeforeMount`. Will do some further investigation.

chgeo:
Fixes https://github.com/cap-js/docs/issues/472

PR #369, opened 2023-08-16
fixed after handler description in core-services.md

sebastianesch:
The documentation mentioned, that after handlers run *before* on handlers. *After* is correct.


swaldmann (approved):
Thanks @sebastianesch!

PR #377, opened 2023-08-22
Entity name as singular representation in domain modelling

narendran-nv:


renejeglinsky:
Hi @narendran-nv ,

thank you for your contribution.
In the [next section](https://cap.cloud.sap/docs/guides/domain-modeling#pluralize-entity-names), we recommend to use plural for entities. Your suggestion would be contrary to that. Could you explain the reasoning you had in mind?

All the best,
René

narendran-nv:
My bad you are right!

PR #380, opened 2023-08-23
Asynchronous event handlers for `served` and `shutdown`

chgeo:
Document the current implementation

Needed by `cds-plugin-ui5`(see cap/issues#14102)

### DIFF
--- a/node.js/cds-server.md
+++ b/node.js/cds-server.md
@@ -179,6 +179,7 @@ cds.on('served', (services)=>{
 })
 ```
 
+This event supports _asynchronous_ event handlers.
 
 
 ### listening {.event}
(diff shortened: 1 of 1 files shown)


danjoa (approved):
👍 super. Vielen Dank!

node.js/cds-server.md
```diff
@@ -191,10 +192,36 @@ A one-time event, emitted when the server has been started and is listening to i
 
 A one-time event, emitted when the server is closed and/or the process finishes.  Listeners can execute cleanup tasks.
 
+This event supports _asynchronous_ event handlers.
 
 
+### Event Handlers
```
renejeglinsky, line 198: Could this be a level two header? And could both headers be one line? Like: Sync vs. Async Event Handlers or Event Handler: Sync/Async

PR #388, opened 2023-08-30
FK constraint for .texts entities

stewsk:


guides/databases.md
```diff
@@ -642,9 +644,8 @@ No constraints are generated for...
 * Associations annotated with `@assert.integrity: false`
 * Associations which's source or target entity is annotated with `@cds.persistence.exists` or `@cds.persistence.skip`
 
-
-
-If the association is the backlink of a **composition**, the constraint's delete rule changes to `CASCADE`. For example that applies to the `parent` association in here:
+If the association is the backlink of a **composition**, the constraint's delete rule changes to `CASCADE`.
+For example that applies to the `parent` association in here:
```
renejeglinsky, line 648:
```suggestion
That applies, for example, to the `parent` association in here:
```

guides/databases.md
```diff
@@ -654,7 +655,35 @@ entity Genres {
 }
 ```
 
+As a special case, a referential constraint with `delete cascade` is also generated
+for text table of a [localized entity](../guides/localized-data#localized-data),
```
renejeglinsky, line 659:
```suggestion
for the text table of a [localized entity](../guides/localized-data#localized-data),
```

guides/databases.md
```diff
@@ -654,7 +655,35 @@ entity Genres {
...
+although no managed association is present in the `texts` entity.
+
+Add a localized element to entity `Books` in the example above:
```
renejeglinsky, line 662: Does the reference to the previous example really matter here? Seems self-contained to me.
```suggestion
Add a localized element to entity `Books` from the previous example:
```

```suggestion
Add a localized element to entity `Books`:
```

guides/databases.md
```diff
@@ -654,7 +655,35 @@ entity Genres {
 }
 ```
 
+As a special case, a referential constraint with `delete cascade` is also generated
+for the text table of a [localized entity](../guides/localized-data#localized-data),
+although no managed association is present in the `texts` entity.
+
+Add a localized element to entity `Books`:
```
stewsk, line 662:
```suggestion
Add a localized element to entity `Books` from the previous example:
```

stewsk:
> Does the reference to the previous example really matter here? Seems self-contained to me.

Usually the Books entity always has a localized element. So here I think it makes sense to refer to the example above where we first have defined Books w/o localized element.

stewsk:
ok for me

renejeglinsky (approved):

PR #389, opened 2023-08-30
No filter/sort for vitual elements

stewsk:


johannes-vogel:
The snippets looks like a bit out of context to the rest of the page. Is there also a preview available?

renejeglinsky (approved):

PR #390, opened 2023-08-30
Update example of crud style api

etimr:


johannes-vogel:
@etimr please also update the references a few lines above

johannes-vogel (approved):

renejeglinsky (approved):

PR #397, opened 2023-09-01
Release: Restore order in index page (#391)

renejeglinsky:
This got broken with 98e4566, so that during sorting, we compared paths with and without base paths.


smahati (approved):

renejeglinsky:
Hi @chgeo ,

could you have a look at the build issue? Seems related to the data.index.js file but I can't really tell what the issue is and why it appears in advanced and nowhere else. Probably breaks at first occurrence and stops then.

chgeo:
Your PR branch was not up to date with `cap.cloud.sap`, which caused an old Vitepress version to be used.  I also added the missing import of `join`, which probably got lost in a merge operation.

PR #411, opened 2023-09-06
Bump vitepress to rc-14

chgeo:


chgeo:
As discussed, let's merge this. Color changes are fine.

PR #412, opened 2023-09-06
Fix background color for brand-colored button

swaldmann:
Our current start page buttons look weird in light mode since the most recent VitePress color scheme changes:

Screenshot 2023-09-06 at 23 11 07

### DIFF
--- a/.vitepress/theme/custom.scss
+++ b/.vitepress/theme/custom.scss
@@ -8,6 +8,7 @@
     'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
   --vp-c-brand: #db8b0b;
   --vp-c-brand-dark: #9f660a;
+  --vp-button-brand-bg: #db8b0b;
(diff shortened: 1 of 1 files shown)


.vitepress/theme/custom.scss
```diff
@@ -8,6 +8,7 @@
     'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
   --vp-c-brand: #db8b0b;
   --vp-c-brand-dark: #9f660a;
+  --vp-button-brand-bg: #db8b0b;
```
swaldmann, line 11: btw I also tried less magic and removing all definitions except for `--vp-c-brand` but it still seems every color needs to be specified.

chgeo (changes requested):
This change becomes obsolete once we are on Vitepress rc-10, which changes its color system.
See PR https://github.com/cap-js/docs/pull/411 and its counterpart in cap/docs w/ a preview deployment.  There, I tried to apply our color theme, which makes the buttons a lot lighter (maybe too light).

PR #417, opened 2023-09-08
Detect missing container type

MatthiasAtSAP:
Detects when containers `:::`  are used without a type. The GH Bot tries to suggest a matching container type or `info` when it cannot suggest any matching type.


swaldmann (approved):
Nice! Thanks!

renejeglinsky (approved):

PR #429, opened 2023-09-19
cds.test: overhaul and warning for static file imports

johannes-vogel:
In order for plugins of the project to be detected with `cds.test`, one must use `cds.test.in` before any file import that might access `@sap/cds`. Any property access of `@sap/cds` might trigger `cds.env`. With `cds@7.3` the plugin loading will be more stable but `cds.test.in` must ensure that the plugins are loaded from the right directory.

Sample test:

```js
const cds = require('@sap/cds')
cds.test.in(__dirname + '/..')
cds.env

cds.test()

test('ensure that cds.env is not accessed too early', () => {
  expect(cds.env.requires.db.impl).toMatch(/@cap-js/)
})
```

### DIFF
--- a/node.js/cds-test.md
+++ b/node.js/cds-test.md
@@ -14,7 +14,7 @@ status: released
 
 The `cds.test` library provides best practice utils for writing tests for CAP Node.js applications.
 
-::: tip Find examples in [*cap/samples*](https://github.com/sap-samples/cloud-cap-samples/tree/master/test) and in the [*SFlight sample*](https://github.com/SAP-samples/cap-sflight/tree/main/test).
+::: tip Find examples in [*cap/samples*](https://github.com/sap-samples/cloud-cap-samples/tree/main/test) and in the [*SFlight sample*](https://github.com/SAP-samples/cap-sflight/tree/main/test).
 :::
 
 
@@ -442,7 +442,7 @@ class MyService extends cds.Service {} // [!code ++]
 
 :::warning Do switch on `CDS_TEST_ENV_CHECK` !
 
-It is strongly recommended to switch on `CDS_TEST_ENV_CHECK` in all your tests to detect such errors. It will likely become default in upcomming releases.
+It is strongly recommended to switch on `CDS_TEST_ENV_CHECK` in all your tests to detect such errors. It will likely become default in upcoming releases.
 
 :::
 
--- a/node.js/cds-test.md
+++ b/node.js/cds-test.md
@@ -56,12 +56,9 @@ cds.test.in(__dirname)
 For example, this would have `cds.env` loading the configuration from _package.json_ and _.cdsrc.json_ files found next to the test file, that is, in the same folder.
 
 ::: danger
-**Important:** Don't use `process.chdir()` in Jest tests, as they may leave test containers in screwed state, leading to failing subsequent tests.
-:::
-
-::: danger
-**Important:** If you also need to import other files that use `@sap/cds`, make sure that `cds.test.in(...)` is called before any **import**. This ensures a proper initialization of the test environment.
-In Typescript this requires use of dynamic imports.
+- **Don't use `process.chdir()`** in Jest tests, as they may leave test containers in screwed state, leading to failing subsequent tests.
+- Make sure that `cds.test.in(...)` is called **before any import** of files that use `@sap/cds`. This ensures a proper initialization of the test environment.
+In ESM/Typescript this requires use of dynamic imports.
 :::
 
 ::: tip
--- a/node.js/cds-test.md
+++ b/node.js/cds-test.md
@@ -1,106 +1,60 @@
 ---
-shorty: cds.test
-# synopsis: How test CAP Node.js applications.
 status: released
 ---
 
-# Testing
-
-Learn more about best practices to test CAP Node.js applications using the `cds.test` toolkit.
-Find samples for such tests in [cap/samples](https://github.com/sap-samples/cloud-cap-samples/tree/master/test) and in [CAP SFLIGHT app](https://github.com/SAP-samples/cap-sflight/tree/main/test).
-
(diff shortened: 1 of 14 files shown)


node.js/cds-test.md
```diff
@@ -59,6 +59,11 @@ For example, this would have `cds.env` loading the configuration from _package.j
 **Important:** Don't use `process.chdir()` in Jest tests, as they may leave test containers in screwed state, leading to failing subsequent tests.
 :::
 
+::: danger
```
chgeo, line 62:
```suggestion
::: warning
```

`danger` is a bit too dramatic here, `warning` is enough imho.

Is this then going to be resolved with 7.3?  If yes, we should add a note like _In sap/cds < 7.3, ..._

johannes-vogel: No, the problem persists in > 7.3. `cds.test.in` sets `cds.root`. If `cds.env` is accessed before, the plugins are not found because `cds.root` point to the directory where `jest/mocha` is started.

chgeo: I see.

chgeo:
I have combined the two _danger_ boxes into one, for better readability.

chgeo (approved):

node.js/cds-test.md
```diff
@@ -56,7 +56,9 @@ cds.test.in(__dirname)
 For example, this would have `cds.env` loading the configuration from _package.json_ and _.cdsrc.json_ files found next to the test file, that is, in the same folder.
 
 ::: danger
-**Important:** Don't use `process.chdir()` in Jest tests, as they may leave test containers in screwed state, leading to failing subsequent tests.
+- **Don't use `process.chdir()`** in Jest tests, as they may leave test containers in screwed state, leading to failing subsequent tests.
```
renejeglinsky, line 59: screwed combines mire states than just failed? Or can we say "in failed state"?

node.js/cds-test.md
```diff
@@ -56,7 +56,9 @@ cds.test.in(__dirname)
...
+- Make sure that `cds.test.in(...)` is called **before any import** of files that use `@sap/cds`. This ensures a proper initialization of the test environment.
```
renejeglinsky, line 60: This danger note is now a mix of Do and Don't. Would it be an option to put this information in the next Tip box, as this then relates and both are Do's?

node.js/cds-test.md
chgeo: > Or can we say "in failed state"?

Yes, or _unknown state_. or so.

node.js/cds-test.md
chgeo: What about rephrasing this so that it becomes a _don't_?
Like _Don't import files that use `@sap/cds` before `cds.test.in...`_

johannes-vogel: Should we also include the env variable for activating the  early `cds.env` access check?

chgeo: What does this variable do?  Don't remember it..

johannes-vogel: It'll activate a check that `cds.env` was accessed before `cds.test.in`.

chgeo: Ok.  Why not then mention it, as a troubleshooting means.

danjoa:
I gave the docs for cds.test() a major overhaul.

@swaldmann: what is this failing lint stuff about? → I've no clue how to fix that

swaldmann:
Hi @MatthiasAtSAP, can you check the failing linter job here?

@danjoa you can ignore it for now, looks more like a problem with the linter job itself. It's by design not a required check, so not blocking the PR.

renejeglinsky:
As this has grown now anyway, could we also include information about authorization and authentication? We have an issue asking for that.

danjoa:
> As this has grown now anyway, could we also include information about authorization and authentication? We have an issue asking for that.

That's a separate topic I think

danjoa (approved):

node.js/cds-test.md
```diff
@@ -203,185 +139,409 @@ Similarly, you can use other test watchers like `mocha -w`.
...
 
-Provides utilities to manage test data: {.indent}
-- `.autoReset (boolean)` enables automatic deletion and redeployment of CSV data before each test. Default is `false`.
-- `.delete ()` deletes data in all database tables
-- `.reset ()` deletes data in all database tables and deploys CSV data again {.indent}
+To fix this, always ensure your calls to `cds.test.in(folder)` or `cds.test(folder)` goes first, before anything else loading `cds.env`:
 
-#### `.in (...paths)` → `Test` { #in}
+```js
+cds.test(__dirname) //> always should go first
```
johannes-vogel, line 437: Shouldn't this be `cds.test.in` as `cds.test` will only start the server in `beforeAll` but the `cds.env` access happens before?

chgeo (commented):
I noticed that `data.delete`, `data.autoreset` and `test.verbose` are not mentioned anymore?  Is this by intention?

node.js/cds-test.md
renejeglinsky: @danjoa ?

renejeglinsky:
> I noticed that `data.delete`, `data.autoreset` and `test.verbose` are not mentioned anymore? Is this by intention?

@danjoa ?

chgeo:
> I noticed that `data.delete`, `data.autoreset` and `test.verbose` are not mentioned anymore? Is this by intention?

Yes, by intent.
- `data.delete/.autoreset` -> apps shall decided on whether to reset to avoid performance issues
- `test.verbose` -> `test.log`

renejeglinsky:
So can we merge? Pipelines are red, as one link has been "fixed" in main that is waiting for this PR.

chgeo:
Yes, should merge..

renejeglinsky (approved):
Will do my review later

PR #438, opened 2023-09-25
Java db guide

renejeglinsky:
Closes https://github.com/cap-js/docs/issues/437

### DIFF
--- a/guides/databases.md
+++ b/guides/databases.md
@@ -348,17 +348,29 @@ db.queryForList("SELECT from sqlite_schema where name like ?", name);
 
 When you run your server with `cds watch` during development, an in-memory database is bootstrapped automatically, with SQL DDL statements generated based on your CDS models.
 
+You can also do this manually with the CLI command `cds compile --to <dialect>`.
+
 </div>
 
 <div markdown="1" class="impl java">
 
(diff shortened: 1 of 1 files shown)


guides/databases.md
```diff
@@ -403,15 +406,97 @@ service CatalogService {
...
+
+--- some more technical views skipped ...
+```
+
+:::
+
+</div>
+
+
+<div class="impl java">
```
renejeglinsky, line 473: @agoerler Could you please check:
- The generated schema if that is valid for H2 or if not, add a proper example?
- If in this context here generating `schema.sql` is sufficient? In the Java persistence guide it's amore specific place for the actual deployment. I understand it here as "playground" and there kept project root as the place for this file.

guides/databases.md
```diff
@@ -403,15 +406,97 @@ service CatalogService {
 
 Generate an SQL DDL script by running this in the root directory containing both *.cds* files:
 
+<div class="impl node">
+
+
+```sh
+cds compile srv/cat-service --to sql --dialect sqlite > schema.sql
```
renejeglinsky, line 413: Hi @swaldmann ,
Are there upcoming changes wrt to this? If not, I'd merge for the upcoming release.

guides/databases.md
renejeglinsky: Hi @agoerler ,

could you please have a look?

guides/databases.md
swaldmann: Not yet, should be fine

guides/databases.md
renejeglinsky: @MattSchur Could you pls check if the example is correct?

PR #439, opened 2023-09-25
Described pseudo role internal-user

BraunMatthias:


johannes-vogel (commented):
looks good for general internal-user docs.
the java part I can't judge.

guides/authorization.md
```diff
@@ -99,15 +99,29 @@ It’s frequently required to define access rules that aren’t based on an appl
 
 The following predefined pseudo roles are currently supported by CAP:
 
-* `authenticated-user` refers to (named or unnamed) users who have presented a valid authentication claim such as a logon token.
+* `authenticated-user` refers to named or unnamed users who have presented a valid authentication claim such as a logon token.
 * `system-user` denotes an unnamed user used for technical communication.
```
renejeglinsky, line 103:
```suggestion
* [`system-user` denotes an unnamed user used for technical communication.](#system-user)
```

guides/authorization.md
```diff
@@ -99,15 +99,29 @@ It’s frequently required to define access rules that aren’t based on an appl
...
+* `internal-user` is dedicated to distinguish application internal communication (see [application zone](../guides/security/overview#application-zone).
```
renejeglinsky, line 104:
```suggestion
* [`internal-user` is dedicated to distinguish application internal communication.](#internal-user)
```

guides/authorization.md
```diff
@@ -99,15 +99,29 @@ It’s frequently required to define access rules that aren’t based on an appl
...
 ::: tip
 For XSUAA or IAS authentication, the request user is attached with the pseudo role `system-user` if the presented JWT token has been issued with grant type `client_credentials` or `client_x509` for a trusted client application.
 :::
 
+#### internal-user
+Pseudo-role `internal-user` allows to define application endpoints that can be accessed exclusively by the own PaaS tenant (technical communication). The advantage is that similar to `system-user` no technical CAP roles need to be defined to protect such internal endpoints. However, in contrast to `system-user`, the endpoints protected by this pseudo-role do not allow requests from any external technical clients. Hence is suitable for **technical intra-application communication**.
```
renejeglinsky, line 114:
```suggestion
Pseudo-role `internal-user` allows to define application endpoints that can be accessed exclusively by the own PaaS tenant (technical communication). The advantage is that similar to `system-user` no technical CAP roles need to be defined to protect such internal endpoints. However, in contrast to `system-user`, the endpoints protected by this pseudo-role do not allow requests from any external technical clients. Hence is suitable for **technical intra-application communication**, see [Security > Application Zone](../guides/security/overview#application-zone). 
```

guides/authorization.md
BraunMatthias: should the full sentence be a hyperlink?

renejeglinsky: Yes. Because having it only on the code doesn't make sense and I want it to be the same or similar for internal-user. Haven't found a good way to select text in that regard. Therefore: full sentence.

java/architecture.md
```diff
@@ -147,57 +152,100 @@ The dependencies of a Spring Boot application with OData V4 endpoints could look
...
 Choosing a feature by adding the Maven dependency *at compile time* enables the application to make use of the feature *at runtime*. If a chosen feature misses the required environment at runtime, the feature won't be activated. Together with the fact that all features have a built-in default implementation ready for local usage, you can run the application locally with the same set of dependencies as for productive mode.
 For instance, the authentication feature `cds-feature-hana` requires a valid `hana`-binding in the environment. Hence, during local development without this binding, this feature gets deactivated and the stack falls back to default feature adapted for SQLite.
 
-### Available Modules
+### CAP Java Standard Modules
 
-Following application frameworks are available:
+CAP Java comes with a rich set of prepared modules in all different layers of the stack:
+
+#### Application Frameworks
```
renejeglinsky, line 192: I don't think we need those headers. If possible, we want to only use up to level 3 headers, as we anyways only show those headers in our ToCs. If it's about emphasizing, let's make them bold.

BraunMatthias: let's adress this in a different PR

java/architecture.md
```diff
@@ -6,7 +6,7 @@ status: released
 uacp: Used as link target from Help Portal at https://help.sap.com/products/BTP/65de2977205c403bbc107264b8eccf4b/9186ed9ab00842e1a31309ff1be38792.html
 ---
 
-# Stack Architecture
```
renejeglinsky, line 9: The menu entry would then also be Modular Architecture, right?

java/architecture.md
```diff
@@ -17,7 +17,7 @@ One of the key [CAP design principles](../about/#open-and-opinionated) is to be
 The [modular architecture](#modular_architecture) reflects this requirement, allowing fine-grained [stack configuration](#stack_configuration) and custom extensions.
```
renejeglinsky, line 17:
```suggestion
The [stack architecture](#modular_architecture) reflects this requirement, allowing fine-grained [stack configuration](#stack_configuration) and custom extensions.
```

java/architecture.md
```diff
@@ -17,7 +17,7 @@ One of the key [CAP design principles](../about/#open-and-opinionated) is to be
...
 
 
-## Modular Architecture { #modular_architecture}
+## Stack Architecture { #modular_architecture}
 
 ### Architecture Overview
```
renejeglinsky, line 22:
```suggestion
### Overview
```

java/architecture.md
renejeglinsky: See https://github.com/cap-js/docs/pull/439/commits/f1eb2930f9b2629c1116eeec814d67981ba356aa

PR #443, opened 2023-09-26
Default for managed associations

stewsk:
Failure of snippet checker is due to using old compiler version. Fixed with https://github.com/cap-js/docs/pull/458


stewsk:
Hi @bugwelle , snippet checker fails -> doesn't like the default value. Is there something that needs to be updated?

bugwelle:
Requires https://github.com/cap-js/docs/pull/458 :-)

renejeglinsky (approved):

PR #450, opened 2023-09-27
Fix link styling for semantic boxes

swaldmann:


swaldmann:
Seems fixed in the meantime

PR #457, opened 2023-09-29
Expose assoc filter

stewsk:


cds/cdl.md
```diff
@@ -756,6 +756,71 @@ GET /Teams?$expand=members($expand=user)
...
+As associations are first class citizens, you can put them into the select list
+of a view or projection ("publish") like regular elements. A `select *` includes all associations.
+If you need to rename an association, you can provide an alias.
+
+Example:
+```cds
+entity P_Employees as projection on Employees {
+  ID,
+  addresses
+}
```
bugwelle, line 770:
```suggestion
  addresses,
};
```

cds/cdl.md
```diff
@@ -756,6 +756,71 @@ GET /Teams?$expand=members($expand=user)
...
+
+#### Publish Associations with Filter (beta) {#publish-associations-with-filter}
+
+::: warning
+This is a beta feature. Beta features aren't part of the officially delivered scope that SAP guarantees for future releases.
+For more information, see [Important Disclaimers and Legal Information](https://help.sap.com/viewer/disclaimer).
+:::
+
+When publishing an unmanaged association in a view or projection, you can add a filter condition.
+The ON condition of the resulting association is the ON condition of the original
```
bugwelle, line 784:
```suggestion
The ON condition of the resulting association is the ON condition of the original
```

@renejeglinsky We use `ON condition` in capire, and `ON-condition` in the compiler. Any guidance from your side which we should prefer? :)

cds/cdl.md
```diff
@@ -756,6 +756,71 @@ GET /Teams?$expand=members($expand=user)
...
+```
+
+An association that has been published with a filter is read-only. It must not be
+used to modify the target entity.
+
+Filters usually are provided only for to-many associations, which usually are unmanaged.
+Thus publishing with a filter is almost exclusively used for unmanaged associations.
+Nevertheless you can also publish a managed association with a filter. This will automatically
+turn the resulting association into an unmanaged one. You must ensure that all foreign key elements
+needed for the ON condition are explicitly published.
```
bugwelle, line 813: Well, the same applies to ON-conditions as well. But being explicit is fine here, I think :)

cds/cdl.md
```diff
@@ -756,6 +756,71 @@ GET /Teams?$expand=members($expand=user)
...
+
+Example:
+```cds
+entity P_Books as projection on Books {
+  author.ID as authorID,  // needed for ON condition of deadAuthor
+  author[dateOfDeath < $now] as deadAuthor  // -> unmanaged association
```
bugwelle, line 819: Just from a modeling perspective, I wonder what happens if `dateOfDeath` is `null`. 😄 

```suggestion
  author[dateOfDeath is not null] as deadAuthor  // -> unmanaged association
```

bugwelle (approved):

cds/cdl.md
stewsk: smart!

cds/cdl.md
stewsk: It looks nicer w/o the `,` and `;`

bugwelle: Trailing commas are a good practice to avoid git diffs on changes. But I agree for this small example, non are better. :)

cds/cdl.md
renejeglinsky: Is `ON-condition` a term? I always felt it's similar to `WHERE clause` and other constructs like that. Therefore I went with `ON condition` (could be another condition as well). Or is this a term we need to define for us?
What is your take?

stewsk: From my POV it is similar to "WHERE condition"

renejeglinsky (approved):

cds/cdl.md
bugwelle: The same for me. Similar to SQL's `JOIN … ON`-condition, but for associations.

renejeglinsky: So we keep it as is :)

PR #460, opened 2023-09-29
node.js: leading slash in the path argument of rest-style api

etimr:


node.js/core-services.md
```diff
@@ -1125,7 +1125,7 @@ srv.patch('/Books',...)   -->  srv.send('PATCH','/Books',...)
 srv.delete('/Books',...)  -->  srv.send('DELETE','/Books',...)
 ```
 
-You can also use them as REST-style variants to run queries by omitting the leading slash in the `path` argument, or by passing a reflected entity definition instead. In that case they start constructing *bound* [`cds.ql` query objects](cds-ql), as their [CRUD-style counterparts](#crud-style-api):
+Leading slash in the `path` argument results in the same behaviour as in `srv.send()`: `path` is sent to a service unmodified. But you can also use these methods as REST-style variants to run queries by omitting the leading slash in the `path` argument, or by passing a reflected entity definition instead. In that case they start constructing *bound* [`cds.ql` query objects](cds-ql), as their [CRUD-style counterparts](#crud-style-api):
```
johannes-vogel, line 1128: I think this paragraph is overly complex. Maybe we can simplify but I don't have a good idea atm....

etimr: @johannes-vogel could you please check the simplified variant?

node.js/core-services.md
```diff
@@ -1125,7 +1125,7 @@ srv.patch('/Books',...)   -->  srv.send('PATCH','/Books',...)
 srv.delete('/Books',...)  -->  srv.send('DELETE','/Books',...)
 ```
 
-You can also use them as REST-style variants to run queries by omitting the leading slash in the `path` argument, or by passing a reflected entity definition instead. In that case they start constructing *bound* [`cds.ql` query objects](cds-ql), as their [CRUD-style counterparts](#crud-style-api):
+Leading slash in the `path` argument results in the same behaviour as in `srv.send()`: `path` is sent to a service unmodified. Omitting the leading slash or passing a reflected entity definition instead, leads to construction of *bound* [`cds.ql` query objects](cds-ql), equivalent to [CRUD-style API](#crud-style-api):
```
johannes-vogel, line 1128:
```suggestion
Leading slash in the `path` argument results in the same behaviour as in `srv.send()`: `path` is sent to a service unmodified. Omitting the leading slash or passing a reflected entity definition instead, constructs *bound* [`cds.ql` query objects](cds-ql), equivalent to [CRUD-style API](#crud-style-api):
```

johannes-vogel: @renejeglinsky could you please also double check the wording?

node.js/core-services.md
```diff
@@ -784,7 +784,7 @@ return : result of this.dispatch(req)
 Use this method to send synchronous requests to a service for execution.
 
 -  `method` can be a HTTP method, or a name of a custom action or function
```
renejeglinsky, line 786:
```suggestion
-  `method` can be an HTTP method, or a name of a custom action or function
```

node.js/core-services.md
```diff
@@ -784,7 +784,7 @@ return : result of this.dispatch(req)
...
--  `path` can be an arbitrary URL, starting with a leading `'/'`
+-  `path` can be an arbitrary URL, starting with a leading `'/'`, it is passed to a service without any modification as a string.
```
renejeglinsky, line 787:
```suggestion
-  `path` can be an arbitrary URL, starting with a leading `'/'`, it is passed to a service without any modification as a string
```

node.js/core-services.md
```diff
@@ -1125,7 +1125,7 @@ srv.patch('/Books',...)   -->  srv.send('PATCH','/Books',...)
 srv.delete('/Books',...)  -->  srv.send('DELETE','/Books',...)
 ```
 
-You can also use them as REST-style variants to run queries by omitting the leading slash in the `path` argument, or by passing a reflected entity definition instead. In that case they start constructing *bound* [`cds.ql` query objects](cds-ql), as their [CRUD-style counterparts](#crud-style-api):
+Leading slash in the `path` argument results in the same behaviour as in `srv.send()`: `path` is sent to a service unmodified. Omitting the leading slash or passing a reflected entity definition instead, constructs *bound* [`cds.ql` query objects](cds-ql), equivalent to [CRUD-style API](#crud-style-api):
```
renejeglinsky, line 1128:
```suggestion
Leading slash in the `path` argument results in the same behaviour as in `srv.send()`: `path` is sent unmodified to a service. Omitting the leading slash, or passing a reflected entity definition instead, constructs *bound* [`cds.ql` query objects](cds-ql), equivalent to [CRUD-style API](#crud-style-api):
```

renejeglinsky (approved):

PR #473, opened 2023-10-06
install dependencies after cds init

njames:
This needs some more thought but if you don't add this step when you start on 2. capture domain models your terminal will go berserk with errors and it is because the db dependencies are not yet installed. 

I don't know java developers need to do here /shrug

### DIFF
--- a/get-started/in-a-nutshell.md
+++ b/get-started/in-a-nutshell.md
@@ -73,6 +73,23 @@ Users on macOS must first run a command (*Shell Command: Install 'code' command 
 
 :::
 
+2.5 Initialise dependencies in an [*Integrated Terminal*](https://code.visualstudio.com/docs/terminal/basics) 
+
+Install all the dependencies into your system. 
+
+::: code-group
+
+```sh [Node.js]
+npm install
+```
+
+```sh [Java]
+/shrug
+```
+
+:::
+
+
 3. Run `cds watch` in an [*Integrated Terminal*](https://code.visualstudio.com/docs/terminal/basics)
 
 ::: code-group


renejeglinsky:
Hi Nigel,
that should work and for me it does work w/o any issues. Not sure what causes the issue on your end. Could you share more about your setup?

njames:
So if I start at https://cap.cloud.sap/docs/get-started/in-a-nutshell#start-a-project
running down that page

1. `cds init lookshop`
2. `code lookshop` (I check the code on cli for mac)
3. `cds watch` -> waiting for models etc
4. Add ./db/schema.db file from docs

```
❗️ ERROR on server start: ❗️

 Error: Configuration for 'cds.requires.db' lacks mandatory property 'kind' or 'impl'
    at options4 (/Users/nigeljames/.nvm/versions/node/v18.17.0/lib/node_modules/@sap/cds-dk/node_modules/@sap/cds/lib/srv/cds-connect.js:59:50)
```
So I kill the watch and run  `npm i` and start the watch `cds watch` 
everything is as expected

I ran this on both my mac and linux machine and same result. 
I did check `cds -v` on both to ensure they were the same (or close)

gregorwolf:
Is @sap/cds-dk installed globally?

njames:
Yes 
```
cds -v
@cap-js/sqlite: 1.3.0
@sap/cds: 7.3.0
@sap/cds-compiler: 4.3.0
**@sap/cds-dk (global): 7.2.0**
@sap/cds-fiori: 1.1.0
@sap/cds-foss: 4.0.2
@sap/cds-mtxs: 1.11.0
@sap/eslint-plugin-cds: 2.6.3
Node.js: v18.17.0
home: .../Code/Cap/lookshop/node_modules/@sap/cds
lookshop: 1.0.0
```

njames:
Is the expectation that `cds watch` will run `npm install` ? 

I would assume then that after running `cds watch` you get a `node_modules` folder ?

gregorwolf:
What Node.JS Version do you use @njames ? I wasn't able to replicate the issue on my Mac with Node 18.

njames:
I swear I did this yesterday twice and last week as well. The only think I can think that happened was when I opened code is I was in the wrong folder when running cds watch. 

I have just rerun it this morning and it worked fine. 

/shrug Thanks for your comments.

njames:
Then as soon as I added a simple model I got the same error:

```
using { Currency, managed, sap } from '@sap/cds/common';

namespace Bookshop;

entity Books : managed {
    key ID : Integer;
    title : localized String(100);
 }
 ```
 


```  db/schema.cds
  ../../../.nvm/versions/node/v18.17.0/lib/node_modules/@sap/cds-dk/node_modules/@sap/cds/common.cds


❗️ ERROR on server start: ❗️

 Error: Configuration for 'cds.requires.db' lacks mandatory property 'kind' or 'impl'
```

so I run npm i and re- watch and it is as I expect. 

this is just weird that none of you are getting this.

renejeglinsky:
I haven't seen such an error. AFAIK there's always a default for cds.requires.db. What's the output of `cds env requires.db`?

renejeglinsky:
Ok, w/o `npm i` my cds.env.requires.db is also undefined but it works. /shrug here as well.

I see that you're using cds-dk@7.2.0 and already cds@7.3.0 - could you update cds-dk to 7.3.0 as well and check again?

PR #498, opened 2023-10-24
Advanced > OData: remove valuelist from snippet

renejeglinsky:


advanced/odata.md
```diff
@@ -216,10 +216,6 @@ OData defines a strict two-fold key structure composed of `@<Vocabulary>.<Term>`
 
 ```cds
 @Common.Label: 'Customer'
-@Common.ValueList: {
```
renejeglinsky, line 219: Or should this be replaced by sth else?

ralfhandl: Do you want
1. a structured annotation on an entity, or 
2. a correct ValueList example?

If 1, then use @UI.HeaderInfo, see for example https://pages.github.tools.sap/cap/docs/guides/extensibility/customization#extending-ui-annotations

stewsk:
Hi @Rene, what is the purpose of this chang?

renejeglinsky:
> Hi @rene, what is the purpose of this chang?

I sent you an email.

stewsk:
I think we want an example for a structured annotation at that place.
Have added an example with `@UI.Header` as proposed by @ralfhandl

ralfhandl (approved):

PR #499, opened 2023-10-24
@cds.collate : false results in case-sensitive order

MattSchur:


agoerler (approved):

java/persistence-services.md
```diff
@@ -37,9 +37,12 @@ entity Books : cuid {
     title        : localized String(111);
     descr        : localized String(1111);
     @cds.collate : false // [!code focus]
-    isbn         : String(40);  // does not require locale-specific sorting // [!code focus]
+    isbn         : String(40);  // does not require locale-specific handling // [!code focus]
 }
 ```
+::: warning
+When disabling locale-specific handling for a String element, binary comparison is used, which is generally faster but results in *case-sensitive* order (A, B, a, b).
+:::
```
renejeglinsky, line 45: To avoid multiple notes...people reading this chapter will catch it w/o the visual appearance of a warning.
```suggestion
> When disabling locale-specific handling for a String element, binary comparison is used, which is generally faster but results in *case-sensitive* order (A, B, a, b).
```

renejeglinsky (approved):

PR #502, opened 2023-10-26
Correct default cardinality for a managed composition of aspect

hjboth:
If no cardinality is specified the default is 'to-one' and to be more precisely [0..1], as the header may have no items but an item must have exactly one header [1..1]. This is expressed with the `min`/`max` properties on the `up_` association.


bugwelle (approved):

cds/cdl.md
```diff
@@ -722,7 +722,7 @@ aspect OrderItems {
 
 #### Default Target Cardinality
 
-If not otherwise specified, a managed composition of an aspect has the default target cardinality *to many*.
+If not otherwise specified, a managed composition of an aspect has the default target cardinality *of one*.
```
agoerler, line 725:
```suggestion
If not otherwise specified, a managed composition of an aspect has the default target cardinality *to-one*.
```

( I think we use to-one, to-many, many-to-many in [other places](https://cap.cloud.sap/docs/cds/cdl#managed-associations) in Capire)

renejeglinsky: I think `of one` is fine as it relates to the modeling, as in  `composition of many`. @hjboth What is your take? Should it be aligned with the code or with general terms we use to express relationships? Is there (Can there be) a difference between expressing a relationship (to-one) and describing a cardinality (of one)?

hjboth: Well, we basically have these options:

1a) whenever the cardinality refers to a composition, call it to `of-...`, BUT, this reads horribly if the source cardinality comes into play `one-of-many` or `many-of-many` is something I have never heard before.
1b) whenever the cardinality refers to an association, call it `...-to-...`
2) Be consistent and use `...to...` for all relations everywhere irrespective of their semantics and its actual CDL syntax.

 I changed 'to-one' into `of-one` because Evan as a native speaker suggested this. In general I prefer not to irritate users and long time readers with changed subtleties, so I vote for (2).

agoerler: 🤔

> 1a) whenever the cardinality refers to a composition, call it to of-..., BUT, this reads horribly if the source cardinality comes into play one-of-many or many-of-many is something I have never heard before.

+1

I would say:

"A composition of one aspect has a cardinality of one-to-one".

or 

"A composition of one aspect has the cardinality one-to-one".

likewise 

"An association to one entity has the cardinality one-to-one or many-to-one".

or, shorter

"An association to one entity has the cardinality to-one".

renejeglinsky: Following the discussion I like Adrian's proposal, which is in line with Hans-Jörg.
For me it boils down to the difference between `to one` and `to-one`. So, everytime we talk about cardinalities we use `to-one`. Maybe language-wise it's not the cleanest solution to our question, but it's understood by SMEs and that's what counts here most for me.

PR #524, opened 2023-11-07
Java plugin docu

rjayasinghe:


java/plugin_concept.md
```diff
@@ -0,0 +1,260 @@
...
+
+[Learn more about providing and using reuse packages.](../guides/extensibility/composition){.learn-more}
+
+This technique can be used indepently or together with one or more of the techniques described on this page.
+
+## Event Handlers for custom types and annotations
+
+Besides extending the capabilities for inbound requests it is also possible to extend the behaviour of event handling during request handling of the CAP Java runtime. Meaning that you can define custom handlers that react on model characteristics (common types or annotations) or also on entity values e.g. validations.
+
+Inside your reuse module you can define a custom event handler and and a registration hook as plain Java code. Once this module is addded to any CAP Java application as a dependency the contained event handler code will be active automatically.
```
dimamost, line 84: typo  -> and and

java/plugin_concept.md
```diff
@@ -0,0 +1,260 @@
...
+      // any custom Java code using the event context and CQL APIs
+    }
+}
+```
+
+The shown handler code is registered for any entity type on any [ApplicationService](../guides/providing-services). Dependending on the use case the target scope could narrowed to specific entities and/or services. The handler registration applies to the same rules as custom handlers that are directly packaged with a CAP Java application.
+
+The real difference to your typical event handler in your application code is the actual handler reigstration code. Although you could use the same `@Componenent` Spring Framework mechanism this is not recommendable because your handler code would then be dependent on Spring and not only on CAP Java. Your code would need to maintain it's own versioned dependency to the Spring Framework or Spring Boot and would need to react to changes in the given frameworks.
+
+So, in order to have a framework independent handler registration the `` interface needs to be implemented like this:
```
dimamost, line 103: Which interface? -> ''

rjayasinghe: fixed

java/plugin_concept.md
```diff
@@ -0,0 +1,260 @@
...
+
+
+
+[Learn more about event handling in our EventHandler documentation](provisioning-api){.learn-more}
+
+A complete end-to-end example for reusable event handlers can be found in this [blog post](https://blogs.sap.com/2023/05/16/how-to-build-reusable-plugin-components-for-cap-java-applications/).
+
+## Custom Protocol Adapters
+
+In CAP Java, the protocol adapter is the mechanism to implement inbound communication (another service or the UI) to the CAP service in development. The task of a protocol adapter is to translate any incoming requests of a defined protocol to CQL statements that then can be executed on locally defined CDS services. CAP Java comes with 3 protocol adapters (OData V2 and V4, and HCQL) but can be extended with custom implementations. In this section we'll have a deeper look on how such a protocol adapter can be built and registered with the CAP Java runtime.
```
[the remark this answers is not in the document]
rjayasinghe: HCQL is a valid abbreviation in CAP Java

java/plugin_concept.md
```diff
@@ -0,0 +1,260 @@
...
+
+```cds
+using { CatalogService } from 'com.sap.capire/bookshop';
+```
+
+> Note that [CDS editor](../tools/#cds-editor) does not yet support this new location and hence shows an error marker for this line. This will be fixed soon.
+
+[Learn more about providing and using reuse packages.](../guides/extensibility/composition){.learn-more}
+
+This technique can be used indepently or together with one or more of the techniques described on this page.
```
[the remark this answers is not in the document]
rjayasinghe: fixed

dimamost (approved):

rjayasinghe:
@renejeglinsky all checks passed, a menu entry is created and we have approval from the Java side.

This guide is not documenting a feature from the recent release but concepts and extension points that have been part of CAP Java for a while now.

java/plugin_concept.md
```diff
@@ -0,0 +1,261 @@
...
+
+> Note that the location syntax differs from the usual syntax and adheres to the groupId/artifactId syntax defined above. Also, the [CDS editor](../tools/#cds-editor) does not yet support this new location and hence shows an error marker for this line. This will be fixed soon.
+
+[Learn more about providing and using reuse packages.](../guides/extensibility/composition){.learn-more}
+
+This technique can be used independently or together with one or more of the techniques described on this page.
+
+## Event Handlers for custom types and annotations
+
+Besides extending the capabilities for inbound requests it is also possible to extend the behaviour of event handling during request handling of the CAP Java runtime. Meaning that you can define custom handlers that react on model characteristics (common types or annotations) or also on entity values e.g. validations.
```
beckermarc, line 83: Where is the difference between "capabilities of inbound requests" and "behaviour of event handling during request handling"?

java/plugin_concept.md
```diff
@@ -0,0 +1,261 @@
...
+    @After
+    public void handleSample(CdsReadEventContext context) {
+      // any custom Java code using the event context and CQL APIs
+    }
+}
+```
+
+The shown handler code is registered for any entity type on any [ApplicationService](../guides/providing-services). Dependending on the use case the target scope could narrowed to specific entities and/or services. The handler registration applies to the same rules as custom handlers that are directly packaged with a CAP Java application.
+
+The real difference to your typical event handler in your application code is the actual handler registration code. Although you could use the same `@Componenent` Spring Framework mechanism this is not recommendable because your handler code would then be dependent on Spring and not only on CAP Java. Your code would need to maintain it's own versioned dependency to the Spring Framework or Spring Boot and would need to react to changes in the given frameworks.
```
beckermarc, line 102: We should probably explain both approaches: The Spring approach is valid, if you need integration with Spring, e.g. due to dependency injection of some other beans provided by another common Spring-based library. This is a common approach and we are often asked to do the auto-configuration in that case (people don't know about Spring Boots auto-configuration mechanisms and usually fallback to extending the classpath scan in the applications).

java/plugin_concept.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+With the factory in place we can start to build the actual protocol adapter. As mentioned before, most adapters will implement HTTP connectivity and will therefore be an extension of the Jakarta `HttpServlet` class. Based on the incoming request path the protocol adapter needs to determine the corresponding CDS `ApplicationService`. Parts of the request path together with potential request parameters (this depends on the protocol to be implemented) then need to be mapped to a CQL statement which is then executed on the previously selected CDS `ApplicationService`.
+
+```java
+public class SampleAdapter extends HttpServlet {
+
+	private final CdsRuntime runtime;
+
+	public CdsHcqlServlet(CdsRuntime runtime) {
```
beckermarc, line 214:
```suggestion
	public SampleAdapter(CdsRuntime runtime) {
```

beckermarc (commented):
Some additional comments from my side :)

java/plugin_concept.md
```diff
@@ -0,0 +1,261 @@
...
+```
+> Please be ware that the module that uses the reuse module needs to be a Maven module itself or a submodule to a Maven module that declares the dependency to the Maven module. Usually you would declare the dependency in the `srv` module of your CAP Java project and use the reuse model in the service's CDS files then. In case you need to make sure that your `db` module is a Maven module and include it to the project's parent `pom.xml` file.
+
+When your Maven build is set up correctly you can use the reuse models in your CDS files using the standard `using` directive:
+
+```cds
+using { CatalogService } from 'com.sap.capire/bookshop';
+```
+
+> Note that the location syntax differs from the usual syntax and adheres to the groupId/artifactId syntax defined above. Also, the [CDS editor](../tools/#cds-editor) does not yet support this new location and hence shows an error marker for this line. This will be fixed soon.
```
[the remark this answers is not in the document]
rjayasinghe: provided a new version of that section with the latest commit

java/plugin_concept.md
```diff
@@ -0,0 +1,261 @@
...
+      <id>cds.resolve</id>
+      <goals>
+        <goal>resolve</goal> // [!code focus]
+      </goals>
+    </execution>
+    ...
+  </executions>
+</plugin>
+```
+> Please be ware that the module that uses the reuse module needs to be a Maven module itself or a submodule to a Maven module that declares the dependency to the Maven module. Usually you would declare the dependency in the `srv` module of your CAP Java project and use the reuse model in the service's CDS files then. In case you need to make sure that your `db` module is a Maven module and include it to the project's parent `pom.xml` file.
```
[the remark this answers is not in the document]
rjayasinghe: fixed

java/plugin_concept.md
rjayasinghe: Reworked the whole section a bit. The start of this section was based on a different ordering in the guide.

t-bonk (changes requested):
Please add a remark, that the groupId `com.sap.cds` shall not be used by 3rd party plugins, adapters etc.

java/plugin_concept.md
```diff
@@ -0,0 +1,302 @@
...
+
+When you provide your custom handler as part of an reuse library external to you application things change a bit. At first you need to decide whether you want to use Spring Boot's component model and rely on dependency injection or if you want to use one of the CAP Java ServiceLoader based extention points.
+
+The decision between the two is pretty straightforward: In case your handler depdends on other Spring components e.g. relies on dependency injection you should use the the [Spring approach](#spring-autoconfiguration). This applies as soon as you need to access another CAP Service like [`CqnService`](https://cap.cloud.sap/docs/java/application-services), [`PersistenceService`](https://cap.cloud.sap/docs/java/persistence-services) or a service via it's [typed service interface](https://cap.cloud.sap/docs/releases/nov23#typed-service-interfaces).
```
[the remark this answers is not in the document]
rjayasinghe: fixed

rjayasinghe:
> Please add a remark, that the groupId `com.sap.cds` shall not be used by 3rd party plugins, adapters etc.

fixed!

rjayasinghe:
> Please add a remark, that the groupId `com.sap.cds` shall not be used by 3rd party plugins, adapters etc.

I added a section for that.

t-bonk (approved):
LGTM

renejeglinsky:
Should I rename the file from plugin-concept to just plugins? Or sth else?

java/plugin_concept.md
```diff
@@ -0,0 +1,313 @@
...
+
+Especially when working with larger projects that may consists of many individual CAP Java applications or when building platform services that need to be integrated with CAP applications there is the requirement to extend CAP Java with custom, yet reusable code. 
+
+In the following sections the different extension points and mechanisms will be explained.
+
+## General Considerations
+
+### Java version
+
+When building CAP Java plugin modules you need to keep in mind that the generated Java byte code of the plugin has to be compatible with the Java byte code version of the potential consumers of the plugin. To be on the safe side we recommend to use *Java 17* as this is anyways the minimum Java version for CAP Java (for 2.x release) applications. In case you deviate from this you need to check and align with the potential consumers of the plugin.
```
renejeglinsky, line 29: Could it be an option to build the plugin for several versions? Or is this usually not feasible?

java/plugin_concept.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Maven GroupId and Java packages
+
+Of course it's up to your project / plugin how you call the corresponding Maven GroupId and Java packages. In order to avoid confusion and also to make responsibilities clear `com.sap.cds` for GroupId and Java package names are reserved for components maintained by the CAP Java team and must not be used for external plugins. This rule also includes sub-structures to `com.sap.cds` like `com.sap.cds.foo.plugin`.
```
renejeglinsky, line 33: Is this targeted towards an SAP-internal audience?

t-bonk: @renejeglinsky @rjayasinghe In my opinion this is targeted to all audiences, no matter whether they're internal or external.

java/plugin_concept.md
rjayasinghe: Not really, at the moment we need to define the cds-services as well as the JDK version fixed. This works for sure. For anything else we need some more thinking.

java/plugin_concept.md
```diff
@@ -0,0 +1,313 @@
...
+
+Starting with the 2.2 release CDS models, CSV import data and i18n files can now be shared through Maven dependencies in addition to npm packages. This means you can now provide CDS models, CSV files, i18n files and Java code (for example, event handlers) in a single Maven dependency.
+
+### Create the Reuse Model in a New Maven Artifact
+
+Simply create a plain Maven Java project and place your CDS models in the `main/resources/cds` folder of the reuse package under a unique module directory (for example, leveraging group ID and artifact ID): `src/main/resources/cds/com.sap.capire/bookshop/`. With `com.sap.capire` being the group ID and `bookshop` being the artifact ID.
+
+### Reference the Reuse Model in an Existing CAP Java Project
+
+Projects wanting to import the content simply add a Maven dependency to the reuse package to their `pom.xml` in the `<dependencies>` section.
```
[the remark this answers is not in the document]
renejeglinsky: Do I consume that, until this point locally? If yes, I guess the location of the new artifact is in parallel to my CAP project, as it is also for Node.js, right?
Or do we need a "Publish to Maven" step in between this and the previous step?

java/plugin_concept.md
```diff
@@ -0,0 +1,313 @@
...
+
+Due to those complications it's best practice to use the `AutoConfiguration` mechanism provided by Spring Boot in reuse modules that ship Spring components. For further details please refer to the [Spring Boot reference documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/using.html#using.auto-configuration).
+
+
+A complete end-to-end example for reusable event handlers can be found in this [blog post](https://blogs.sap.com/2023/05/16/how-to-build-reusable-plugin-components-for-cap-java-applications/).
+
+## Custom Protocol Adapters {#protocol-adapter}
+
+In CAP Java, the protocol adapter is the mechanism to implement inbound communication (another service or the UI) to the CAP service in development. The task of a protocol adapter is to translate any incoming requests of a defined protocol to CQL statements that then can be executed on locally defined CDS services. CAP Java comes with 3 protocol adapters (OData V2 and V4, and HCQL) but can be extended with custom implementations. In this section we'll have a deeper look on how such a protocol adapter can be built and registered with the CAP Java runtime.
```
renejeglinsky, line 190: How about shortening this intro and link to https://cap.cloud.sap/docs/java/architecture#protocol-adapters? This chapter about protocol adapters would need an update then, I guess?

renejeglinsky:
Will merge this and go over some points with Robin later.

renejeglinsky (approved):

PR #527, opened 2023-11-08
outbox: move to top level, make transactional outbox default

David-Kunz:
To be released with @sap/cds 7.4.0.


johannes-vogel (approved):
links to be adapted internally

PR #528, opened 2023-11-08
destinationOptions.jwt: null

sjvans:
with cds^7.4


guides/using-services.md
```diff
@@ -1516,7 +1535,7 @@ Read more in the full reference of all [supported retrieval strategy values](htt
 </div>
 
 
-## ⑤ Add Qualities
+## ⑥ Add Qualities
```
sjvans, line 1538:
```suggestion
## ⑤ Add Qualities
```

guides/using-services.md
```diff
@@ -1494,6 +1494,25 @@ With the value `alwaysProvider` you can ensure that the destination is always re
...
+        /* ... */
+        "forwardAuthToken": false
+      }
+    }
+  }
+}
+```
+
+This can be beneficial when the authentication flow does not require a JWT (cf. [Authentication and JSON Web Token (JWT) Retrieval](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destinations#authentication-and-json-web-token-jwt-retrieval)). For example, ...
+
```
sjvans, line 1515:
```suggestion
Further, you can instruct CAP to withhold the request's JWT from SAP Cloud SDK by setting destination option `jwt` to `null`:

```jsonc
"cds": {
  "requires": {
    "SERVICE_FOR_PROVIDER": {
      /* ... */
      "credentials": {
        /* ... */
      },
      "destinationOptions": {
        "jwt": null
      }
    }
  }
}
```

Passing the request's JWT to SAP Cloud SDK has implications on, amongst others, the effective defaults for selection strategy and isolation level.
Please see [Authentication and JSON Web Token (JWT) Retrieval](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destinations#authentication-and-json-web-token-jwt-retrieval) for more details.

```

File: guides/using-services.md [hunk 3]
```diff
@@ -1494,6 +1494,27 @@ With the value `alwaysProvider` you can ensure that the destination is always re
...
+```jsonc
+"cds": {
+  "requires": {
+    "SERVICE_FOR_PROVIDER": {
+      /* ... */
+      "credentials": {
+        /* ... */
+      },
+      "destinationOptions": {
+        "jwt": null
```
renejeglinsky, line 1508: Would it make sense to include this into the previous snippet? Or is the combination of `selectionStrategy` and `"jwt: null` odd?

guides/using-services.md
```diff
@@ -1494,6 +1494,27 @@ With the value `alwaysProvider` you can ensure that the destination is always re
...
+      }
+    }
+  }
+}
+```
+
+Passing the request's JWT to SAP Cloud SDK has implications on, amongst others, the effective defaults for selection strategy and isolation level.
```
renejeglinsky, line 1515: What is the effect if I configure `"jwt": null`? This sentence explains roughly what I need to consider when not adding this config, right? What do we recommend from our CAP angle?

guides/using-services.md
sjvans: they can be used in combination but also separately, depending on the scenario. we can merge if that's understood. i defer to your expertise on that question. ;)

guides/using-services.md
sjvans: i'm afraid we can't recommend anything, it depends on the scenario. needing this config is rare, but not "invalid", so i think we can't label it as "not recommended"...

sjvans, line 1515:
```suggestion
Passing the request's JWT to SAP Cloud SDK has implications on, amongst others, the effective defaults for selection strategy and isolation level.
In rare cases, these defaults are not suitable, for example when the request to the upstream server does not depend on the current user.
```

guides/using-services.md
renejeglinsky: I'd say it depends on the content around that snippet if that is understood. But I think it's visually more appealing to not have this duplication of snippets where only a small piece is different.
I'll go ahead then and make a proposal in one commit, so that we can revert it if needed easily.
Will have that until tomorrow hopefully.

renejeglinsky: If I set `"jwt": null` and "selectionStrategy": "alwaysProvider" it's the provider who holds the destination to be used. It's redundant, right?

If I set `"jwt": null` and "selectionStrategy": "alwaysSubscriber" it's the subscriber who holds the destination to be used but I don't want to use the JWT.  Instead I want to retrieve the destination from DestinationAccessorOptions as described [here](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destinations#destination-lookup-without-a-json-web-token-jwt).

Is this correct?

sjvans: > If I set `"jwt": null` and `"selectionStrategy": "alwaysProvider"` it's the provider who holds the destination to be used. It's redundant, right?

yes with regards to which tenant is the target of the request, but caching in cloud sdk, for example, is different with `"selectionStrategy": "alwaysProvider"` and no `"jwt": null`.

> If I set `"jwt": null` and `"selectionStrategy": "alwaysSubscriber"` it's the subscriber who holds the destination to be used but I don't want to use the JWT. Instead I want to retrieve the destination from DestinationAccessorOptions as described [here](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destinations#destination-lookup-without-a-json-web-token-jwt).
> 
> Is this correct?

`alwaysSubscriber` and no jwt wouldn't work because, afaik, the cloud sdk relies on the jwt to know which tenant shall be used.

renejeglinsky: I gave it a try: https://github.com/cap-js/docs/pull/528/commits/53614091ff9d611572ba7790fedbef6cd859fa27
I'll approve and leave it to you to merge if you think it's fine.

PR #535, opened 2023-11-10
Enhanced sqlite docu for how to use in production

schiwekM:
Hi Rene,

I've added a block about running sqlite in production to the docu to improve guidance in valid use cases like running CAP as a proxy for S4. 

BR,
Marten

### DIFF
--- a/guides/databases-sqlite.md
+++ b/guides/databases-sqlite.md
@@ -782,3 +782,13 @@ Having said this, there can indeed be scenarios where SQLite might also be used 
 ::: warning
 SQLite only has limited support for concurrent database access due to its very coarse lock granularity. This makes it badly suited for applications with high concurrency.
 :::
+
+If your application is only used as a proxy for example for S4/HANA without any own database artifacts, you can leverage SQLite in production to avoid issues. For that require `@cap-js/sqlite` as a normal dependency in your package.json file and set `cds.env.requires.db` to `sqlite`, because else SQlite is not used in production. 
+
+```json
+"cds": {
+  "requires": {
+    "db": "sqlite"
+  }
+}
+```


renejeglinsky:
Hi Marten,

Thanks a lot! I'll discuss this and get back to you. :)

renejeglinsky:
Hi Marten,
could you enhance the description of the scenario and the boundary conditions, like losing data when the corresponding container is dropped and restarted? The technical configuration is the easier part compared to judging and identifying if one would really want to SQLite in prod.

schiwekM:
Hi Rene,

I encounter the issue with a partner in an S4 transformation. Basically they use their CAP app as a proxy for the S4 API and have no entities to persist. However when trying to start without any database, the server always crashed when trying to do cds.connect.to upon the request. The workaround was to just to add the sqlite db in prod as the requests are anyways forwarded to S4.

schiwekM:
> > 4 transformation. Basically they use their CAP app as a proxy for the S4 API and have no entities to persist. However when trying to start without any database, the server always crashed when trying to do cds.connect.to upon the request. The workaround was to just to add the sqlite db in prod as the requests are anyways forwarded to S4.
> 
> I rather think that this is a configuration issue on the customer side. If you do not have any database, do not set a `cds.requires.db` entry in the package.json.

Strangely when they did not, it crashed upon requests.

johannes-vogel:
@schiwekM Do you by any chance have a pointer for me look at the config?

schiwekM:
> @schiwekM Do you by any chance have a pointer for me look at the config?

Send an email with contact details

johannes-vogel:
@schiwekM I'd propose to close this PR.
I did not hear back and I'm rather sure that this is not required, if you do not have the need for a database.

PR #538, opened 2023-11-13
Better styling for `cds-plugins`

swaldmann:
- Fixed external link icon position + spacing
- Reusing classes instead of repeated inline styling
- Subtle hover effect for plugin images

→ can discuss in UA sync


swaldmann:
Also cap/docs/pull/458/ internally to align it.

chgeo (approved):

PR #558, opened 2023-11-27
Be more explicit in the @odata.Type warning

hjboth:
not all customers understand that there is absolutely NO data conversion/transformation induced by those annotations

### DIFF
--- a/advanced/odata.md
+++ b/advanced/odata.md
@@ -182,9 +182,9 @@ entity Books {
 ```
 
 ::: warning
-This annotation affects the client side facing API only, there is no automatic data modification of any kind (like truncation, converstion, etc.) done
-behind the scenes. It is in your responsibility to perform all required modifications on the data stream. If you are not doing the required conversions,
-it is possible to "cast" any scalar CDS type into any (in-)compatible EDM type:
+This annotation affects the client side facing API only, there is no automatic data modification of any kind (like rounding, truncation, conversion, etc.) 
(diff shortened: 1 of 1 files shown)


chgeo (approved):

stewsk (approved):

advanced/odata.md
```diff
@@ -182,7 +182,9 @@ entity Books {
 ```
 
 ::: warning
-It is possible to "cast" any scalar CDS type into any (in-)compatible EDM type:
+This annotation affects the client side facing API only, there is no automatic data modification of any kind (like rounding, truncation, conversion, etc.) 
+behind the scenes. It is in your responsibility to perform all required modifications on the data stream such that the values match their type in the API. 
+If you are not doing the required conversions, it is possible to "cast" any scalar CDS type into any (in-)compatible EDM type:
```
renejeglinsky, line 187:
```suggestion
This annotation affects the client side facing API only. There's no automatic data modification of any kind behind the scenes, like rounding, truncation, conversion, and so on. It's your responsibility to perform all required modifications on the data stream such that the values match their type in the API. 
If you are not doing the required conversions, you can "cast" any scalar CDS type into any incompatible EDM type:
```

renejeglinsky: The last sentence, as the introduction to the following snippets, implies that the following example is a **Don't** example. Is that correct?

hjboth: yes it is

renejeglinsky (approved):

PR #559, opened 2023-11-28
Update cds-utils.md

daogrady:
`decodeURI` and `decodeURIComponent` were missing the parameter in their heading, which other methods with parameters were more explicit about. Also, I think it would be helpful to link to the original `decodeURI` and `decodeURIComponent` documentation. Do we have goto-documentation for native functions? MDN maybe?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI and
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent respectively


swaldmann:
> MDN maybe?

Yes, MDN is by far the best resource for that imo.

PR #570, opened 2023-12-06
consuming services: connect to on-prem from bas

sjvans:
add tip linking to [Connecting to External Systems From the Business Application Studio](https://sap.github.io/cloud-sdk/docs/js/guides/bas)


guides/using-services.md
```diff
@@ -1267,6 +1267,11 @@ cds bind --exec -- mvn spring-boot:run \
 [Learn more about `cds bind --exec`.](../advanced/hybrid-testing#run-arbitrary-commands-with-service-bindings){.learn-more}
 
 
+::: tip
+If you are developing in the Business Application Studio and want to connect to an on-premise system, you will need to do so via Business Application Studio's built-in proxy, for which you need to add configuration in an `.env` file. See [Connecting to External Systems From the Business Application Studio](https://sap.github.io/cloud-sdk/docs/js/guides/bas) for more details.
```
beckermarc, line 1271: Is there also an equivalent documentation for this for Java? The docs you linked to seem to be for the Cloud SDK for JavaScript. Maybe link to this section here for Java: https://sap.github.io/cloud-sdk/docs/java/features/connectivity/destination-service#reach-on-premise-service-from-the-sap-business-application-studio

beckermarc: I don't know these docs well and also missed so far that they the page I found links to the section Sebastian added here. I think however the .env files are different for Java and Node.js after reading through the docs. Maybe @MatKuhr from Cloud SDK can shed some light on this as well?

sjvans: 🙏

sjvans: > ... instead you have to give proxy host + port in the destination as described in the Java article.

but does that mean you need to change coding/ existing files? the `.env` approach is nice because it makes it completely transparent for the app...

MatKuhr: > change coding/ existing files

No, your code remains unchanged. The only difference between Java and JS is that for Java the destination (however you prefer to define it) has to have the additional properties proxy host + port.

renejeglinsky: @beckermarc Should we add a similar note with the link you provided for Java here?

beckermarc: I think would make sense. From my understanding this topic affects both Java and Node.js

renejeglinsky: Done: https://github.com/cap-js/docs/pull/570/commits/dcc2c42b4674451b1455a02820fe3890c97a4c0e

guides/using-services.md
```diff
@@ -1266,6 +1271,10 @@ cds bind --exec -- mvn spring-boot:run \
 
 [Learn more about `cds bind --exec`.](../advanced/hybrid-testing#run-arbitrary-commands-with-service-bindings){.learn-more}
 
+::: tip
+If you are developing in the Business Application Studio and want to connect to an on-premise system, you will need to do so via Business Application Studio's built-in proxy, for which you need to add configuration to your destination environment variable. See [Reach On-Premise Service from the SAP Business Application Studio](https://sap.github.io/cloud-sdk/docs/java/features/connectivity/destination-service#reach-on-premise-service-from-the-sap-business-application-studio) for more details.
+:::
```
renejeglinsky, line 1276: @beckermarc Please review and if needed suggest changes. Thanks a lot :)

beckermarc (approved):

renejeglinsky (approved):

PR #539, opened 2023-11-13
Describe Actions / Functions in service POJOs

mofterdinger:

PR #572, opened 2023-12-07
Incorporate feedback about actions and functions

mofterdinger:
Followup to #539


java/application-services.md
```diff
@@ -308,7 +308,7 @@ Now, just call the review action from custom handler code:
   ...
 ```
 
-To trigger the action or function, the event context needs to be filled with the parameter values and emitted on the service:
+Alternatively the event context can be used to trigger the action or function. This approach is really usefull for a generic use case, where typed interfaces don't work. Therefore the event context needs to be filled with the parameter values and emitted on the service:
```
BraunMatthias, line 311:
```suggestion
Alternatively, the event context can be used to trigger the action or function. This approach is useful for the generic use case, where typed interfaces don't work. The event context needs to be filled with the parameter values and emitted on the service:
```

java/application-services.md
```diff
@@ -224,6 +224,30 @@ service CatalogService {
 }
 ```
 
+The `cds-maven-plugin` generates event-specific event context interfaces for the action or function, based on its CDS model definition. These event context interfaces give direct access to the parameters and the return value of the action or function.
```
BraunMatthias, line 227:
```suggestion
The `cds-maven-plugin` generates event-specific event context interfaces for the action or function, based on its CDS model definition. These event context interfaces provide direct access to the parameters and the return value of the action or function.
```

BraunMatthias (approved):

java/application-services.md
```diff
@@ -227,7 +227,7 @@ service CatalogService {
 The `cds-maven-plugin` generates event-specific event context interfaces for the action or function, based on its CDS model definition. These event context interfaces provide direct access to the parameters and the return value of the action or function.
 For bound actions or functions the event context interface provides a [CqnSelect](query-api#select) statement, which targets the entity the action or function was triggered on.
 
-Event-specific event context, generated by the CAP Java SDK Maven Plugin:
+Action-specific event context, generated by the CAP Java SDK Maven Plugin:
```
BraunMatthias, line 230: but the context is generated for an event here, isn't it?

java/application-services.md
```diff
@@ -284,31 +308,7 @@ Now, just call the review action from custom handler code:
...
-
-    // The return value
-    void setResult(Reviews review);
-    Reviews getResult();
-
-}
-```
-
-To trigger the action or function, the event context needs to be filled with the parameter values and emitted on the service:
+Alternatively, the event context can be used to trigger the action or function. This approach is useful for the generic use case, where typed interfaces don't work. The event context needs to be filled with the parameter values and emitted on the service:
```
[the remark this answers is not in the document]
mofterdinger: done, replaced specific context with generic EventContext interface.

java/consumption-api.md
```diff
@@ -100,6 +100,9 @@ public class CatalogServiceAPI {
 }
 ```
 
+The `cds-maven-plugin` generates specific interfaces for services in the CDS model. These service interfaces provide Java methods for each action and function, which can be used to trigger the corresponding event.
+For further details, see [Trigger Action or Function](./application-services#trigger-action-or-function).
```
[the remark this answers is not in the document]
mofterdinger: I will schedule a meeting to discuss what shall be removed or changed. I'm not sure what you exactly mean.

java/application-services.md
```diff
@@ -256,9 +256,9 @@ The event handler registration and implementation is as follows:
 public class CatalogServiceHandler implements EventHandler {
 
     @On(event = "review", entity = Books_.CDS_NAME)
-    public void reviewAction(ReviewEventContext context) {
+    public void reviewAction(EventContext context) {
```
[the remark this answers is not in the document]
mofterdinger: ah thanks, will fix it.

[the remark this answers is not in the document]
beckermarc: But I didn't mean to adapt this part. Using the generic context when implementing a concrete action shouldn't be done. What I was referring to was the example below, where the event context is used to trigger the action. Using the typed context interface there is no longer useful -> should use typed Service interface instead. If we want to still describe how to trigger the action via context, then we should give an example using a generic context there.

However examples for implementing an action usually should always use the typed event context interface. As actions are usually not implemented generically.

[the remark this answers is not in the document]
mofterdinger: ok, I will revert my last change and modify the trigger action scenario with a generic event context.

java/application-services.md
```diff
@@ -284,36 +308,12 @@ Now, just call the review action from custom handler code:
...
-To trigger the action or function, the event context needs to be filled with the parameter values and emitted on the service:
+Alternatively, the event context can be used to trigger the action or function. This approach is useful for generic use cases, where typed interfaces are not available. The event context needs to be filled with the parameter values and emitted on the service:
 
 ```java
-    ReviewEventContext context = ReviewEventContext.create();
-    context.setCqn(Select.from(BOOKS).byId("myBookId"));
-    context.setStars(review.getRating());
+    EventContext context = EventContext.create("review", Books_.CDS_NAME);
+    context.put("cqn", Select.from(BOOKS).byId("myBookId"));
+    context.put("rating", review.getRating());
```
beckermarc, line 316: generic event context will not have `getResult()` method used below.

java/application-services.md
mofterdinger: lets keep the suggestion from Marc.

java/application-services.md
mofterdinger: fixed.

java/consumption-api.md
```diff
@@ -69,39 +75,23 @@ public class EventHandlerClass implements EventHandler {
...
+    @Autowired
+    private AdminService adminService;
 
 }
 ```
-
-The `cds-maven-plugin` generates specific interfaces for services in the CDS model. These service interfaces provide Java methods for each action and function, which can be used to trigger the corresponding event.
-For further details, see [Trigger Action or Function](./application-services#trigger-action-or-function).
+::: tip
+For the injection of specific service interfaces the annotation `@Qualifier` is not required.
```
beckermarc, line 93:
```suggestion
For the injection of specific service interfaces the annotation `@Qualifier` is usually not required.
```

If there are multiple instances of the same CDS service at runtime a qualifier might still be required. But this is not often the case.

java/application-services.md
```diff
@@ -224,6 +224,30 @@ service CatalogService {
 }
 ```
 
+The `cds-maven-plugin` generates event-specific event context interfaces for the action or function, based on its CDS model definition. These event context interfaces provide direct access to the parameters and the return value of the action or function.
```
beckermarc, line 227:
```suggestion
The `cds-maven-plugin` generates event context interfaces for the action or function, based on its CDS model definition. These event context interfaces provide direct access to the parameters and the return value of the action or function.
```

java/consumption-api.md
```diff
@@ -47,6 +47,12 @@ Application Services are created with the fully qualified name of their CDS defi
 ApplicationService adminService = catalog.getService(ApplicationService.class, "AdminService");
 ```
 
+As of version 2.4.0, the [CAP Java SDK Maven Plugin](./development/#cds-maven-plugin) is capable of generating specific interfaces for services in the CDS model. These service interfaces also provide Java methods for actions and functions, which allow direct access to the action's or function's parameters. These specific interfaces can also be used to access the service:
```
beckermarc, line 50:
```suggestion
As of version 2.4.0, the [CAP Java SDK Maven Plugin](./development/#cds-maven-plugin) is capable of generating specific interfaces for services in the CDS model. These service interfaces also provide Java methods for actions and functions, which allows easily calling actions and functions with their parameters. These specific interfaces can also be used to get access to the service:
```

beckermarc (approved):

renejeglinsky (approved):

PR #575, opened 2023-12-08
Recommend to use req.data

mariayord:
Fixes https://github.tools.sap/cap/issues/issues/13696


node.js/events.md
```diff
@@ -169,7 +169,16 @@ The name of the incoming event, which can be one of:
...
 Contains the event data. For example, the HTTP body for `CREATE` or `UPDATE` requests, or the payload of an asynchronous event message.
 
+Use `req.data` for modifications as shown below:
 
+```js
+this.before ('UPDATE',Books, req => {
+  // DO:
+  req.data.author = 'Schmidt'
+  // DON'T:
+  req.query.UPDATE.data.author = 'Schmidt'
```
johannes-vogel, line 179:
```suggestion
  req.data.author = 'Schmidt'  // [!code ++]
  req.query.UPDATE.data.author = 'Schmidt'  // [!code --]
```

mariayord: Learned something new :) Thanks!

node.js/events.md
```diff
@@ -169,7 +169,14 @@ The name of the incoming event, which can be one of:
 
 Contains the event data. For example, the HTTP body for `CREATE` or `UPDATE` requests, or the payload of an asynchronous event message.
 
+Use `req.data` for modifications as shown below:
 
+```js
+this.before ('UPDATE',Books, req => {
+  req.data.author = 'Schmidt'  // [!code ++]
+  req.query.UPDATE.data.author = 'Schmidt'  // [!code --]
```
renejeglinsky, line 177: Why would someone have used this way? Is it an old way of doing the same?

mariayord: @renejeglinsky It looks like there are still people using the old way: https://github.tools.sap/cap/issues/issues/13696

johannes-vogel: It's not actually the "old" way but a more complicated alternative. We want to point people into the easier solution.

renejeglinsky: Ok. Usually we have those "Don't do this anymore but do that" in release notes because some things changed. I understand that this is not the root cause here. So fine with me. Maybe in the future we could think about deleting this "Don't" version to not point people on "wrong" things.

renejeglinsky (approved):

PR #580, opened 2023-12-11
cdl: Mention that propagated annotations can't be extended via `...`

bugwelle:


stewsk (approved):

cds/cdl.md
```diff
@@ -1168,7 +1168,9 @@ annotate TravelService.Travel with @UI.LineItem: [
 ];
 ```
 
-
+::: tip
+Only direct annotations, not propagated ones, can be extended using `...`.
```
renejeglinsky, line 1172:
```suggestion
Only direct annotations can be extended using `...`. It's not supported to extend propagated annotations, for example, from aspects or associations.
```

bugwelle:
```suggestion
Only direct annotations can be extended using `...`. It's not supported to extend propagated annotations, for example, from aspects or types.
```

Thanks. I'd go for "type" here. :)

renejeglinsky (approved):

PR #584, opened 2023-12-12
Open Telemetry support for Java

StefanHenke:
Includes:
 - Configuration on Cloud Foundry
 - custom instrumentation

Out:
 - Kyma support


java/observability.md
```diff
@@ -472,3 +472,111 @@ public class AppActuator {
...
+In addition, it's possible to provide manual instrumentation from within a CAP Java application using the [Open Telemetry Java API](https://opentelemetry.io/docs/instrumentation/java/manual/), for example, from a custom event handler.
+
+### Configuration
+
+1) Bind your CAP Java application to a service instance of `cloud-logging`. On creation of the service instance, it's important to enable the Open Telemetry capabilities by passing the following parameter:
+    ```json
+    {
+      "ingest_otlp": true
+    }
+    ```
```
renejeglinsky, line 497: The preferred way of MTA deployment would allow us to give here the whole mta.yaml config. I think that would be very helpful. Maybe we could even include step 3 in this snippet and then explain both parts.
WDYT?

StefanHenke: @renejeglinsky : The whole mta.yaml file is rather big. For our bookshop sample it is about 100 lines. Or do you mean to describe how to pass the payload for service creation within a mta.yaml? The payload syntax given so far is quite generic and could be passed also when manually creating the service instance via CLI or cockpit. If we want to concentrate on mta.yaml syntax, I could transform the payload accordingly.

renejeglinsky: Yes, I mean only the part for creating the service here. :)

StefanHenke: After thinking more about it, I´m wondering if we should change this part completely.

The parameter `ingest_otlp` is in most cases not the only parameter. So, it simply needs to be added to all other parameters which are required for Cloud Logging Service, but which are not specific for Open Telemetry. People might get the feeling that this is the complete payload which is not the case.

renejeglinsky: Thank you! :)

renejeglinsky (approved):

java/observability.md
```diff
@@ -472,3 +472,118 @@ public class AppActuator {
...
+
+  @After(entity = Books_.CDS_NAME)
+  public void afterAddReview(AddReviewContext context) {
+      Span childSpan = tracer.spanBuilder("setBookRating").startSpan();
+      childSpan.setAttribute("book.title", context.getResult().getTitle());
+      childSpan.setAttribute("book.id", context.getResult().getBookId());
+      childSpan.setAttribute("book.rating", context.getResult().getRating());
+
+      try(Scope scope = childSpan.makeCurrent()) {
+        ratingCalculator.setBookRating(context.getResult().getBookId());
```
BraunMatthias, line 562: indentation

BraunMatthias (commented):
* We should mention the logger and its effects
* You could add a nice picture with the bar chart

Overall nice piece of docu!

java/observability.md
```diff
@@ -472,3 +472,118 @@ public class AppActuator {
 }
 ```
 The `AppActuator` bean registers an actuator with name `app` that exposes a simple version string.
+
+## Open Telemetry { #open-telemetry }
```
BraunMatthias, line 476: Add a link to OT as the new standard in the monitoring section: https://pages.github.tools.sap/cap/docs/java/observability#monitoring

StefanHenke:
> * We should mention the logger and its effects
> * You could add a nice picture with the bar chart
> 
> Overall nice piece of docu!

I have added both

BraunMatthias (approved):

PR #592, opened 2023-12-15
cosmetics

smahati:


node.js/core-services.md
```diff
@@ -99,16 +99,16 @@ const db = await cds.connect.to('db')
 
 ## Implementing Services
 
-By default `cds.serve` creates instances of `cds.ApplicationService` for each found service definition, which provides generic implementations for all CRUD operations, including full support for deep document structures, declarative input validation and many other out-of-the-box features. Yet, you'd likely need to provide domain-specific custom logic, especially for custom actions and functions, or for custom validations. Learn below about:
+By default `cds.serve` creates an instance of `cds.ApplicationService` for each service definition it finds. Each instance provides generic implementations for all CRUD operations, including full support for deep document structures, declarative input validation and many other out-of-the-box features. Yet, you'd likely need to provide domain-specific custom logic, especially for custom actions and functions, or for custom validations. In the next sections, you can learn:
```
renejeglinsky, line 102:
```suggestion
By default `cds.serve` creates an instance of `cds.ApplicationService` for each service definition it finds. Each instance provides generic implementations for all CRUD operations, including full support for deep document structures, declarative input validation and many other out-of-the-box features. Yet, you'd likely need to provide domain-specific custom logic, especially for custom actions and functions, or for custom validations. In the next sections, you can learn the following:
```

node.js/core-services.md
```diff
@@ -99,16 +99,16 @@ const db = await cds.connect.to('db')
...
 
 - **How** to provide custom implementations?
-- **Where**, that is, in which files, to put that?
+- **Where**, i.e. in which files, to add the implementation?
```
renejeglinsky, line 105: We try to avoid latin abbreviations. But if you can phrase it differently that would also be fine :)
```suggestion
- **Where**, that is, in which files, to add the implementation?
```

renejeglinsky (commented):
Thanks a lot! :)

renejeglinsky (approved):

PR #603, opened 2023-12-22
cds.outboxed

David-Kunz:


David-Kunz:
Hi @renejeglinsky ,

Yes, you mean the API style link in the navigation bar? That's already merged: 

https://pages.github.tools.sap/cap/docs/node.js/outbox

node.js/outbox.md
```diff
@@ -7,23 +7,87 @@ status: released
...
+```js
+const outboxed = cds.outboxed(srv, { kind: 'persistent-outbox' })
+```
+
+::: warning
+The persistent outbox can only be used if it's enabled globally with `cds.requires.outbox = true` because it requires a dedicated database table.
+:::
+
+::: warning
+Once you outboxed a service, you cannot override its outbox configuration options again.
```
renejeglinsky, line 48: Does that mean, if I outboxed a service w/o optional configuration, I can't add configuration later on? Or is it just not possible to remove optional configuration?

node.js/outbox.md
```diff
@@ -7,23 +7,87 @@ status: released
...
+You can also configure services to be outboxed by default:
+
+```json
+{
+  "requires": {
+    "yourService": {
+      "kind": "odata",
+      "outbox": {
+        "kind": "persistent-outbox"
+      }
```
renejeglinsky, line 62: Can this also be `"outbox": "true"`? I'd assume that this would use the persistent outbox as default.

node.js/outbox.md
```diff
@@ -7,23 +7,87 @@ status: released
...
+```js
+const srv = await cds.connect.to('yourService')
+const outboxed = cds.outboxed(srv)
+
+await outboxed.emit('someEvent', { some: 'message' }) // asynchronous
+await outboxed.send('someEvent', { some: 'message' }) // asynchronous
+```
+
+::: tip
+You still need to `await` these operations, because messages might be stored in a database first, inside your main transaction.
```
renejeglinsky, line 34: Is this still correct?
```suggestion
You still need to `await` these operations inside your main transaction, because messages might be stored in a database first.
```

node.js/outbox.md
David-Kunz: Yes, that is correct. The following would not be allowed

```js
const o1 = cds.outboxed(srv)
const o2 = cds.outboxed(srv, { maxRetries: 1000 })
```

node.js/outbox.md
David-Kunz: Yes, it can also be `outbox: true`, then the config from `cds.requires.outbox` is used (not necessarily persistent-outbox).

node.js/outbox.md
David-Kunz: I think this changes the meaning a bit, no? You have to `await` those operations. Then they are stored in the database, together with the main transaction.

node.js/outbox.md
```diff
@@ -7,23 +7,87 @@ status: released
...
 <!--- Migrated: @external/node.js/Messaging/0-index.md -> @external/node.js/messaging.md -->
 
+# Outboxing with `cds.outboxed`
 
-# Transactional Outbox
+## Overview
 
-Usually the emit of messages should be delayed until the main transaction succeeded. Otherwise recipients will also receive messages in case of a rollback.
-To solve this problem, an outbox is used internally to defer the emit of messages until the success of the current transaction.
+Often, remote operations should be delayed until the main transaction succeeded. Otherwise they're also triggered in case of a rollback.
```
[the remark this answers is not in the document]
David-Kunz: the 'remote operations'

node.js/outbox.md
renejeglinsky: That's why I was asking. To me it'd mean the same :)
```suggestion
You still need to `await` these operations. Then the messages are stored in the database, together with the main transaction.
```

node.js/outbox.md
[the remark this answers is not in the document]
renejeglinsky:
```suggestion
Often, remote operations should be delayed until the main transaction succeeded. Otherwise, the remote operations are also triggered in case of a rollback.
```

node.js/outbox.md
renejeglinsky: So later means, later in the same process and not later at another point in time? I was assuming that it's more time-related and if I want to change sth related to outbox later on, I'd need to rename a service or sth like that. Too complicated thoughts, right? ;)

node.js/outbox.md
renejeglinsky: But here it's just about outboxing and not explicitly about persistent-outbox, right? Or is it more common, that if I do config on service level, that I specify it?

node.js/outbox.md
David-Kunz: Temporal later. If this code is called again with different settings, those options are not taken into account. In another process, it could be that it is taken into account by accident... So the general guideline should be not to use it with different options.

node.js/outbox.md
David-Kunz: That's true, one can set it to either `true` (in that case, the `cds.requires.outbox` settings are used) or to a specific outbox. Yes, it probably makes sense to set it to `true` here.

node.js/outbox.md
```diff
@@ -7,23 +7,95 @@ status: released
...
+You can also configure services to be outboxed by default:
+
+```json
+{
+  "requires": {
+    "yourService": {
+      "kind": "odata",
+      "outbox": {
+        "kind": "persistent-outbox"
+      }
```
David-Kunz, line 70:
```suggestion
      "outbox": true
```

node.js/outbox.md
renejeglinsky: I think I understand. It seems pretty hard to ensure that this doesn't happen with multiple people working on the same project. Isn't it? Do we have support for that, like in the editor or so?

renejeglinsky (approved):

PR #609, opened 2024-01-04
ts custom auth

johannes-vogel:
will be develivered with cds 7.6

### DIFF
--- a/node.js/authentication.md
+++ b/node.js/authentication.md
@@ -407,10 +407,13 @@ module.exports = function custom_auth (req, res, next) {
 The TypeScript equivalent has to use the default export.
 
 ```ts
-export default function custom_auth(req: any, res: any, next: any) {
+import cds from "@sap/cds";
+import {Request, Response, NextFunction} from "express";
+type Req = Request & { user: cds.User, tenant: string };
+
+export default function custom_auth(req: Req, res: Response, next: NextFunction) {
   // do your custom authentication ...
 }
-```
 
 [If you want to customize the user ID, please also have a look at this example.](./middlewares#customization-of-req-user){.learn-more}
 
--- a/node.js/authentication.md
+++ b/node.js/authentication.md
@@ -404,6 +404,14 @@ module.exports = function custom_auth (req, res, next) {
 }
 ```
 
+The TypeScript equivalent has to use the default export.
+
+```ts
+export default function custom_auth(req: any, res: any, next: any) {
+  // do your custom authentication ...
+}
+```
+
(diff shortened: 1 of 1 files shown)


sjvans (approved):

node.js/authentication.md
```diff
@@ -404,6 +404,14 @@ module.exports = function custom_auth (req, res, next) {
 }
 ```
 
+The TypeScript equivalent has to use the default export.
+
+```ts
+export default function custom_auth(req: any, res: any, next: any) {
```
hakimio, line 410: You can use proper types:
```ts
import { Request } from '@sap/cds';

export default function custom_auth(request: Request, response: Response, next: Function) {}
```

johannes-vogel: The `request` here is the express request object. There might be types available for express as well but I'd prefer to have that snippet as simple as possible.

johannes-vogel: The `cds.Request` type is simply wrong as e. g. `query` has a completely different semantic in the express request object.

For proper type support one would need to enhance the request type of express. I'll double check for improving the snippet. If it grows too big, I'll add a comment in the snippet.

node.js/authentication.md
```diff
@@ -404,6 +404,14 @@ module.exports = function custom_auth (req, res, next) {
...
+  // do your custom authentication ...
+}
+```
```
johannes-vogel, line 413:
```suggestion
```ts
import cds from "@sap/cds";
import {Request, Response, NextFunction} from "express";
type Req = Request & { user: cds.User, tenant: string };

export default function custom_auth(req: Req, res: Response, next: NextFunction) {
  // do your custom authentication ...
}
```
```

Comment (hakimio) [reply node.js/authentication.md:413]: Looks good to me.

Comment (hakimio) [reply node.js/authentication.md:413]: Just a side note: maybe this type should be included in `cds-types`:
```ts
type AuthRequest = Request & { user: cds.User; tenant: string };
```

node.js/authentication.md
```diff
@@ -404,6 +404,17 @@ module.exports = function custom_auth (req, res, next) {
...
+
+```ts
+import cds from "@sap/cds";
+import {Request, Response, NextFunction} from "express";
+type Req = Request & { user: cds.User, tenant: string };
+
+export default function custom_auth(req: Req, res: Response, next: NextFunction) {
+  // do your custom authentication ...
+}
+
```
renejeglinsky, line 417:
```suggestion
```
```

renejeglinsky (approved):

PR #617, opened 2024-01-10
Fix orphaned "Overview" in /guides

swaldmann:
This orphaned "Overview" re-appeared from a side-effect of a different fix:

Screenshot 2024-01-10 at 15 50 02

First I let consumers pass in a `_filter` method, but I like this generic solution even better – just make sure the site doesn't end with the base path for ToC generation.

This variant also preserves the "Overview" entry in the top nav bar.


chgeo (approved):
Thanks!

PR #618, opened 2024-01-10
Tip headings in messaging cookbook

swaldmann:
As the headings already link to those pages and they are easily discoverable in the left sidebar nav I think we can remove those additional hints.

### DIFF
--- a/guides/messaging/index.md
+++ b/guides/messaging/index.md
@@ -59,7 +59,7 @@ class Receiver extends cds.Service { async init() {
 }}
 ```
 
-::: tip
+::: tip Emitters vs Receivers
 **Emitters** usually emit messages to *themselves* to inform *potential* listeners about certain events.
 **Receivers** connect to *Emitters* to register handlers to such emitted events.
 :::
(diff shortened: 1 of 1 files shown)


swaldmann:
Just one thing: the order of the S4 section + Event Mesh section are in a different order than our sidecar nav. Any preference which should come first?

renejeglinsky:
I think the whole point of those sections is to contain that link ( ok, a little exaggerated ;) )
Anyways, w/o those links I'd argue that we don't need those intro sections at all because I assume:
1. Users reading the section here probably haven't noticed the navigation and that there's the guide they're looking for. So they need this tip.
2. User who see the guide in the navigation do jump directly into that guide and not read these sections at all.

Does it make sense?

swaldmann:
Re-added the links and only replaced some default "TIP"s. WDYT?

guides/messaging/index.md
```diff
@@ -59,7 +59,7 @@ class Receiver extends cds.Service { async init() {
 }}
 ```
 
-::: tip
+::: tip Emitters vs Receivers
```
swaldmann, line 62: Inspired from below, where we use

```
::: tip Event Listeners vs Interceptors
```

guides/messaging/index.md
```diff
@@ -571,7 +571,7 @@ usually in combination with `cloudevents` format, as in this excerpt from a _pac
 [Learn more about `cds.env` profiles](../../node.js/cds-env#profiles){.learn-more}
 
 
-::: tip
+::: tip Learn more
```
renejeglinsky, line 574: Learn more if often used with our links. So I'd like to distinguish those learn more links from these notes. Maybe to differentiate from a "If you want to learn more" link, make it more engaging:
```suggestion
::: tip Read the guide
```
Just a suggestion, I'm open for your ideas. Your input is always much appreciated!

guides/messaging/index.md
```diff
@@ -610,7 +610,7 @@ const S4Bupa = await cds.connect.to ('API_BUSINESS_PARTNER')
 S4Bupa.on ('BusinessPartner.Changed', msg => {...})
 ```
 
-::: tip
+::: tip Learn more
```
renejeglinsky, line 613:
```suggestion
::: tip Read the guide
```

renejeglinsky (approved):

PR #625, opened 2024-01-15
Added section about Java plugin technique

BraunMatthias:


plugins/index.md
```diff
@@ -60,8 +60,31 @@ For Node.js all these plugins are implemented using the [`cds-plugin`](../node.j
...
+2. Add annotations to your model:
 
+   ```cds
+   annotate Customer with @PersonalData ...;
+   ```
+   > → audit logs are written to console in dev mode.
+   
+3. Bind the platform service.
 
+   > → audit logs are written to Audit Log V2 service.
```
renejeglinsky, line 86: I think here the platform service is meant. WDYT?

```suggestion
   > → audit logs are written to Audit Log service.
```

BraunMatthias: Actually this plugin demonstrated here is for V2 integration and NOT AuditlogService as in Node plugin

renejeglinsky: Discovery center knows only one version of the Audit Log service: https://discovery-center.cloud.sap/serviceCatalog/audit-log-service?region=all&tab=service_plan

BraunMatthias: In CAP we have technical Auditlog Service which can delegate to different audit log sinks such as console or BTP Audit log Service V2.

renejeglinsky: But what is Audit Log Service V2? There's only one service that also provides the premium plan. So I don't see the need to distinguish here. Maybe add a link to the Audit Log V2 handler for people who want to dive into the technical details?

BraunMatthias: Then let's name it BTP Audit Log Service to avoid confusion

renejeglinsky (approved):

PR #627, opened 2024-01-15
adoption of Cloud SDK 5 in CAP Java

StefanHenke:
**https://github.wdf.sap.corp/cds-java/home/issues/1251**

I´m wondering if it is really worth to spend a dedicated section in the migration guides. Alternatively, we could add the information about the additional dependency to the "minimum dependency version" page.

### DIFF
--- a/java/migration.md
+++ b/java/migration.md
@@ -1055,5 +1055,5 @@ CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For
 </dependency>
 ```
 
-In case your application makes use of Cloud SDK integration (`cds-integration-cloud-sdk`), no code changes in custom code should be necessary.
+In case your application uses the Cloud SDK integration package (`cds-integration-cloud-sdk`), no code changes in your custom code should be necessary.
 
--- a/java/migration.md
+++ b/java/migration.md
@@ -1055,4 +1055,6 @@ CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For
 </dependency>
 ```
 
+In case your application makes use of Cloud SDK integration (`cds-integration-cloud-sdk`), no code changes in custom code should be necessary.
+
 It is highly recommended to use at least version `5.2.0` of Cloud SDK.
--- a/java/migration.md
+++ b/java/migration.md
@@ -1043,3 +1043,16 @@ After rebuilding and restarting your application, your Application Services are 
 <!-- TODO: Move this to "Development" section -->
 
 <span id="afterenablingodata" />
+
+## Cloud SDK 4 to 5 { #cloudsdk}
+
+CAP Java 2.6.0 and higher is compatible with Cloud SDK in version 4 and 5. For reasons of backward compatibility, CAP Java assumes Cloud SDK 4 as the default. If you want to upgrade your CAP Java application to Cloud SDK 5, it is necessary to add the following maven dependency to your CAP Java application:
+
+```xml
+<dependency>
+	<groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+	<artifactId>connectivity-apache-httpclient4</artifactId>
+</dependency>
+	```
+
+It is highly recommended to use at least version 5.2.0 of Cloud SDK.
--- a/java/migration.md
+++ b/java/migration.md
@@ -1057,4 +1057,3 @@ CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For
 
 In case your application makes use of Cloud SDK integration (`cds-integration-cloud-sdk`), no code changes in custom code should be necessary.
 
-It is highly recommended to use at least version `5.2.0` of Cloud SDK.
--- a/java/migration.md
+++ b/java/migration.md
@@ -1046,7 +1046,7 @@ After rebuilding and restarting your application, your Application Services are 
 
 ## Cloud SDK 4 to 5 { #cloudsdk5 }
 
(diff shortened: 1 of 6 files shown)


BraunMatthias (approved):

java/migration.md
```diff
@@ -1043,3 +1043,18 @@ After rebuilding and restarting your application, your Application Services are
 <!-- TODO: Move this to "Development" section -->
 
 <span id="afterenablingodata" />
+
+## Cloud SDK 4 to 5 { #cloudsdk5 }
+
+CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For reasons of backward compatibility, CAP Java assumes Cloud SDK 4 as the default. If you want to upgrade your CAP Java application to Cloud SDK 5, it is necessary to add the following maven dependency to your CAP Java application:
```
renejeglinsky, line 1049: Why do we need to add exactly this dependency?
```suggestion
CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For reasons of backward compatibility, CAP Java assumes Cloud SDK 4 as the default. However, we highly recommended to use at least version `5.2.0` of Cloud SD. To upgrade your CAP Java application to Cloud SDK 5, add the following maven dependency to your CAP Java application:
```

java/migration.md
```diff
@@ -1043,3 +1043,18 @@ After rebuilding and restarting your application, your Application Services are
...
+```xml
+<dependency>
+	<groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+	<artifactId>connectivity-apache-httpclient4</artifactId>
+</dependency>
+```
+
+In case your application makes use of Cloud SDK integration (`cds-integration-cloud-sdk`), no code changes in custom code should be necessary.
+
+It is highly recommended to use at least version `5.2.0` of Cloud SDK.
```
renejeglinsky, line 1060:
```suggestion
```

java/migration.md
```diff
@@ -1043,3 +1043,18 @@ After rebuilding and restarting your application, your Application Services are
...
+CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For reasons of backward compatibility, CAP Java assumes Cloud SDK 4 as the default. If you want to upgrade your CAP Java application to Cloud SDK 5, it is necessary to add the following maven dependency to your CAP Java application:
+
+```xml
+<dependency>
+	<groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+	<artifactId>connectivity-apache-httpclient4</artifactId>
+</dependency>
+```
+
+In case your application makes use of Cloud SDK integration (`cds-integration-cloud-sdk`), no code changes in custom code should be necessary.
```
renejeglinsky, line 1058: Why do we mention those two packages explicitly? Are those heavily used by our stakeholders?
```suggestion
In case your application uses the Cloud SDK integration package (`cds-integration-cloud-sdk`), no code changes in your custom code should be necessary.
```

StefanHenke: It is actually only about one. It´s just the human readable name and its technical maven module name. We mention it here explicitly, it provides convenience features for integrating with Cloud SDK (which is an external dependency for us). This integration part is implemented by us and hence under our control. However, applications can also use Cloud SDK directly which means that they can also use APIs which have changed incompatible in Cloud SDK.

java/migration.md
```diff
@@ -1043,3 +1043,17 @@ After rebuilding and restarting your application, your Application Services are
 <!-- TODO: Move this to "Development" section -->
 
 <span id="afterenablingodata" />
+
+## Cloud SDK 4 to 5 { #cloudsdk5 }
+
+CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For reasons of backward compatibility, CAP Java assumes Cloud SDK 4 as the default. However, we highly recommended to use at least version `5.2.0` of Cloud SD. To upgrade your CAP Java application to Cloud SDK 5, add the following maven dependency to your CAP Java application:
```
renejeglinsky, line 1049:
```suggestion
CAP Java `2.6.0` and higher is compatible with Cloud SDK in version 4 and 5. For reasons of backward compatibility, CAP Java assumes Cloud SDK 4 as the default. However, we highly recommend to use at least version `5.2.0` of Cloud SDK. To upgrade your CAP Java application to Cloud SDK 5 you don't need to do anything if you're using the Cloud SDK integration package (`cds-integration-cloud-sdk`). If you're using Cloud SDK directly, add the following maven dependency to your CAP Java application:
```

renejeglinsky: Is that statement correct?

StefanHenke: `... ,in most cases, you  don't need to adapt any code if you rely on the Cloud SDK integration package (`cds-integration-cloud-sdk`). In these cases, it is sufficient to add the following maven dependency to your CAP Java application:`

You still have to add the dependency. If you rely on our integration package, there will be most likely no code change required.

java/migration.md
```diff
@@ -1043,3 +1043,17 @@ After rebuilding and restarting your application, your Application Services are
...
+
+```xml
+<dependency>
+	<groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+	<artifactId>connectivity-apache-httpclient4</artifactId>
+</dependency>
+```
+
+In case your application uses the Cloud SDK integration package (`cds-integration-cloud-sdk`), no code changes in your custom code should be necessary.
```
renejeglinsky, line 1058:
```suggestion
```

renejeglinsky (approved):

PR #630, opened 2024-01-15
Document limitations for Select ... lock()

vmikhailenko:
Add a summary and some examples of the cases when the `Select` statement with the `lock()` clause does not work.

### DIFF
--- a/java/query-api.md
+++ b/java/query-api.md
@@ -764,7 +764,7 @@ There are few notable examples of such restrictions:
 
 * You cannot use `lock()` and `distinct()` in the same statement.
 * Localized entities can be locked only if you manually reset the locale via dedicated request context without the locale as described here [Modifying Request Context](./request-contexts#modifying-requestcontext). 
-* Entities that contains "on-read" caclulated elements can't be locked when the statement contains statement that references them. Only non-calculated part of the entity can be used in the statement.
+* Entities that contains "on-read" calculated elements can't be locked when the statement contains statement that references them. Only non-calculated part of the entity can be used in the statement.
 
 As a general rule, prefer the statements that select primary keys with a simple conditions, such as `byId` or `matching`, to select the target entity set for locking.
 
--- a/java/query-api.md
+++ b/java/query-api.md
@@ -758,13 +758,13 @@ import static com.sap.cds.ql.cqn.CqnLock.Mode.SHARED;
 Select.from("bookshop.Books").byId(1).lock(SHARED);
 ```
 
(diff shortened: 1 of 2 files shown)


java/query-api.md
```diff
@@ -758,6 +758,16 @@ import static com.sap.cds.ql.cqn.CqnLock.Mode.SHARED;
 Select.from("bookshop.Books").byId(1).lock(SHARED);
 ```
 
+Not every database entity exposed via the CDS entity can be locked with the `lock()` clause. Databases require that the target of such statements is represented by a single table or a view that is based on a single table without aggregation, joining or coalescence of values and the statement itself must not introduce aggregations, for example, by adding `groupBy()` clauses. Some databases might have additional restrictions or limitations specific to them.  
+
+There are few notable examples of such restrictions: 
+
+* You cannot use `lock()` and `distinct()` in the same statement.
```
agoerler, line 765:
```suggestion
* You cannot use `lock()` with `distinct()` or `groupBy()`.
```

java/query-api.md
```diff
@@ -758,6 +758,16 @@ import static com.sap.cds.ql.cqn.CqnLock.Mode.SHARED;
...
+* Localized entities can be locked only if you manually reset the locale via dedicated request context without the locale as described in the chapter [Modifying Request Context](./request-contexts#modifying-requestcontext).
```
agoerler, line 766:
```suggestion
* Localized entities can be locked only if you the query is executed without a locale as described in the chapter [Modifying Request Context](./request-contexts#modifying-requestcontext). 
```

java/query-api.md
```diff
@@ -758,6 +758,16 @@ import static com.sap.cds.ql.cqn.CqnLock.Mode.SHARED;
 Select.from("bookshop.Books").byId(1).lock(SHARED);
 ```
 
+Not every database entity exposed via the CDS entity can be locked with the `lock()` clause. Databases require that the target of such statements is represented by a single table or a view that is based on a single table without aggregation, joining or coalescence of values and the statement itself must not introduce aggregations, for example, by adding `groupBy()` clauses. Some databases might have additional restrictions or limitations specific to them.
```
agoerler, line 761:
```suggestion
Not every entity exposed via the CDS entity can be locked with the `lock()` clause. Databases require that the target of such statements is represented by a single table or by a view that is simple enough so that the database can ambiguously identify which rows must be locked. On views, which are joining multiple tables, which are aggregating data, or which are computing expressions or coalescence of values `lock()` can't be executed. Some databases might have additional restrictions or limitations specific to them.  
```

agoerler (changes requested):

agoerler (approved):

renejeglinsky (approved):

PR #644, opened 2024-01-23
Fix CQN query link

mariayord:


SamuelBrucksch (approved):

node.js/events.md
```diff
@@ -345,7 +345,7 @@ const [ author, book ] = req.params
 
 ### . query {.property}
 
-Captures the incoming request as a [CQN](../cds/cqn) query object. For example, an HTTP request like `GET http://.../Books` is captured as follows:
+Captures the incoming request as a CQN [query](cds-ql#class-cds-ql-query). For example, an HTTP request like `GET http://.../Books` is captured as follows:
```
renejeglinsky, line 348: This link now points to a totally different section. Probably it was a very generic link on purpose? Would it make sense to add the more specific link in addition, maybe in a "Learn more..." fashion? If you say, the generic link doesn't help and this one is the link our users need, I'm fine. It's just a question I had when reviewing.

```suggestion
Captures the incoming request as a [CQN query](cds-ql#class-cds-ql-query). For example, an HTTP request like `GET http://.../Books` is captured as follows:
```

mariayord: @renejeglinsky The previous link was wrong , as the captured request is not a object, but a query. The cds.ql.Query API constructs a CQN object. 
For example the [SELECT](https://cap.cloud.sap/docs/cds/cqn#select) is an object with select properties. But [SELECT](https://cap.cloud.sap/docs/node.js/cds-ql#select) is an API with other properties as `.elements` or methods for example.

renejeglinsky: Thanks @mariayord !

renejeglinsky (approved):

PR #645, opened 2024-01-23
Docu for event as proj on event

stewsk:


bugwelle (approved):

cds/cdl.md
```diff
@@ -1716,6 +1714,15 @@ service MyOrders { ...
...
 
+An event can also be defined as projection on an entity, type, or another event.
+Only the effective signature of the projection is relevant.
+```cds
+service MyOrders { ...
+  event OrderCanceledNarrow : projection on OrderCanceled { orderID }
+}
+```
+
+
```
renejeglinsky, line 1725: Should we mention sth like this also here? I feel that this is not obvious but I might be wrong.
```suggestion
::: tip External event
When importing services, you can define an event in CAP as projection on the "external" event.
:::
```

hjboth: AFAIK, there are no 'external' events, at least not imported via API import (because events are not part of an OData API. Don't add this hint without further clafication. It can be added later anytime if correct.

hjboth (approved):
LGTM except for this hint which should not be added:

> When importing services, you can define an event in CAP as projection on the "external" event.

cds/cdl.md
renejeglinsky: Whatever we decide here, we need to adapt the release note as well.

renejeglinsky (approved):

PR #647, opened 2024-01-26
Update odata.md

renejeglinsky:


SamuelBrucksch (approved):

SamuelBrucksch:
Actually I think it also works for all input payloads in REST +  params and return types for functions (in REST) let me check again.

advanced/odata.md
```diff
@@ -1000,7 +1000,7 @@ The complex and structured types are deserialized to `java.util.Map`, whereas co
 
 ::: warning
 The full support of Open Types (`@open`) in OData is currently available for the Java Runtime only.
-The Node.js runtime supports the feature only in REST Adapter as well as for parameters and return types of actions and functions.
+The Node.js runtime supports the feature only for the REST Adapter when using parameters and return types of actions and functions.
```
smahati, line 1003: Are we changing the meaning here with this change? I read the previous documentation as: the "Open Types" are supported in the REST Adapter and also for parameters and return types of any actions and functions (not just in the REST Adapter IIUC). Thoughts?

PR #649, opened 2024-01-26
default length on db level

### LINKED ISSUE #614: Bug: Built-in Type - no input validation for default string length
Environment:
- Deployed on SAP BTP Cloud Foundry
- Database: SAP HANA Cloud
- CDS 7.5.2
- NodeJs v18.17.1

CAP does not validate the user input in Requests for String attributes where no length is specified:
As of the documentation (https://cap.cloud.sap/docs/cds/types), the default length for an attribute of type String with SAP HANA is 5000.
```
entity Test : cuid, managed {
   name: String;
   // ...
}
```
should be equal to
```
entity Test : cuid, managed {
   name: String(5000);
   // ...
}
```

However, the first example returns a 

> 500: Internal Server Error,

when creating a Test-Entity that has a name with length > 5000. The second example returns a 

> 400: Deserialization Error: Invalid value '...'. for property "name". The length of the Edm.String value must not be greater than the MaxLength facet value (5000).

The default length of 5000 is only set in SAP HANA and not validated in the application. Thus, creating a console output of

> [cds] - Error: inserted value too large for column: Failed in "NAME" column with the value

renejeglinsky:
Closes #614


cds/types.md
```diff
@@ -46,7 +46,7 @@ The following built-in types are provided:
 
 > <sup>(4)</sup> Up to 7 digits of fractional seconds; if a data is given with higher precision truncation may occur
 
-> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. If not set, default length 5000 is used for SAP HANA and default length 255 is used for all other SQL backends.
+> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. The default length used on the database level, is 5000 for SAP HANA and 255 for all other SQL backends.
```
renejeglinsky, line 49: I'm not sure if this is good enough to explain it. I somehow refrain from mentioning explicitly that input validation isn't supported because I don't know what else we then need to mention as excluded. If we do so, it should be a comprehensive list. Or we mention explicitly what is included, that's what I've tried to do. You're invited to improve it :)

danjoa: No ;) → I'll answer in the issue...

renejeglinsky:
```suggestion
> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. The default length is 5000 for SAP HANA and 255 for all other SQL backends. Note that default lengths are only applied on database level. Specify fixed lengths to get length checks on service level and/or inbound data.
```

stewsk (approved):

PR #651, opened 2024-01-28
SAP Cloud Logging is now supported

sjvans:


renejeglinsky (approved):

danjoa (commented):
Is that also in the release notes?

sjvans:
> Is that also in the release notes?

it sure is!

PR #652, opened 2024-01-29
fix code issue in otel custom instrumentation sample

StefanHenke:
Fixes a copy and paste issue in one of the open telemetry code samples which results in non-compilable code.


StefanHenke:
Included in https://github.com/cap-js/docs/pull/713

PR #660, opened 2024-01-29
Plugin support

renejeglinsky:


danjoa (changes requested):
I thought we agreed to add a single, eye catching, info at the top of the page, instead of repeating the same again and again.

renejeglinsky:
> I thought we agreed to add a single, eye catching, info at the top of the page, instead of repeating the same again and again. P.S. And we should also mention this in the Support section of our Resources page

I made the requested changes. To reuse as much as possible, I now redefined the icon such that it holds the link to the repository. In case we had documentation links on the icon, I moved that to a "Learn more". Right now the Java icon doesn't hold a link. I could add the link to the community there as well. That repetition wouldn't harm. WDYT?

danjoa (approved):

PR #662, opened 2024-01-30
Get Started: Updates

renejeglinsky:
Went through the guide and made some updates


swaldmann:
Fully agree with replacing the TIP heading as you did. However, the query options sample is important enough to the flow of the guide to remain top-level imo. I fear if included in the tip box many users will skip this part.

smahati:
The tip is relevant to the other part, so maybe we could move the information out of the tip and into the main text.

renejeglinsky:
That would be another option:
image

Again, I think this is very colorful. Is there a way to render the link differently?

renejeglinsky:
And the Toggle doesn't make it easier here. Java part looks like this:
image

If we get rid of the toogle, it would like as follows:
image
 
But here the question arises: Do we / Can we expect Java developers to read the (obvious) Node.js part?

renejeglinsky:
Switch from TIP to INFO makes it nicer IMO:

Node:

image

Java:
image

danjoa:
> Switch from TIP to INFO makes it nicer IMO:

No. this needs to stay an positive eye catcher → hence tip

swaldmann:
As the orange in the link + the green in the tip have semantic and not just ornamental value I think the coloring here is fine, even if a bit colorful.

Imo the tip is most helpful before the sentence explaining `$select` etc. for setting base expectations.

There's a bigger problem with the link, though: it's incorrect for both Node.js and Java. Node.js explicitly sets its path to `browse` without an `odata/v4` prefix in the CDS snippet. Java doesn't cut off the "Service" suffix → we need to check this

renejeglinsky:
If the sentence explaining `$select` is a continuation of the note, it should be in the note.
Anyway, will make changes and merge. At least paths and stuff are fixed.

get-started/in-a-nutshell.md
```diff
@@ -323,19 +322,22 @@ Both services defined above contain security annotations that restrict access to
...
 
+:::
+
 </div>
 
-::: tip
+You can even use advanced query options, such as `$select`, `$expand`, `$search`, and many more. For example, try out this link:
 
-CAP-based services are full-fledged OData services out of the box. Without adding any provider implementation code, they translate OData request into corresponding database requests, and return the results as OData responses.
+http://localhost:4004/odata/v4/browse/Books?$search=Brontë&$select=title,author&$expand=currency($select=code,name,symbol)&$orderby=title
```
renejeglinsky, line 331: At this point of the guide, we don't have any data yet. So this link doesn't show anything. Does it add value anyhow? Or should we move this section a bit?

get-started/in-a-nutshell.md
```diff
@@ -559,13 +574,15 @@ In Node.js, the easiest way to provide implementations for services is through e
...
 @Component
 @ServiceName(CatalogService_.CDS_NAME)
 public class CatalogHandler implements EventHandler {
   // your custom code will go here
 }
 ```
+:::
 
 ::: tip
 Place the code in your package of choice and use your IDE to generate the needed `import` statements.
```
renejeglinsky, line 588: @rjayasinghe Is this really what we want? Overall I think we need to add some hints at the beginning, for people starting with VS Code but having no Java extensions. I installed the Extension Pack for Java and then I was able to generate the needed imports (and fix them as I did it wrong).

get-started/in-a-nutshell.md
```diff
@@ -460,7 +461,7 @@ c.s.c.s.impl.persistence.CsvDataLoader   : Filling sap.capire.bookshop.Books fro
 
 Now that we've a connected, fully capable SQL database, filled with some initial data, we can send complex OData queries, served by the built-in generic providers:
 
-- _[…/Books?$select=ID,title](http://localhost:4004/odata/v4/browse/Books?$select=ID,title)_ {.impl .node}
+- _[…/Books?$select=ID,title](http://localhost:4004/browse/Books?$select=ID,title)_ {.impl .node}
```
alyonavyshnevska, line 464: Since we don't specify the path of the `admin` endpoint, it has the `odata/v4` prefix, while the `browse` endpoint doesn't. Pro: it showcases what happens if we don't specify the path. Con: inconsistency of the two endpoints. 
An alternative approach could be to specify the paths for both endpoints. Pro: we showcase `path` and it is less error-prone for the user to switch between endpoints.

renejeglinsky: Hi @alyonavyshnevska ,
I think that is made clear in line 301. Or at least a little. We don't explain, that if you just use `browse` instead of `/browse` that `odata/v4` prefix wouldn't be omitted. For what we want to tell here, I think it's fine as it is.

get-started/in-a-nutshell.md
```diff
@@ -677,6 +735,53 @@ public class SubmitOrderHandler implements EventHandler {
...
+        Select<Books_> byId = Select.from(cds.gen.catalogservice.Books_.class).byId(context.getBook());
+        Books book = persistenceService.run(byId).single().as(Books.class);
+        book.setStock(book.getStock() - context.getQuantity());
+
+        persistenceService.run(Update.entity(Books_.CDS_NAME).data(book));
+
+        context.setCompleted();
+    }
+}
+
```
renejeglinsky, line 780: @rjayasinghe Can we add that an error message is shown when stock reaches 0? Or if the next order would be larger than the stock available?

get-started/in-a-nutshell.md
```diff
@@ -677,6 +736,55 @@ public class SubmitOrderHandler implements EventHandler {
...
+    public SubmitOrderHandler(PersistenceService persistenceService) {
+        this.persistenceService = persistenceService;
+    }
+
+    @On()
+    public void onSubmitOrder(SubmitOrderContext context) {
+        Select<Books_> byId = Select.from(cds.gen.catalogservice.Books_.class).byId(context.getBook());
+        Books book = persistenceService.run(byId).single().as(Books.class);
+        if (context.getQuantity() > book.getStock())
+            throw new IllegalArgumentException(context.getQuantity() + " exceeds stock for book #" + book.getTitle());
```
renejeglinsky, line 775: @rjayasinghe Is that fine for you?

get-started/in-a-nutshell.md
```diff
@@ -479,11 +480,28 @@ Now that we've a connected, fully capable SQL database, filled with some initial
 
 ### Deploying Persistent Databases
 
-We can also use persistent instead of in-memory databases. For example, still with SQLite:
+We can also use persistent instead of in-memory databases. For example, still with SQLite, add the following configuration:
+
+> Add sqlite config from SQLite DB Guide
+> cds init could update @cap.js/sqlite dependency to ^1.4
```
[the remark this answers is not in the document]
renejeglinsky:
```suggestion
```

rjayasinghe (approved):

PR #667, opened 2024-01-31
[docs] cds compile command fix

d1gitalflow:
instead of `cds compile srv/cat-service --to sqlite > schema.sql`
(which does not work)

it should be:
`cds compile srv/cat-service --to sql > schema.sql`
(which correctly works)


renejeglinsky (approved):
Thanks a lot for your contribution! :)

PR #668, opened 2024-01-31
proposal for new streaming (BLOB Handling)

renejeglinsky:
Related to https://github.com/cap-js/docs/pull/590

### DIFF
--- a/cds/types.md
+++ b/cds/types.md
@@ -31,7 +31,7 @@ The following built-in types are provided:
 | `Timestamp` | _µs_ precision <sup>(4)</sup> | `'2021-06-27T14:52:23.123Z'` |  _TIMESTAMP_  |
 | `String` | (length ) <sup>(5)</sup> | `'hello world'` | _NVARCHAR_  |
 | `Binary` | (length) <sup>(5)</sup> | |  _VARBINARY_  |
-| `LargeBinary` |  | | _BLOB_ <sup>(7)</sup> |
+| `LargeBinary` |  | | _BLOB_ |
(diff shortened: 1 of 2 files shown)


cds/types.md
```diff
@@ -50,6 +50,8 @@ The following built-in types are provided:
 
 > <sup>(6)</sup> Mapping to ANSI SQL types are given for comparison. Note though, that you need to have the specification of your target database in mind when considering, for example, length restrictions.
 
+> <sup>(7)</sup> BLOBs hold potentially large amounts of data and handling them can be resource consuming (especially requiring the large amount of `LargeBinary` elements). Please consider using non-large `Binary` data type instead, which are returned as it is.
```
johannes-vogel, line 53: I think we should rewrite that section and recommend to add the stream annotation as also mentioned in the release notes. Most likely BLOBs are huge media objects. Only if they are small enough, they can fit in regular `Binary` data type.

guides/databases.md
```diff
@@ -345,6 +345,20 @@ db.queryForList("SELECT from sqlite_schema where name like ?", name);
 ```
 </div>
 
+### Reading `LargeBinary` / BLOB {.impl .node}
```
johannes-vogel, line 348: lgtm

cds/types.md
renejeglinsky: Would you rewrite it here? Or remove it here to not annotate LargeBinary with media objects in this generic types section? I definitely like the related release note! That is much clearer to me and better understandable.

renejeglinsky: Will remove this here and merge the rest. We can take this as a next step.

PR #675, opened 2024-02-06
Update core-services.md

mocsa:
For drafts-enabled applications we may need to use the drafts entity.


renejeglinsky:
Hi @mocsa ,
thanks a lot for your contribution. How to do this with draft, is explained in the [Fiori Support](https://cap.cloud.sap/docs/node.js/fiori) guide. Have you seen that? Usually we try to avoid duplications. Maybe sth like "For draft see ..." would be an option?
@johannes-vogel What do you think about adding this sentence here?

johannes-vogel:
a link would be ok, I think

renejeglinsky (approved):

PR #679, opened 2024-02-09
Add draft operations warning

mariayord:
I am not sure if this additional info should be in the warning part...


node.js/fiori.md
```diff
@@ -129,6 +129,8 @@ same editable properties without creating drafts for each row.
 :::warning
 Note that this feature creates additional entry points to your application. Custom handlers are triggered with delta
 payloads rather than the complete business object.
+
+Custom handlers need to take care that a draft entity doesn't exist, if a bound action/function is called on an active entity.
 :::
```
renejeglinsky, line 134:
```suggestion
:::warning Consider security effects
Note that this feature creates additional entry points to your application. 
:::

For custom handlers in this scenario you need to consider the following:

- They are triggered with delta payloads rather than the complete business object.

- They need to take care that a draft entity doesn't exist, if a bound action/function is called on an active entity.

```

David-Kunz (approved):

node.js/fiori.md
```diff
@@ -57,6 +57,10 @@ The examples are provided for `on` handlers, but the same is true for `before` a
 
 It's also possible to use the array variant to register a handler for both entities, for example: `srv.on('boundActionOrFunction', ['MyEntity', 'MyEntity.drafts'], /*...*/)`.
 
+:::warning
+If a bound action/function is called on an active entity, custom handlers need to take care that a draft entity doesn't exist.
+:::
```
renejeglinsky, line 62: This is only needed when the draft is bypassed, right?
```suggestion
:::warning When [bypassing SAP Fiori draft](#bypassing-the-sap-fiori-draft-flow) flow ...
If a bound action/function is called on an active entity, custom handlers need to take care that a draft entity doesn't exist.
:::
```

node.js/fiori.md
```diff
@@ -57,6 +57,10 @@ The examples are provided for `on` handlers, but the same is true for `before` a
 
 It's also possible to use the array variant to register a handler for both entities, for example: `srv.on('boundActionOrFunction', ['MyEntity', 'MyEntity.drafts'], /*...*/)`.
 
+:::warning
+If a bound action/function is called on an active entity, custom handlers need to take care that a draft entity doesn't exist.
```
David-Kunz, line 61:
```suggestion
If a bound action/function is called on an active entity, custom handlers need to take care that a draft entity doesn't exist if the action/function modifies the active instance, otherwise all changes are overridden when saving the draft.
```

node.js/fiori.md
David-Kunz: No, also without bypass draft.

node.js/fiori.md
mariayord: But the second part `if the action/function modifies the active instance` repeats the beginning `If a bound action/function is called on an active entity `

node.js/fiori.md
renejeglinsky: Thanks! I understand it now :)

renejeglinsky (approved):

PR #680, opened 2024-02-12
Java versions

BraunMatthias:
Added new section about Java versions and depenedencies


java/versions.md
```diff
@@ -0,0 +1,174 @@
...
+</dependencyManagement>
+```
+:::
+
+	
+### Update Versions
+
+Regular [updates and patches](#versions) of CAP Java keeps your project in sync with the most recent Free and Open Source Software (FOSS) dependency versions.
+
+However, a security vulnerability could be published in between CAP Java releases and in turn prevent your application from being released due to failing security scans.
```
renejeglinsky, line 153: In one of the dependencies?
```suggestion
However, a security vulnerability could be published, by one of your dependencies, in between CAP Java releases and in turn prevent your application from being released due to failing security scans. 
```

PR #681, opened 2024-02-13
link the plugin docu pages

rjayasinghe:
link from the node.js plugin docu to the java plugin docu and vice versa.

### DIFF
--- a/java/plugins.md
+++ b/java/plugins.md
@@ -22,6 +22,10 @@ Especially, when working with larger projects that may consist of many individua
 
 In the following sections, the different extension points and mechanisms are explained.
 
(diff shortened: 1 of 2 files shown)


renejeglinsky:
Has there been a customer request for this? I'd think the Plugins page is the tie for those topics

rjayasinghe:
> Has there been a customer request for this? I'd think the Plugins page is the tie for those topics

No request. Just my own idea. But I was not aware that the plugin page already has a link to both versions. So, for me it would be ok, if we close this PR. :)

renejeglinsky:
Thanks Robin. I'm also in favor of closing but thanks for thinking of ways to improve capire 👍

PR #686, opened 2024-02-15
Docs for new `sap.common.Timezone`

chgeo:
See cap/docs#593

### DIFF
--- a/cds/common.md
+++ b/cds/common.md
@@ -184,12 +184,12 @@ type Language : Association to sap.common.Languages;
 
 [It's the same as for `Country`.](#type-country){ .learn-more}
 
-### Type `TimeZone`
+### Type `Timezone`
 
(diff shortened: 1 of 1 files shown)


danjoa (dismissed):
Similar to timestamp, timezone is a single word, not two (see also [MDM dosc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset)) 
→ Please change all definitions to `Timezone` and adjust the docs

chgeo:
Well, the spelling is not that clear. 

IANA‘s time zone database (tzdata) and wiki use „time zone“ quite consistently:
- https://www.iana.org/time-zones
- https://en.m.wikipedia.org/wiki/Time_zone

Tendency in modern US English seems to go for the compound „timezone“, but not in plural „time zones“: https://en.m.wiktionary.org/wiki/timezone

PR #694, opened 2024-02-22
add new cds.fiori.draft_lock_timeout for configuration of draft lock

johannes-vogel:


node.js/fiori.md
```diff
@@ -83,9 +83,18 @@ For compatibility to previous variants, set `cds.fiori.draft_compat` to `true`.
...
 
 ```properties
-cds.drafts.cancellationTimeout=1h
+cds.fiori.draft_lock_timeout=30min
 ```
 
+You can set the property to one of the following:
+- number of hours like `'1h'` 
+- number of minutes like `'10min'`
+- number of milliseconds like `1000`
```
renejeglinsky, line 92: Is true also a possible value? Just as for `draft_deletion_timeout`?

```suggestion
You can set the property to one of the following:
- number of hours like `'1h'` 
- number of minutes like `'10min'`
- number of milliseconds like `1000`
```

johannes-vogel: true is already the default

renejeglinsky: Then it's similar to here, right? 
https://pages.github.tools.sap/cap/docs/node.js/fiori#garbage-collection-of-stale-drafts

renejeglinsky (approved):

Issue #657, opened 2024-01-29
Please correct in documentation caching behavior of cds.connect.to


dinurp:
In the documentation of [cds.connect.to (name, options?)](https://cap.cloud.sap/docs/node.js/cds-connect#cds-connect-to-name-options-%E2%86%92-service), it i stated that _subsequent connects with the same service name return the initially connected one_. This does not appear to be correct when the method is called with 2 arguments. Any cached instance is not returned. Instead a new instance is created and replaced in the cache.  Could you pleased update the documentation.

renejeglinsky:
Thanks @dinurp for opening this issue. We fixed that in the mentioned PR.
Cheers René

PR #698, opened 2024-02-22
fix: ad-hoc options are not cached

johannes-vogel:
Closes: #657

### DIFF
--- a/node.js/cds-connect.md
+++ b/node.js/cds-connect.md
@@ -191,7 +191,7 @@ _**Arguments:**_
 
 _**Caching:**_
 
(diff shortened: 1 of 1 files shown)


node.js/cds-connect.md
```diff
@@ -193,6 +193,8 @@ _**Caching:**_
 
 Service instances are cached in [`cds.services`](cds-facade#cds-services), thus subsequent connects with the same service name return the initially connected one. As services constructed by [`cds.serve`] are registered with [`cds.services`](cds-facade#cds-services) as well, a connect finds and returns them as local service connections.
```
renejeglinsky, line 194:
```suggestion
Service instances are cached in [`cds.services`](cds-facade#cds-services), thus subsequent connects with the same service name return the initially connected one. As services constructed by [`cds.serve`](cds-serve#cds-serve) are registered with [`cds.services`](cds-facade#cds-services) as well, a connect finds and returns them as local service connections.
```

renejeglinsky (approved):

PR #704, opened 2024-02-26
cds.linked evolution

danjoa:
@daogrady also fyi → we should add these to cds-types


daogrady:
Thanks for notifying me; I will [track it](https://github.com/cap-js/cds-typer/issues/168)

node.js/cds-reflect.md
```diff
@@ -422,6 +463,49 @@ class cds.type extends any {...}
...
+
+```js
+let [ Foo, Bar ] = Sue.entities	
+```
+
+```js
+for (let each of Sue.entities) console.log (each.name) 
+//> Sue.Foo
+//> Sue.Bar
+```
```
renejeglinsky, line 503:
```suggestion
```cds
service CatalogService {
  entity Books {}
  entity Authors {}
}
```

And got a reference to the service definition, e.g. like that:

```js
let { CatalogService } = cds.linked(model).definitions
```

You can reflect the entities in any of these ways: 

```js
let { Books, Authors } = CatalogService.entities
```

```js
let [ Books, Authors ] = CatalogService.entities	
```

```js
for (let each of CatalogService.entities) console.log (each.name) 
//> CatalogService.Books
//> CatalogService.Authors
```
```

**Review by renejeglinsky — APPROVED** [review]

File: node.js/cds-reflect.md [hunk 2]
```diff
@@ -359,21 +359,58 @@ All linked entity definitions are instances of this class, which primarily provi
...
+  key ID : UUID;
+  author : Association to Authors;
+  pages  : Composition of Pages;
+}
+```
+
+And got a reference to the entity definition, e.g. like that:
+
+```js
+let { Books } = cds.linked(model).definitions
```
David-Kunz, line 380: I'm afraid users will call `cds.linked(model)` manually, shouldn't we just refer to `cds.model.definitions` (already linked)?

node.js/cds-reflect.md
```diff
@@ -359,21 +359,58 @@ All linked entity definitions are instances of this class, which primarily provi
...
+```
+
+You can reflect the respective elements in any of these ways: 
+
+```js
+let { ID } = Books.keys, { author, pages } = Books.associations
+```
+
+```js
+let [ ID ] = Books.keys, [ author, pages ] = Books.associations
```
David-Kunz, line 390: I would skip this example, relying on the order of associations is a bit dangerous, I prefer the `for... of` loop below.

node.js/cds-reflect.md
danjoa: We here we are within the documentation of `cds.linked` and calling it manually is ok.

danjoa:
@renejeglinsky forgot tov mention: this is not for the current release but for the next one in March earliest → we should merge to main only _after_ our next capire release

PR #711, opened 2024-02-28
Update cds-ql.md

larslutz96:
Add warning:
When using the combination of $apply with groupby and only a subset of keys is included, it can lead to the result not being uniquely sorted anymore.


renejeglinsky:
As this guide has no mention of `$apply` I searched for other places where this information could fit better. How about this:
[advanced/odata#example-1](https://pages.github.tools.sap/cap/docs/advanced/odata#example-1)

advanced/odata.md
```diff
@@ -780,6 +780,10 @@ GET /Orders(10)/books?
 
 This request operates on the books of the order with ID 10. First it filters out the books from the year 2000 to an intermediate result set. The intermediate result set is then grouped by author name and the price is averaged. Finally, the result set is sorted by title and only the top 3 entries are retained.
 
+::: warning
+When using $apply with groupby and only a subset of keys is included, it can lead to the result not being uniquely sorted anymore.
```
johannes-vogel, line 784:
```suggestion
If the `groupby` transformation only includes a subset of the entity keys, the result order might be unstable.
```

renejeglinsky (approved):

PR #717, opened 2024-03-04
Support for CAP plugins

DanSchlachter:
distinguish the two organizations, clarify process a bit more

- [ ] CAP Plugin page should also mention version of  incident management with plugins included to easily try out plugins


plugins/index.md
```diff
@@ -6,15 +6,6 @@ status: released
 
 Following is a curated list of plugins that are available for the SAP Cloud Application Programming Model (CAP) which provide integration with SAP BTP services and technologies, or other SAP products.
 
-::: tip Maintained by CAP and SAP
-These plugins are created and maintained in close collaboration and shared ownership of CAP development teams and other SAP and BTP development teams.
-:::
```
renejeglinsky, line 11: I'd keep this tip to point out that all plugins on this page are CAP and SAP Plugins. Maybe we need to rephrase it for that purpose a little.

plugins/index.md
```diff
@@ -91,13 +82,29 @@ For [Audit Logging V2](../java/auditlog#handler-v2) it looks like this:
    
 ## Support for Plugins
 
+The regular CAP support process does not apply for plugins. Issues are to be created in the issue page of the respective github repository.
+
 |    |    |
 | --- | --- |
-| Node.js | Click the Node.js icon for the plugin to find the repository and open an issue there. 
+| Node.js | Click the Node.js icon of the plugin in the list below to find the repository and open an issue there.
```
renejeglinsky, line 89: When keeping the note above we can keep shorter here and avoid a linebreak:
image

```suggestion
| Node.js | Click the Node.js icon of the plugin to find the repository and open an issue there. 
```

plugins/index.md
```diff
@@ -91,13 +82,29 @@ For [Audit Logging V2](../java/auditlog#handler-v2) it looks like this:
...
+
 |    |    |
 | --- | --- |
-| Node.js | Click the Node.js icon for the plugin to find the repository and open an issue there. 
+| Node.js | Click the Node.js icon of the plugin in the list below to find the repository and open an issue there. 
 | Java | Use the community to [ask a question](https://community.sap.com/t5/forums/postpage/board-id/technology-questions/interaction-style/qanda). 
 
 <div id="internal-support" />
 
+Node.js plugins are available from two sources with different levels of support and maintenance:
```
renejeglinsky, line 94:
```suggestion
Node.js plugins are available with two different levels of support and maintenance:
```

renejeglinsky: Also avoiding the linebreak:
image

plugins/index.md
```diff
@@ -91,13 +82,29 @@ For [Audit Logging V2](../java/auditlog#handler-v2) it looks like this:
...
+[CAP JS](https://github.com/cap-js) plugins are:
+ - created and maintained in close collaboration and shared ownership of CAP development teams and other SAP and BTP development teams
+ - supported like CAP itself
+
+[CAP JS Community](https://github.com/cap-js-community) plugins are:
+ - created and maintained by the community
+ - not directly supported from CAP
+
+:::info Complete list of plugins
+As CAP is blessed with an active community, there are many useful plugins available created by the community. Have a look at the two communities to browse all available plugins.
```
renejeglinsky, line 105: What other community? If there's more than one, we should add a link. Maybe we should add one anyway.

plugins/index.md
```diff
@@ -91,13 +82,29 @@ For [Audit Logging V2](../java/auditlog#handler-v2) it looks like this:
    
 ## Support for Plugins
 
+The regular CAP support process does not apply for plugins. Issues are to be created in the issue page of the respective github repository.
```
renejeglinsky, line 85: This sentence is only valid for Node.js and can be found similarly in the following table.
```suggestion
The [regular CAP support process](/resources/#reporting-incidents) does not apply for plugins.
```

plugins/index.md
DanSchlachter: the plugin page shows a selection (not all) cap-js and no cap-js-community plugins, is that the case? Then the tip is correct and we can leave it in

plugins/index.md
DanSchlachter: The reason I put this in was that reading the page from top to bottom it is not clear what item I should click (as I haven't seen the list below yet) - do you think it is clear enough without it? 
Is "Click the icon of the plugin in the list to find the repository and open an issue there." better?

One problem I see as well is that cap-js-community plugins are not on the list so there you'd need to find the repo in a different way.

plugins/index.md
renejeglinsky: It shows cap-js and SAP-maintained cap-js-community plugins.

plugins/index.md
renejeglinsky: I see your point. How about that?
```suggestion
| Node.js | Click the Node.js icon of the plugin in the list below and open an issue there.
```

Or:

```suggestion
| Node.js | Click the Node.js icon of the plugin in the list below and open an issue on GitHub.
```

renejeglinsky: > One problem I see as well is that cap-js-community plugins are not on the list so there you'd need to find the repo in a different way.

I guess we need to assume that users can make this transfer and find the plugins on GitHub. Therefore adding a link to community in the info note could be helpful as well.

plugins/index.md
renejeglinsky: I think the goal should be to have all plugins from cap-js which are public (also in terms of repo=public) listed on this page.

I'll also discuss if we can include a link to https://bestofcapjs.org/ here. Needs just some legal clarification.

plugins/index.md
DanSchlachter: e.g. cds-typer is in cap-js and not on the list (and I understand why as it has its own section in the docu already) but then it's not clear its included into the support process here (or is it not?)

renejeglinsky:
@chgeo This PR is my proposal. I tried to keep it as condensed as possible. If you think some information, let's add it.

chgeo (approved):
I have condensed the section a bit, see 789b61c
Please check

renejeglinsky (approved):

PR #718, opened 2024-03-05
Add repl

renejeglinsky:


node.js/cds-compile.md
```diff
@@ -426,12 +427,15 @@ If no files are found, `undefined` is returned.
...
+> cds.env.roots             // + schema and services in cwd
+> cds.resolve('*',false)    // + models in cds.env.requires
+> cds.resolve('*')          // > the resolved existing files
+> cds.resolve(['db'])       // > the resolved existing files
+> cds.resolve(['db','srv']) // > the resolved existing files
+> cds.resolve('none')       // > undefined
 ```
 > Try this in cds repl launched from your project root to see that in action.
+
+*Question:* Note or repl in codeblock? Or both?
```
renejeglinsky, line 441: @smahati As I now included the repl part in the codeblock, do you think this additional note can be omitted? Or should we rather put in on every page, where we use repl to make people aware? So only on the first occurrence on a page for example?

renejeglinsky:
```suggestion
> Try this in cds repl launched from your project root to see that in action.
```

PR #720, opened 2024-03-05
try to adapt each after handler

renejeglinsky:
Here's a first try


renejeglinsky:
Done in https://github.com/cap-js/docs/pull/775

PR #728, opened 2024-03-07
Docs for IntelliJ plugin in new location

chgeo:
Requires cap-js/cds-intellij repo to be public


tim-sh (approved):
So we won't publish this now despite merging, right?

chgeo:
Yes, that’s my idea. But @renejeglinsky should have a final say when to merge this so that it doesn’t get accidentally published before the March release.

renejeglinsky:
Then let's first merge it mid of march so that I can make full releases of capire in the upcoming, let's say, 10 days.

PR #729, opened 2024-03-07
added limitations for compositions of one

etimr:
required in an external ticket


cds/cdl.md
```diff
@@ -690,6 +690,12 @@ entity Orders.Items {
 Essentially, Compositions are the same as _[associations](#associations)_, just with the additional information that this association represents a _contained-in_ relationship so the same syntax and rules apply in their base form.
 :::
 
+::: warning Limitations of Compositions of one
+Don't overuse Compositions of one. There is often no added value of using them as the information can be placed in the root entity. Compositions of one have limitations as follow:
+- Very limited Draft support. Fiori UIs almost never send Draft requests with Compositions of one. In rare cases when it happens, you need to take case of them in a custom handler.
```
David-Kunz, line 695:
```suggestion
- Very limited Draft support. Fiori elements does not support compositions of one unless you take case of their creation in a custom handler. 
```

David-Kunz: Note sure if `Compositions of one` or `compositions of one`, I tend to use the latter.

cds/cdl.md
```diff
@@ -690,6 +690,12 @@ entity Orders.Items {
 Essentially, Compositions are the same as _[associations](#associations)_, just with the additional information that this association represents a _contained-in_ relationship so the same syntax and rules apply in their base form.
 :::
 
+::: warning Limitations of Compositions of one
+Don't overuse Compositions of one. There is often no added value of using them as the information can be placed in the root entity. Compositions of one have limitations as follow:
+- Very limited Draft support. Fiori UIs almost never send Draft requests with Compositions of one. In rare cases when it happens, you need to take case of them in a custom handler. 
+- No support for POST requests over navigation with Compositions of one. If you can't avoid this scenario, please fill foreign keys manually in a custom handler.
```
David-Kunz, line 696:
```suggestion
- No support for requests over navigation to create Compositions of one. If you can't avoid this scenario, please fill foreign keys manually in a custom handler. 
```

David-Kunz: Note: End users can always send such requests, so users _must_ take care of this in custom handlers.

cds/cdl.md
stewsk: should be "take care of" instead of "take case of"

cds/cdl.md
etimr: @David-Kunz not sure about this change, in the case we discussed: `POST A(id)/b/c`, `b` is the composition of one, `c` is the composition of many, technically we are creating `c` and foreign keys must be filled for `c`.

nkaputnik (approved):
I would approve this from my side, but as a remark: I would go even further as to completely discourage the use of composition of one outside of legacy porting, meaning that if you don't have an existing DB model to map to, then this construct shouldn't be used at all. The existing formulation is a lot more friendly, ut also catches the gist of it

cds/cdl.md
David-Kunz: @etimr , I think it's not just `POST` but also `PATCH` requests (`UPSERT` behaviour), I think it's fine to keep it more general and say that they need to care care when navigating paths.

David-Kunz: I adapted the message a bit.

etimr:
@renejeglinsky , @smahati could you please take a look?

cds/cdl.md
```diff
@@ -690,6 +690,12 @@ entity Orders.Items {
 Essentially, Compositions are the same as _[associations](#associations)_, just with the additional information that this association represents a _contained-in_ relationship so the same syntax and rules apply in their base form.
 :::
 
+::: warning Limitations of Compositions of one
+Using of compositions of one is discouraged. There is often no added value of using them as the information can be placed in the root entity. Compositions of one have limitations as follow:
+- Very limited Draft support. Fiori elements does not support compositions of one unless you take care of their creation in a custom handler. 
+- No extensive support for modifications over paths if compostions of one are involved. You must fill in foreign keys manually in a custom handler.
```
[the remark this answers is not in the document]
etimr: Thank you for the feedback Adrian @agoerler, I've added the word `entity` to make limitations more clear.  
Steffen @stewsk, you are right, I'm writing about entities here.

renejeglinsky (approved):

PR #730, opened 2024-03-08
Clarify namespace, clarify relation service - entity

stewsk:
* a namespace is not an object
* relation entity/service is established via name


stewsk:
Hi @renejeglinsky , the checker complains because in the example I essentially put the content of two cds files in a single code snippet. Remedy is to either disable the syntax checker for this example or to split it into two code blocks. I'd like to leave this decision to you,

cds/cdl.md
```diff
@@ -1539,6 +1539,27 @@ GET: /OrderWithParameter(foo=5)/Set or GET: /OrderWithParameter(5)/Set
 GET: /ViewInService(p1=5, p2=true)/Set
 ```
 
+To expose an entity, it not necessarily has to be lexically enclosed in the service definition.
```
renejeglinsky, line 1542:
```suggestion
To expose an entity, it's not necessary to be lexically enclosed in the service definition.
```

cds/cdl.md
```diff
@@ -1539,6 +1539,27 @@ GET: /OrderWithParameter(foo=5)/Set or GET: /OrderWithParameter(5)/Set
...
+An entity's affiliation to a service is established using its fully qualified name.
+If the name of an entity contains the name of a service as prefix, it belongs to/is exposed by that service.
```
renejeglinsky, line 1544: Does using the namespace and using the prefix have the same result? "fully qualified name" does describe the namespace example, right?

stewsk: Yes, using the namespace and using the prefix have the same result.
```cds
namespace foo.bar;
entity moo { };
```
creates an Entity with "fully qualified name" `foo.bar.moo`.
```cds
entity foo.bar.moo { };
```
creates an Entity with "fully qualified name" `foo.bar.moo`.
```cds
namespace foo;
entity bar.moo { };
```
creates an Entity with "fully qualified name" `foo.bar.moo`.
```cds
namespace foo;
service bar {
  entity moo { };
}
```
creates an Entity with "fully qualified name" `foo.bar.moo`, plus the service `foo.bar`.

renejeglinsky: Thanks!

cds/cdl.md
```diff
@@ -1539,6 +1539,27 @@ GET: /OrderWithParameter(foo=5)/Set or GET: /OrderWithParameter(5)/Set
 GET: /ViewInService(p1=5, p2=true)/Set
 ```
 
+To expose an entity, it's not necessary to be lexically enclosed in the service definition.
+An entity's affiliation to a service is established using its fully qualified name.
+If the name of an entity contains the name of a service as prefix, it belongs to/is exposed by that service.
```
renejeglinsky, line 1544:
```suggestion
To expose an entity, it's not necessary to be lexically enclosed in the service definition. An entity's affiliation to a service is established using its fully qualified name, so you can also use one of the following options:

- Add a namespace.
- Use the service as prefix.
```

cds/cdl.md
```diff
@@ -1539,6 +1539,27 @@ GET: /OrderWithParameter(foo=5)/Set or GET: /OrderWithParameter(5)/Set
...
+In the following example, all entities belong to the same service:
```
renejeglinsky, line 1545:
```suggestion
In the following example, all entities belong to/are exposed by the same service:
```

renejeglinsky (approved):

PR #733, opened 2024-03-11
Java Outbox: Docs for the async service interface and ordered flag

t-bonk:
This documentation is for features for CAP Java delivered in cds-services 2.8.0


java/outbox.md
```diff
@@ -149,6 +150,28 @@ CqnService remoteS4 = ...;
...
+The API `OutboxService.outboxed(Service, Class)` wraps a service with an asynchronous suited API while outboxing it.
+This can be used together with the interface `AsyncCqnService` to outbox remote OData services:
+
+```java
+OutboxService myCustomOutbox = ...;
+CqnService remoteS4 = ...;
+AsyncCqnService outboxedS4 = myCustomOutbox.outboxed(remoteS4, AsyncCqnService.class);
+```
+
+The interface `AsyncCqnService` provides a convenience method to outbox CqnServices directly and wrapping the service with it:
```
beckermarc, line 162:
```suggestion
The interface `AsyncCqnService` provides a convenience method to outbox CqnServices:
```

java/outbox.md
```diff
@@ -149,6 +150,28 @@ CqnService remoteS4 = ...;
...
+```java
+OutboxService myCustomOutbox = ...;
+CqnService remoteS4 = ...;
+AsyncCqnService outboxedS4 = AsyncCqnService.of(remoteS4, myCustomOutbox);
+```
+
+::: warning Asynchronous suited API
+The service to be outboxed must provide the same method signatures that are declared in the passed interface.
+The result type is not relevant since outboxed services method calls don't return a result.
+:::
```
beckermarc, line 173: Maybe not a warning, but a more general tip for custom asynchronous suited API

java/outbox.md
```diff
@@ -214,6 +237,35 @@ Also the handler shall only be registered once on the outbox service.
 
 [Learn more about event handlers.](./event-handlers){.learn-more}
 
+## Handling Outbox Errors
```
beckermarc, line 240: Maybe separate this into a dedicated PR. I'd like to discuss this approach again, based on my other suggestion to allow explicitly calling underlying next ON handlers, which would be well-suited for an Outbox error handler as well.

java/outbox.md
```diff
@@ -149,6 +150,28 @@ CqnService remoteS4 = ...;
...
+```java
+OutboxService myCustomOutbox = ...;
+CqnService remoteS4 = ...;
+AsyncCqnService outboxedS4 = AsyncCqnService.of(remoteS4, myCustomOutbox);
+```
+
+::: tip Asynchronous suited API
+The service to be outboxed must provide the same method signatures that are declared in the passed interface.
+The result type is not relevant since outboxed services method calls don't return a result.
+:::
```
beckermarc, line 173:
```suggestion
::: tip Custom asynchronous suited API
When defining your own custom asynchronous suited API, the interface must provide the same method signatures as the interface of the outboxed service, except for the return types which should be `void`.
:::
```

beckermarc (approved):

java/outbox.md
```diff
@@ -78,6 +78,7 @@ You have the following configuration options:
 - `maxAttempts` (default `10`): The number of unsuccessful emits until the message is ignored. It still remains in the database table.
 - `storeLastError` (default `true`): If this flag is enabled, the last error that occurred, when trying to emit the message
 of an entry, is stored. The error is stored in the element `lastError` of the entity `cds.outbox.Messages`.
+- `ordered` (default `true`): If this flag is enabled, the outbox instance processes the entries in the order they have been submitted to it. Otherwise the outbox may process entries randomly and in parallel, by leveraging outbox processors running in multiple application instances.
 
 > Persistent outbox is supported starting with these versions: `@sap/cds: 5.7.0`,  `@sap/cds-compiler: 2.11.0` (`@sap/cds-dk: 4.7.0`)
```
renejeglinsky, line 83: How about removing this? These are the current versions:
image

```suggestion

```

java/outbox.md
```diff
@@ -144,6 +145,27 @@ CqnService remoteS4 = ...;
 CqnService outboxedS4 = myCustomOutbox.outboxed(remoteS4);
 ```
 
+The API `OutboxService.outboxed(Service, Class)` wraps a service with an asynchronous suited API while outboxing it.
```
renejeglinsky, line 148: What is the difference between "stored in the outbox first, and executed asynchronously" and an "asynchronous suited API"?

java/outbox.md
```diff
@@ -144,6 +145,27 @@ CqnService remoteS4 = ...;
...
+This can be used together with the interface `AsyncCqnService` to outbox remote OData services:
+
+```java
+OutboxService myCustomOutbox = ...;
+CqnService remoteS4 = ...;
+AsyncCqnService outboxedS4 = myCustomOutbox.outboxed(remoteS4, AsyncCqnService.class);
+```
+
+The interface `AsyncCqnService` provides a convenience method to outbox CqnServices:
```
renejeglinsky, line 157: At first glance I get the impression, that both those snippets do the same, as if this example is a convenient way to achieve the same as in the previous snippet. Is that correct? If not, where is the difference?

java/outbox.md
t-bonk: Example: If you outbox a CqnService, the run method for Insert, Upset, Update, Delete all have return values. Since each of those operations are executed asynchronously, the call to them immediately returns having a return value of null. In this example, an asynchronous suited API would have also the run methods mentioned above, but without a return type. Also Select doesn't make sense to be executed asynchronously ATM.

java/outbox.md
t-bonk: @renejeglinsky The second example I putting another API in front of the service to be outboxed. The magic is, that the interface AsyncCqnService needn't to be implemented by the service to be outbox, it just needs to declare a similar API. In the end it's different.

renejeglinsky: Thanks! I'd need some more explanation, but I'm sure our target group will get it :) If not the will let us know.

renejeglinsky (approved):

PR #738, opened 2024-03-12
Update core-services.md

renejeglinsky:
@johannes-vogel Could you explain to me what it means to "register handlers to run before handlers registered by others before"?
What if others also used `srv.prepend()`? Who are others?


johannes-vogel:
> @johannes-vogel Could you explain to me what it means to "register handlers to run before handlers registered by others before"? What if others also used `srv.prepend()`? Who are others?

I agree that the sentence is confusing.

What we try to express is that in some situations, there are already handlers registered to a service but you have to register additional handlers that must run before the ones that are already registered. Does that make sense?

node.js/core-services.md
```diff
@@ -457,7 +457,7 @@ Ensure to call `super.init()` to allow subclasses to register their handlers. Do
 function srv.prepend(()=>{...})
 ```
 
-Sometimes, you need to register handlers to run before handlers registered by others before. Use srv.prepend() to do so ´, for example like this:
+Sometimes, you need to register handlers to run before handlers registered by others before. Use `srv.prepend()` to do so, for example like this:
```
smahati, line 460:
```suggestion
If you need to register a handler that has to run before the existing ones, use `srv.prepend()` to do so, for example like this:
```
Would this be fine? @johannes-vogel

renejeglinsky: That sentence is clearer, thanks! What still isn't clear to me: What if there are multiple prepends on the same service? Is that not possible and execution fails? If not, how is the order of execution defined then?

renejeglinsky: @johannes-vogel I'll merge this as I think the sentence is better/clearer and definitely not wrong. If you want to adapt it, we an of course do so :)

PR #741, opened 2024-03-14
[cds^8] remove req._ (we have req.http now)

sjvans:


renejeglinsky:
Removed the label as we can now merge cds@8 PRs

johannes-vogel (approved):

node.js/events.md
```diff
@@ -79,15 +79,18 @@ In addition, you can access the current event context from wherever you are in y
...
 
 ```js
 this.on ('*', req => {
   let { res } = req.http
+  res.set('Content-Type', 'text/plain')
   res.send('Hello!')
 })
 ```
 
+Keep in mind that multiple requests (that is, instances of `cds.Request`) may share the same incoming http request and outgoing http response (for example, in case of an OData batch request).
```
renejeglinsky, line 92:
```suggestion
Keep in mind that multiple requests (that is, instances of `cds.Request`) may share the same incoming HTTP request and outgoing HTTP response (for example, in case of an OData batch request).
```

smahati (approved):

PR #754, opened 2024-03-18
Java outbox error handling

t-bonk:
This documentation is for features for CAP Java delivered in cds-services 2.8.0


java/outbox.md
```diff
@@ -209,6 +209,47 @@ Also the handler shall only be registered once on the outbox service.
 
 [Learn more about event handlers.](./event-handlers){.learn-more}
 
+## Handling Outbox Errors
```
BraunMatthias, line 212:
```suggestion
## Handling Outbox Errors { #handling-outbox-errors }
```

java/outbox.md
```diff
@@ -209,6 +209,47 @@ Also the handler shall only be registered once on the outbox service.
 
 [Learn more about event handlers.](./event-handlers){.learn-more}
 
+## Handling Outbox Errors { #handling-outbox-errors }
+
+Outbox errors can be handled by providing an `On` handler for an outbox and a target service.
```
beckermarc, line 214: I think we need to make clear that this approach only needs to be followed, if the original ON handler that handles the outbox message publishing is not under your control. If it is, it is much easier to add the try/catch there.

smahati:
After merging this we need to add a link to the release notes.

smahati:
@beckermarc I've reworded the handling, could you review it?

java/outbox.md
```diff
@@ -165,50 +165,92 @@ The default outbox services can be used for outboxing arbitrary CAP services. If
...
 
-## Technical Outbox API
+## Technical Outbox API { #technical-outbox-api }
 
-Outbox services provide the technical API `OutboxService.submit(String, OutboxMessage)` that can
-be used to send custom messages via outbox.
-When submitting a custom message, an `OutboxMessage` needs to be provided that can contain parameters for the
-event. As the `OutboxMessage` instance is serialized and stored in the database, all data provided in that message
+Outbox services provide the technical API `OutboxService.submit(String, OutboxMessage)` that can be used to outbox custom messages for an arbitrary event (processing logic).
+When submitting a custom message, an `OutboxMessage` that can contain parameters for the event needs to be provided.
```
smahati, line 171: when we say that "can" contain parameters, does it mean that they are optional or do we mean that it has requirements like being serializable?

java/outbox.md
```diff
@@ -165,50 +165,92 @@ The default outbox services can be used for outboxing arbitrary CAP services. If
...
 :::
 
-## Technical Outbox API
+## Technical Outbox API { #technical-outbox-api }
 
-Outbox services provide the technical API `OutboxService.submit(String, OutboxMessage)` that can
-be used to send custom messages via outbox.
-When submitting a custom message, an `OutboxMessage` needs to be provided that can contain parameters for the
-event. As the `OutboxMessage` instance is serialized and stored in the database, all data provided in that message
+Outbox services provide the technical API `OutboxService.submit(String, OutboxMessage)` that can be used to outbox custom messages for an arbitrary event (processing logic).
```
smahati, line 170: I'm not sure why "(processing logic)" is here, could you clarify?

beckermarc: Event and processing logic are kind of different names for the same thing here in this context. Further down we sometimes talk about processing logic, so that's why I wanted to connect this here as well. We could also rephrase to `to an arbitrary event or processing logic`

java/outbox.md
beckermarc: The OutboxMessage must be provided, but the parameters inside the outbox message are optional (it could be an empty message)

java/outbox.md
smahati: done :)

java/outbox.md
smahati: I made it more explicit in case there is someone who gets confused like me :)

smahati (approved):

PR #763, opened 2024-03-20
Warning HEX mode

agoerler:
Add warning that some queries might not be executable on the SAP HANA HEX engine


java/cqn-services/persistence-services.md
```diff
@@ -144,9 +144,11 @@ To generate SQL that is optimized for the new [HEX engine](https://help.sap.com/
 cds.sql.hana.optimizationMode: hex
 ```
 
-:::tip
 Use the [hints](../working-with-cql/query-execution#hana-hints) `hdb.USE_HEX_PLAN` and `hdb.NO_USE_HEX_PLAN` to overrule the configured optimization mode per statement.
-:::
+
+::: warning
+Generated SQL is typically compatible with the HEX engine. However, when using the `HEX` optimization mode some queries, especially when using [native HANA views](../../advanced/hana#create-native-sap-hana-object), may hit a limitation of the HEX engine and become nonexecutable. This is because the SAP HANA HEX engine has a limited feature set. In such cases, use the hint `hdb.NO_USE_HEX_PLAN` to bypass the HEX engine.
```
MattSchur, line 150: or
```suggestion
In some corner cases, particularly when using [native HANA views](../../advanced/hana#create-native-sap-hana-object), queries in `HEX` optimization mode may fail with a "hex enforced but cannot be selected" error. This is the case if the statement execution requires the combination of HEX only features with other features that are not yet supported by the HEX engine. To avoid such issues, make sure you're using the latest [SAP HANA Cloud version](https://help.sap.com/docs/hana-cloud/sap-hana-cloud-overview-guide/releases-and-upgrades-in-sap-hana-cloud). Until the missing feature is supported by the HEX engine, you can add a `hdb.NO_USE_HEX_PLAN` hint to the failing query, so the SQL generator doesn't use features that enforce the HEX engine.
```

smahati: This looks good, can we change the last "enforce" to "use" just to use a simpler word?

MattSchur: What about "require the HEX engine" instead of "enforce"?

smahati: Looks good :)

agoerler (commented):
LGTM (but I'm the author 😳)

MattSchur (approved):

smahati (approved):

smahati:
> LGTM (but I'm the author 😳)

PR #764, opened 2024-03-20
ChangeTracking (Java): Add detail on Fiori version and event handler

vmikhailenko:
Adds one missing precondition (Fiori version needs to be updated)
Adds an example of the event handler.


java/change-tracking.md
```diff
@@ -225,6 +227,30 @@ the root of the change is an order.
...
+## Reacting on Changes
+
+You can write the event handler to observe the change log entries. Keep in mind, that the change log entries 
+are created for each statement and this event will not be bound to any kind of transaction or a batch operation.
+
+```java
+import cds.gen.sap.changelog.Changes;
+
+@Component
+@ServiceName("ChangeTrackingService$Default")
```
BraunMatthias, line 239: We should make use of the constant for demo purpose here

java/change-tracking.md
```diff
@@ -225,6 +227,30 @@ the root of the change is an order.
...
+public class ChangeTrackingHandler implements EventHandler {
+	
+  @After(event = "createChanges")
```
BraunMatthias, line 242: dito

java/change-tracking.md
```diff
@@ -225,6 +227,30 @@ the root of the change is an order.
...
+  void afterCreate(EventContext context) {
+    Result result = (Result) context.get("result");
+	result.listOf(Changes.class).forEach(c -> {
```
BraunMatthias, line 245: indetation

java/change-tracking.md
vmikhailenko: First, we need to expose them. They are not available when dependency is included with runtime scope.

renejeglinsky:
@BraunMatthias Could you review and/or approve? Then we could include it in the release. Thanks :)

renejeglinsky (approved):

renejeglinsky (approved):

PR #775, opened 2024-03-26
Document `.after('each', ...)` handler

schwma:


java/event-handlers/request-contexts.md
```diff
@@ -146,8 +146,8 @@ In the following a few concrete examples are given:
...
-Therefore, the `after` handler needs to create a new Request Context. To achieve this, it's required to call `requestContext()` on the current `CdsRuntime` and use the `systemUser()` method to remove the named user from the new Request Context:
+The incoming JWT token triggers the creation of an initial RequestContext with a named user. Accesses to the database in the OData Adapter as well as the custom `On` handler are executed within <i>tenant1</i> and authorization checks are performed for user <i>JohnDoe</i>. An additionally defined `After` handler wants to call out to an external service using a technical user without propagating the named user <i>JohnDoe</i>.
+Therefore, the `After` handler needs to create a new Request Context. To achieve this, it's required to call `requestContext()` on the current `CdsRuntime` and use the `systemUser()` method to remove the named user from the new Request Context:
```
renejeglinsky, line 150: Could it be `.on` and so on as well?

johannes-vogel: I don't think the Java runtime is affected by this. Is there a need to adapt their docs?

johannes-vogel: > Could it be `.on` and so on as well?

No, that's specific to after handlers.

renejeglinsky: Ah, I didn't even realize that this is a Java topic. Sorry😓

schwma: Yes, this file is part of the Java docs and the actual change is Node.js only.

This was just something I noticed while I was adjusting the Node.js docs. The Java docs seems to use uppercase letters to refer to their handlers in most places (because of `@After`/`@On`). So I adjusted the capitalization here to be consistent.

node.js/core-services.md
```diff
@@ -601,12 +603,21 @@ Use this method to register handlers to run *after* the `.on` handlers, frequent
 - `results` — the outcomes of the `.on` handler which ran before
 - `req` — an instance of [`cds.Request`](./events.md#cds-request)
 
+As a convenience feature, `.after` handlers that are registered on the event `'each'` are called for each individual result entry on `'READ'`.
```
sjvans, line 606: @johannes-vogel @danjoa @schwma maybe add something like this?
```suggestion
::: warning
Asynchronous functions can be registered, but all `after` handlers are executed in parallel, which can lead to race conditions in case multiple handlers apply to the respective request. Hence, use with caution!
:::

As a convenience feature, `.after` handlers that are registered on the event `'each'` are called for each individual result entry on `'READ'`.
```

johannes-vogel: In addition, we should stress that `each` handlers have to be synchronous.

schwma: > In addition, we should stress that `each` handlers have to be synchronous. They are run in a loop, not Promise.all

That is done below in my changes in lines 608-610. However the warning currently is missing an explanation as to why, which I can add

node.js/core-services.md
```diff
@@ -601,12 +603,25 @@ Use this method to register handlers to run *after* the `.on` handlers, frequent
...
 
+::: warning
+Asynchronous functions can be registered, but all `.after` handlers are executed in parallel, which can lead to race conditions in case multiple handlers apply to the respective request. Hence, use with caution!
+:::
+
+As a convenience feature, `.after` handlers that are registered on the event `'each'` are called for each individual result entry on `'READ'`.
+
+::: warning
+Only synchronous functions are allowed to be registered as `.after('each',...)` handlers as they are run in a `.forEach()` loop without promise handling.
+:::
```
schwma, line 614: Note: two warnings in close proximity. Visually too much?

renejeglinsky: I definitely agree. Maybe we merge and publish and then think about how to fix that in a new PR?

renejeglinsky (approved):

PR #778, opened 2024-03-27
Warning: Avoid using distinct with expands

MattSchur:


java/working-with-cql/query-api.md
```diff
@@ -449,6 +449,11 @@ To expand all first level associations of an entity, use `expand()` on the entit
 ```java
 Select.from(BOOKS).columns(b -> b.expand());
 ```
+
+::: warning
+Avoid using `distinct` in queries with expands, since it only removes duplicate rows from the root entity and isn't defined for expands. Especially the execution of to-many expands typically requires selecting the (unique) keys of the parent entities to correctly map the expanded data to their parent rows. In some cases this can lead to an error which can be resolved by removing `distinct`.
```
agoerler, line 454: Maybe:

```suggestion
Avoid using `distinct` in queries with expands. `Distinct` removes duplicate rows from the root entity and hence effectively aggregates rows. However, expanding child entities from aggregated rows is not well-defined. Therefore, using `distinct` in queries with expands can lead to an error, which can be resolved by removing `distinct`.
```

agoerler (approved):

java/working-with-cql/query-api.md
```diff
@@ -449,6 +449,11 @@ To expand all first level associations of an entity, use `expand()` on the entit
 ```java
 Select.from(BOOKS).columns(b -> b.expand());
 ```
+
+::: warning
+Avoid using `distinct` in queries with expands. `Distinct` removes duplicate rows from the root entity and hence effectively aggregates rows. However, expanding child entities from aggregated rows is not well-defined. Therefore, using `distinct` in queries with expands can lead to an error, which can be resolved by removing `distinct`.
```
renejeglinsky, line 454:
```suggestion
Avoid using `distinct` in queries with expands. `Distinct` removes duplicate rows from the root entity and hence effectively aggregates rows. However, expanding child entities from aggregated rows is not well-defined. If you encounter errors using `distinct` in queries with expands, this can be resolved by removing `distinct`.
```

renejeglinsky (approved):

PR #781, opened 2024-03-28
Update best-practices.md

nocin:
Fixed incorrect js object syntax


renejeglinsky (approved):
Thanks for you contribution! Much appreciated :)

PR #783, opened 2024-03-28
Align fonts in SVGs

swaldmann:
Using `'Avenir Next', 'Inter var', 'Inter', Helvetica, Arial, sans-serif`


renejeglinsky:
I think you did that cleanup before, right? Does this happen when I (Windows) open those files and edit them? I thought it would also in that case use the fonts that you had configured. Just trying to avoid reverting your changes by mistake :)

renejeglinsky (approved):

PR #790, opened 2024-04-02
Fix remaining SVG fonts

swaldmann:
Found a few more… 

Currently looks quite ugly in Chrome, e.g.:
https://cap.cloud.sap/docs/guides/messaging/#in-process-eventing


renejeglinsky (approved):
Thanks!

PR #799, opened 2024-04-03
support binding of shared service instances

LotharBender:


chgeo (approved):

advanced/hybrid-testing.md
```diff
@@ -77,22 +78,73 @@ Only the information about **where the credentials can be obtained** is stored o
...
 
 ```log
 [bind] - Retrieving data from Cloud Foundry...
-[bind] - Binding my-ups to Cloud Foundry user provided service my-user-provided-service.
+[bind] - Binding redis to Cloud Foundry service redis-db.
 [bind] - Saving bindings to .cdsrc-private.json in profile hybrid.
 [bind] -
 [bind] - TIP: Run with cloud bindings: cds watch --profile hybrid
+
+```
```
renejeglinsky, line 115: For this the output is not really interesting, or? The more interesting part is the binding information, right?

advanced/hybrid-testing.md
```diff
@@ -77,22 +78,73 @@ Only the information about **where the credentials can be obtained** is stored o
...
+```json {5}[.cdsrc-private.json]
+{
+  "requires": {
+    "[hybrid]": {
+      "redis": {
+        "binding": {
+          "type": "cf",
+          "apiEndpoint": "https://api.sap.hana.ondemand.com",
+          "org": "shared-from-cf-org",
+          "space": "shared-from-cf-space",
```
renejeglinsky, line 126: Those are the interesting ones, right? 
```suggestion
          "org": "shared-from-cf-org", // [!code focus]
          "space": "shared-from-cf-space", // [!code focus]
```

advanced/hybrid-testing.md
LotharBender: Yes, the output is identical to standard managed service bindings

advanced/hybrid-testing.md
```diff
@@ -77,22 +78,63 @@ Only the information about **where the credentials can be obtained** is stored o
...
+    }
+  }
+}
+```
+:::
+
+`cds watch --profile hybrid` will automatically resolve shared service instance bindings using the correct org and space.
+
+::: info Not all services can be shared
+Only services that have the `shareable` flag in the metadata set to `true` can be shared. Use command `cf curl /v3/service_offerings` to read the service catalog metadata.
```
renejeglinsky, line 135: Is it possible to filter the response from the curl command to only return services that have `"sharebale":true`?

LotharBender: Good point, but according to the CF documentation `shareable` isn't a supported query parameter - https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#list-service-offerings

renejeglinsky: Maybe we don't need to point users on that at all. I mean we focus here on users who have access to a shared service instance and not on users sharing an instance, right? So maybe let's remove this note at all and add another sentence at the top like "On SAP BTP Cloud Foundry, service instances can be shared across orgs and spaces. If you have access to a shared service instance, you can also bind to a shared service instance just like any other service instance."

WDYT?

LotharBender: We can add this additional Info at the beginning, but I would leave this additional info at the end. On the one hand users might stumble over this topic (binding shared services instances) when reading the overall binding topic, on the other hand this curl API might not be known to users to figure out whether they can use this feature for  services they are using.

chgeo: > On SAP BTP Cloud Foundry, service instances can be shared across orgs and spaces. If you have access to a shared service instance, you can also bind to a shared service instance just like any other service instance.

Sounds good as a general hint, yes.

But as an FYI for users that want to share instances to others, the info box might still make sense imho.  We can make it a collapsible details block if you want.

advanced/hybrid-testing.md
```diff
@@ -77,22 +78,63 @@ Only the information about **where the credentials can be obtained** is stored o
...
+[bind] - Binding my to Cloud Foundry user provided service bookshop-ups. // [!code focus]
 [bind] - Saving bindings to .cdsrc-private.json in profile hybrid.
 [bind] -
 [bind] - TIP: Run with cloud bindings: cds watch --profile hybrid
 ```
 
-`cds watch --profile hybrid` will automatically resolve user-provided service instance bindings using the same technique as for any other managed service binding.
+#### Shared Service Instances on Cloud Foundry { #binding-shared-service-instances}
+
+Service instances can be shared across orgs and spaces.
```
renejeglinsky, line 98: On SAP BTP Cloud Foundry, service instances can be shared across orgs and spaces. If you have access to a shared service instance, you can also bind to a shared service instance just like any other service instance.

renejeglinsky, line 98:
```suggestion
On SAP BTP Cloud Foundry, service instances can be shared across orgs and spaces. If you have access to a shared service instance, you can also bind to a shared service instance just like any other service instance.
```

renejeglinsky (approved):

PR #804, opened 2024-04-06
Embed asciinema player

chgeo:


chgeo:
Will do this in the other (SAP) repo, as recordings are not used in this content.

PR #808, opened 2024-04-09
Typer docs: add missing comma

alyonavyshnevska:
- Added a missing comma in a snippet
- specified what comes after the `#cds-models` as it may be not 100% clear to people working with the cds-typer for the first time

### DIFF
--- a/tools/cds-typer.md
+++ b/tools/cds-typer.md
@@ -22,11 +22,11 @@ The following chapter describes the [`cds-typer` package](https://www.npmjs.com/
 ```js
 //  without cds-typer
 const { Books } = cds.entities(…)
-service.before('CREATE' Books, ({ data }) => { /* data is of type any */})
+service.before('CREATE', Books, ({ data }) => { /* data is of type any */})
 
 // ✨ with cds-typer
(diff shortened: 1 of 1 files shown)


tools/cds-typer.md
```diff
@@ -22,11 +22,11 @@ The following chapter describes the [`cds-typer` package](https://www.npmjs.com/
 ```js
 //  without cds-typer
 const { Books } = cds.entities(…)
-service.before('CREATE' Books, ({ data }) => { /* data is of type any */})
+service.before('CREATE', Books, ({ data }) => { /* data is of type any */})
 
 // ✨ with cds-typer
-const { Books } = require('#cds-models/…')
-service.before('CREATE' Books, ({ data }) => { /* data is of type Books */})
+const { Books } = require('#cds-models/myService')
```
daogrady, line 28:
```suggestion
const { Books } = require('#cds-models/bookshop')
```

daogrady (approved):
Thanks for the valuable fresh-eyed feedback! I agree with having an explicit model path, but let's stick with CAPire's running sample of using `bookshop`. 🙂
Other than that, approval from my side.

PR #836, opened 2024-04-15
Fixes wrong class in code sample for observability

vmikhailenko:
Use correct class name `Meter` for the code sample.

### DIFF
--- a/java/operating-applications/observability.md
+++ b/java/operating-applications/observability.md
@@ -417,7 +417,7 @@ Similarly, you can record metrics during execution of, for example, a custom eve
 @Component
 @ServiceName(CatalogService_.CDS_NAME)
 class CatalogServiceHandler implements EventHandler {
-   Metric meter = GlobalOpenTelemetry.getMeterProvider().meterBuilder("RatingCalculator").build();
+   Meter meter = GlobalOpenTelemetry.getMeterProvider().meterBuilder("RatingCalculator").build();
 
    @After(entity = Books_.CDS_NAME)
    public void afterAddReview(AddReviewContext context) {


StefanHenke (approved):

StefanHenke:
@vmikhailenko: The dynatrace chapter is separate because you can use Dynatrace even without Open Telemetry. If we integrated the content into the telemetry chapter, people might loose this fact.

However, what I noticed is that the Dynatrace chapter only links to the official, externally available documentation on help.sap.com. For external customers, Dynatrace is not offered as a managed service on BTP. This is only for internal ones which is not mentioned at the moment. Should we add this with a corresponding (internal) link?

vmikhailenko:
> @vmikhailenko: The dynatrace chapter is separate because you can use Dynatrace even without Open Telemetry. If we integrated the content into the telemetry chapter, people might loose this fact.
> 
> However, what I noticed is that the Dynatrace chapter only links to the official, externally available documentation on help.sap.com. For external customers, Dynatrace is not offered as a managed service on BTP. This is only for internal ones which is not mentioned at the moment. Should we add this with a corresponding (internal) link?

I asked the internal Dynatrace docu to link to us.

vmikhailenko:
@StefanHenke How about linking this https://opentelemetry.io/docs/languages/java/instrumentation/#initialize-the-sdk?

`GlobalOpenTelemetry` is discouraged by OpenTelemetry. Maybe we should mention that to prevent the people from simply copy-pasting the snippet?

StefanHenke:
@vmikhailenko: We have to be careful here because the OpenTelemetry config is actually injected by the Java Agent which can  collide with the sdk. We need to ensure that we do not configure Open Telemetry a second time which would result in two parallel configs. At least I failed in the past to avoid `GlobalOpenTelemetry`.

renejeglinsky:
> @vmikhailenko: The dynatrace chapter is separate because you can use Dynatrace even without Open Telemetry. If we integrated the content into the telemetry chapter, people might loose this fact.
> 
> However, what I noticed is that the Dynatrace chapter only links to the official, externally available documentation on help.sap.com. For external customers, Dynatrace is not offered as a managed service on BTP. This is only for internal ones which is not mentioned at the moment. Should we add this with a corresponding (internal) link?

Should we also add an internal link pointing to some internal documentation?

renejeglinsky (approved):

StefanHenke:
> > @vmikhailenko: The dynatrace chapter is separate because you can use Dynatrace even without Open Telemetry. If we integrated the content into the telemetry chapter, people might loose this fact.
> > However, what I noticed is that the Dynatrace chapter only links to the official, externally available documentation on help.sap.com. For external customers, Dynatrace is not offered as a managed service on BTP. This is only for internal ones which is not mentioned at the moment. Should we add this with a corresponding (internal) link?
> 
> Should we also add an internal link pointing to some internal documentation?

Today, we don´t have an internal documentation section for dynatrace. What would be the best way to just add a link for internal usage: https://pages.github.tools.sap/apm/docs/

renejeglinsky:
> > > @vmikhailenko: The dynatrace chapter is separate because you can use Dynatrace even without Open Telemetry. If we integrated the content into the telemetry chapter, people might loose this fact.
> > > However, what I noticed is that the Dynatrace chapter only links to the official, externally available documentation on help.sap.com. For external customers, Dynatrace is not offered as a managed service on BTP. This is only for internal ones which is not mentioned at the moment. Should we add this with a corresponding (internal) link?
> > 
> > 
> > Should we also add an internal link pointing to some internal documentation?
> 
> Today, we don´t have an internal documentation section for dynatrace. What would be the best way to just add a link for internal usage?

Will reach out to you. I'll merge this PR as it's initial goal is achieved.

PR #839, opened 2024-04-16
Add example for selecting from view with parameters

stewsk:


cds/cdl.md
```diff
@@ -1527,7 +1539,7 @@ service MyOrders {
   entity OrderWithParameter( foo: Integer ) as select from data.Orders where id=:foo;
 }
 ```
-A [`view with parameter`](#views-with-parameters) modeled in the previous example, can be exposed as follows:
+A parametrized view like modeled in the section on [`view with parameter`](#views-with-parameters) can be exposed as follows:
```
stewsk, line 1542: @renejeglinsky : i tried to clarify the sentence referring to a previous example that stumbled about yesterday ...

stewsk:
@renejeglinsky : this is the example on using a view with parameters that we talked about yesterday.
I added a TODO where we could link to respective sections of the runtime docu where they show how to do a select on a view with parameters in the runtimes - although I don't know whether such examples do exist

@agoerler @johannes-vogel : do you have docu/examples showing how to select from a view w/ params in the runtimes that we could link here? This was explicitly asked for in a cap issue.

cds/cdl.md
renejeglinsky: Thanks, that is much better :)

johannes-vogel:
> @renejeglinsky : this is the example on using a view with parameters that we talked about yesterday. I added a TODO where we could link to respective sections of the runtime docu where they show how to do a select on a view with parameters in the runtimes - although I don't know whether such examples do exist
> 
> @agoerler @johannes-vogel : do you have docu/examples showing how to select from a view w/ params in the runtimes that we could link here? This was explicitly asked for in a cap issue.

We do not have a sample as we do not have a proper API for that. As of now, it's basically a CQN snippet you need to provide `SELECT.from({ id: 'ViewWithParams', args: { p1: {val: 1}, p2: {val:2} })`.

cds/cdl.md
```diff
@@ -552,12 +552,24 @@ By using a cast, annotations and other properties are inherited from the provide
...
 ```
+
+When selecting from a view with parameters, the parameters are passed by name.
+In the following example, `UsingView` also has a parameter `bar` that is passed down to `SomeView`.
+
+```cds
+entity UsingView ( bar: Boolean )
+as SELECT * from SomeView(foo: 17, bar: :bar);
+```
+
```
renejeglinsky, line 570: @johannes-vogel Could you translate your example on this example? Does it make a difference if it's a view with param or a view selecting from a view with params?

renejeglinsky: I mean this example you added in a comment: 
`SELECT.from({ id: 'ViewWithParams', args: { p1: {val: 1}, p2: {val:2} })`

renejeglinsky:
```suggestion
For Node.js, there's no programmatic API yet. You need to provide a [CQN snippet](/cds/cqn#select).
```

renejeglinsky: @johannes-vogel @MattSchur @agoerler Is the previous suggestion fine for you?

cds/cdl.md
```diff
@@ -552,12 +552,24 @@ By using a cast, annotations and other properties are inherited from the provide
...
+**TODO** add links to how to write queries on views w/ parameters in the runtimes.
```
MattSchur, line 571: or
```suggestion
In CAP Java, run a select statement against the view with named [parameter values](/java/working-with-cql/query-execution#querying-views):
```Java
var params = Map.of("bar", true);
Result result = service.run(Select.from("UsingView"), params);
```
```

MattSchur: -> https://github.com/cap-js/docs/pull/909

renejeglinsky: @johannes-vogel If you'd provide a similar specific example, I'd put both examples in a Node/Java code group.

renejeglinsky:
I think I added everything so that you can do another review please.

renejeglinsky:
Can everyone do another review please?

MattSchur (approved):

renejeglinsky (approved):

renejeglinsky:
@johannes-vogel Is this fine for you?

johannes-vogel (approved):

PR #842, opened 2024-04-17
fix missing bracket in anon ping server.js sample

qmacro:
The sample code was just missing a final closing bracket. I only noticed because I used the code to try out the custom express middleware sample :-)


smahati (approved):
Thank you for fixing it :)

PR #844, opened 2024-04-17
Common badges for alpha, beta, and required versions

chgeo:
Replace generic `<Badge .../>` with `<Alpha/>` and `<Beta/>`: 

image

image

Also have a new `<Since />` badge to specify min. version requirements:

image

### DIFF
--- a/.vitepress/theme/components/Alpha.vue
+++ b/.vitepress/theme/components/Alpha.vue
@@ -2,6 +2,6 @@
   <Badge
     type="warning"
     text="alpha"
-    title="This is an alpha feature. Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change."
+    title="Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change."
   />
 </template>
--- a/.vitepress/theme/components/Alpha.vue
+++ b/.vitepress/theme/components/Alpha.vue
@@ -1,0 +1,7 @@
+<template>
+  <Badge
+    type="warning"
+    text="alpha"
+    title="This is an alpha feature. Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change."
+  />
+</template>
--- a/.vitepress/theme/components/Beta.vue
+++ b/.vitepress/theme/components/Beta.vue
@@ -1,0 +1,7 @@
+<template>
+  <Badge
+    type="warning"
+    text="beta"
+    title="This is a beta feature. Beta features aren't part of the officially delivered scope that SAP guarantees for future releases."
+  />
+</template>
--- a/.vitepress/theme/components/Beta.vue
+++ b/.vitepress/theme/components/Beta.vue
@@ -2,6 +2,6 @@
   <Badge
     type="warning"
     text="beta"
-    title="This is a beta feature. Beta features aren't part of the officially delivered scope that SAP guarantees for future releases."
+    title="Beta features are planned to be generally available in subsequent releases, however, APIs and their behavior are not final and may change in the general release."
   />
 </template>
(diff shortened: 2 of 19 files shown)


chgeo:
@LotharBender FYI

about/index.md
```diff
@@ -579,6 +579,11 @@ Not an official product name, though.
...
-- <Badge type="warning" text="beta" /> - Beta features are planned to be generally available in subsequent releases, however, APIs and their behavior are not final and may change in the general release.
 
-- <Badge type="warning" text="alpha" /> - Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change.
+#### Badges
+
+- <Alpha /> — Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change.
+
+- <Beta /> — Beta features are planned to be generally available in subsequent releases, however, APIs and their behavior are not final and may change in the general release.
+
+- <Since version="1.2.3" of="@sap/..." /> — The marked feature is only available with the given version or higher.
```
chgeo, line 589: @renejeglinsky @smahati Please check on the texts here + the `title` texts in the `.vue` files.

renejeglinsky: _Since_ would just be usable for things that have already been delivered. If we'd want to have a mechanism to say that this doc is only valid from an upcoming version going forward, we'd need to do sth different. But maybe both is an edge case and the "future" badge is even less often I'd guess.
Using "as of" is defined in Merriam Webster as "used to indicate a time or date at which something begins or ends". So could fit in both scenarios.

chgeo: I see your point with the upcoming releases, but I doubt that users would see a difference in _since_ vs. _requires_ vs. _as of_.  They would not find that version and then hopefully get the point and not create a ticket for us.

An alternative would be to specifically use a text like _as of upcoming version_, but this we would have to change again once we publish.  Too much effort imho.

But we could change it to _as of_ if you like.

swaldmann (approved):

about/index.md
smahati: I like requires best because that seems to be the best one to convey the intent (it won't work without this), If we want to make it seem more like "requires at least" maybe  we can use "minimum", like "minimum @sap/cds-dk:V7.5". wdyt?

.vitepress/theme/components/Beta.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+  <Badge
+    type="warning"
+    text="beta"
+    title="This is a beta feature. Beta features aren't part of the officially delivered scope that SAP guarantees for future releases."
```
smahati, line 5:
```suggestion
    title="Beta features are planned to be generally available in subsequent releases, however, APIs and their behavior are not final and may change in the general release."
```
This is what we said in the glossary so maybe we can keep it the same?

.vitepress/theme/components/Alpha.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+  <Badge
+    type="warning"
+    text="alpha"
+    title="This is an alpha feature. Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change."
```
smahati, line 5:
```suggestion
    title="Alpha features are experimental. They may never be generally available. If released subsequently, the APIs and behavior might change."
```

.vitepress/theme/components/Beta.vue
smahati, line 5:
```suggestion
    title="Beta features are planned to be generally available in subsequent releases, however, APIs and their behavior are not final and may change in the general release."
```

smahati (changes requested):

about/index.md
chgeo: I have changed it to _requires_

smahati (approved):

PR #845, opened 2024-04-17
docs(localized): Improve documentation on helper views: mention casts

bugwelle:
It may be necessary to cast an element to `localized`, e.g. if no other element is localized.  Because properties are not propagated from elements to expressions, casts may be necessary. I've added an example where this is the case.

### DIFF
--- a/guides/localized-data.md
+++ b/guides/localized-data.md
@@ -59,10 +59,6 @@ entity Books.texts {
 ```
 
 [See the definition of `sap.common.Locale`.](../cds/common#locale-type){ .learn-more}
-::: warning Note:
-The above shows the situation with CDS compiler v2 and later. Former versions of the
-compiler generated an entity `Books_texts`.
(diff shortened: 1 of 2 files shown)


guides/localized-data.md
```diff
@@ -60,7 +60,7 @@ entity Books.texts {
 
 [See the definition of `sap.common.Locale`.](../cds/common#locale-type){ .learn-more}
 ::: warning Note:
-The above shows the situation with CDS compiler v2. Former versions of the
+The above shows the situation with CDS compiler v2 and later. Former versions of the
 compiler generated an entity `Books_texts`.
 :::
```
renejeglinsky, line 65: I vote for removing this note latest with the next major. v1 should then really be out of the picture 😅

stewsk (approved):

renejeglinsky (approved):

PR #858, opened 2024-04-24
Align `pre.log` styling

swaldmann:
Our `pre.log` styling currently isn't nicely aligned with built-in code fences, which becomes apparent in the "Tools" page for instance, where they are both used. This PR aligns the styling of `pre.log` fences with Vitepress ones.

Before:
Screenshot 2024-04-24 at 13 58 56

After:
Screenshot 2024-04-24 at 13 58 47

### DIFF
--- a/tools/index.md
+++ b/tools/index.md
@@ -30,6 +30,7 @@ Use `cds version` to get information about your installed package version:
 
 <pre class="log">
 <span class="cwd">$</span> <span class="cmd">cds</span> <span class="args">version</span>
+
 <em>@capire/samples:</em> 2.0.0
 <em>@sap/cds:</em> 6.7.0
 <em>@sap/cds-compiler:</em> 3.8.2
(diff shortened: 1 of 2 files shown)


chgeo (approved):

tools/index.md
```diff
@@ -29,7 +29,8 @@ npm i -g @sap/cds-dk
 Use `cds version` to get information about your installed package version:
 
 <pre class="log">
-<i>$</i> cds version
```
renejeglinsky, line 32: This is much more sophisticated. But didn't we use it until now? Is it the effort?

swaldmann: Yeah it's a bit of a hack/misuse of `i` here, which semantically is just an italic text in HTML. Probably just for readability/effort.

swaldmann: We shouldn't need it for the `log` language in a code fence, but for occurrences of `<pre class="log">` it could improve the styling.

I'm also in favor of changing CAPire-wide. I just changed it here in case Daniel disagrees next week :)

renejeglinsky: I agree. Then let's merge :)

renejeglinsky (approved):

PR #859, opened 2024-04-24
Expressions as annotation values

stewsk:
... and their handling in OData


stewsk:
@hjboth @bugwelle - could you please have a look?

cds/cdl.md
```diff
@@ -1003,35 +1003,87 @@ For example, for SAP Fiori models, it's the _4odata_ and _2edm(x)_ processors.
...
+```
+
+* `BooksList` would inherit annotations from `Books`
+* `BooksList.ID` would inherit from `Books.ID`
+* `BooksList.author` would inherit from `Books.author.name`
+* `BooksList.genre` would inherit from type `Genre`
+
+The rules are:
+
+1. Entity-level properties and annotations are inherited from the **primary** underlying source entity — here `Books`.
```
bugwelle, line 1057: Yes and no. If you have sub-queries as the value for FROM, it depends how you'd call that. It's the "leading query".

```cds
@anno
entity E { key ID : String; };
entity P as SELECT FROM (SELECT FROM E);
```

Can we call `E` in query `P` a "primary source"?

cds/cdl.md
```diff
@@ -1003,35 +1003,87 @@ For example, for SAP Fiori models, it's the _4odata_ and _2edm(x)_ processors.
...
+
+::: tip
+Propagation of annotations can be stopped via value `null`, for example, `@anno: null`.
 :::
 
+
+### Expressions as Annotation Values <Beta /> {#expressions-as-annotation-values}
+
 In order to use an expression as an annotation value, it must be enclosed in parentheses:
-<!-- cds-mode: upcoming, cds-compiler v4.5 -->
```
bugwelle, line 1017: Oh, thanks! I forgot to remove these.

cds/cdl.md
```diff
@@ -1070,54 +1122,73 @@ In CSN, the expression is represented as a record with two properties:
...
+  text as descr
 }
 ```
 
-All three annotations are propagated to the projection `Rectangle`. The propagated `@MyLength` is still valid.
-The propagated annotation `MyArea` is invalid, as the referenced element `depth` has been renamed to `width`.
-The propagated annotation `MyHeight` is invalid, as element `height` of `Block` has not been projected at all.
-This results in a compiler error. To make it work, you would have to explicitly overwrite annotations
-`@MyArea` and `@MyHeight` at `Rectangle`.
+Resulting CSN:
```
bugwelle, line 1150: This page is about CDL, do we really want such a long and verbose CSN snippet? I'd favor just mentioning what `P:descr`'s annotation's value is.

cds/cdl.md
```diff
@@ -1070,54 +1122,73 @@ In CSN, the expression is represented as a record with two properties:
...
+        },
+        "descr": {"type": "cds.String"}
+      }
+} } }
+```
+
+::: info
+
+There are situations where automatic rewriting doesn't work (yet).
+Some of them are going to be addressed in upcoming releases.
```
bugwelle, line 1181: You could point the user to https://cap.cloud.sap/docs/cds/compiler/messages#anno-missing-rewrite

advanced/odata.md
```diff
@@ -531,6 +535,192 @@ The second example is for a (record type) term in the [Communication vocabulary]
...
+
+While the flattening of references described in the section above is applied to all
+annotations, the syntactic translation of expressions is only done for annotations
+defined in one of the [OData vocabularies](#vocabularies).
+
+:::
+
+Example:
+```cds
+@Some.Xpr: (a + b)
```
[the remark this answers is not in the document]
stewsk: Yes, but we use the @Some all over the OData docu and have explained that it is only a dummy

cds/cdl.md
```diff
@@ -1003,35 +1003,87 @@ For example, for SAP Fiori models, it's the _4odata_ and _2edm(x)_ processors.
...
+```cds
+using Books from './bookshop-model';
+entity BooksList as select from Books {
+  ID, genre : Genre, title,
+  author.name as author
+};
+```
+
+* `BooksList` would inherit annotations from `Books`
+* `BooksList.ID` would inherit from `Books.ID`
```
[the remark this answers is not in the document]
stewsk: wouldn't it also be `BooksList:ID` instead of `BooksList.ID`  ?

cds/cdl.md
stewsk: Haven't changed that part of the docu, just moved it around ...

cds/cdl.md
stewsk: Not sure, either. I mainly added it because it shows the effect on `=`

advanced/odata.md
```diff
@@ -531,6 +535,192 @@ The second example is for a (record type) term in the [Communication vocabulary]
...
+    status : String;
+  } actions {
+    @Core.OperationAvailable: ( :in.status <> 'A' )
+    action accept (in: $self)
+  }
+}
+```
+
+Most of the OData functions listed in sections
+[5.1.1.5](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360980)
```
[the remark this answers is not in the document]
hjboth: but not for us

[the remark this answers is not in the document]
bugwelle: I thought we're still on OData v4.00?

cds/cdl.md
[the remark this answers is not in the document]
bugwelle: yes. :)

advanced/odata.md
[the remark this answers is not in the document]
bugwelle: ok

advanced/odata.md
```diff
@@ -531,6 +535,192 @@ The second example is for a (record type) term in the [Communication vocabulary]
...
+    @Common.Label: (f.struc.y)
+    val : Integer;
+  }
+  entity F {
+    key id : Integer;
+    struc {
+      y : Integer;
+    }
+  }
+}
```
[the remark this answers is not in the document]
stewsk: In general, yes. But as we recommend to keep the models flat, I'm a bit reluctant to invent a meaningful example with structs at this point in time.
Things are different once we are able to properly deal with foreign keys.
So let's keep this schematic example for now and make a good one when we release the FK handling.

cds/cdl.md
stewsk: One could say that `E` is the primary source for the inner query, and that the subquery is the primary source for the outer query.
To make things really clear, I guess we would have to explicitly mention and explain subqueries. Unless we do that, keep it as it is.

cds/cdl.md
stewsk: Next commit: added a sentence with the new annotation. Collapsed the CSN into "details".
Like that better?

cds/cdl.md
stewsk: Good idea. Comes with next commit

advanced/odata.md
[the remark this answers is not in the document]
bugwelle: Ok. It's just that the `E` and `F` naming isn't used anywhere. No need to have a bookshop-compatible model, but something that is more... meaningful.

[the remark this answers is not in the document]
stewsk: That's the point. I don't want to have something meaningful but only something ... schematic

advanced/odata.md
```diff
@@ -704,16 +710,22 @@ service S {
...
-[5.1.1.5](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360980)
-through [5.1.1.11](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31361018)
-of the [OData specification](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html)
-are also supported.
+In addition, the following functions are supported:
+* `$Null()` for representing the `null` value
+* `DivBy`(or `$DivBy`) and `Mod`(or `$Mod`) for integer division and modulo
+* [`Has`](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_Has) (or `$Has`)
+* the functions listed in sections
+  [5.1.1.5](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360980)
```
bugwelle, line 718: equivalent in v4.0: https://docs.oasis-open.org/odata/odata/v4.0/odata-v4.0-part2-url-conventions.html#_Toc444868693

stewsk: used 4.0.2 spec as in 4.0.1 not all supported functions are listed

advanced/odata.md
```diff
@@ -531,6 +535,218 @@ The second example is for a (record type) term in the [Communication vocabulary]
...
+  } actions {
+    @Core.OperationAvailable: ( :in.status <> 'A' )
+    action accept (in: $self)
+  }
+}
+```
+
+In addition, the following functions are supported:
+* `$Null()` for representing the `null` value
+* `DivBy(...)`(or `$DivBy(...)`) and `Mod(...)`(or `$Mod(...)`) for integer division and modulo
```
hjboth, line 716:
```suggestion
* `Div(...)`(or `$Div(...)`) and `Mod(...)`(or `$Mod(...)`) for integer division and modulo
```

PR #862, opened 2024-04-25
fix: wrong pointer to "Nested Projections"

patricebender:
nested projections != postfix projections

---

actually there is no real documentation available for this feature. At least I didn't find an appropriate place to point at. We should add __some__ documentation on this feature. For now, we should remove the wrong pointer.

### DIFF
--- a/about/index.md
+++ b/about/index.md
@@ -491,7 +491,7 @@ CAP can be combined with event sourcing patterns, that is, by tracking events in
 
 #### CAP supports SQL {#sql}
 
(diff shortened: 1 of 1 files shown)


stewsk:
what's wrong with that link?

patricebender:
Not sure why the job fails with:

´´´log
Found 3 broken link(s) to internal targets in 3 source(s): 
in: /cds/cdl
Unresolved hash link: /cds/cql #nested-expands

in: /java/working-with-cql/query-api
Unresolved hash link: /cds/cql #nested-expands

in: /cds/cdl.html
Unresolved hash link: /cds/cql #nested-expands
```

I searched through the whole code base and there is not a single reference to `#nested-expands`

PR #869, opened 2024-04-26
Document human-readable values for Change tracking in Java

vmikhailenko:
Explains how the change tracking feature (see https://github.com/cap-js/change-tracking) implemented in Java.


java/change-tracking.md
```diff
@@ -158,6 +163,46 @@ not be formatted per user locale or some requirements, for example, different un
...
+
+```cds
+annotate Orders {
+  customer @changelog: [ customer.name ]
+}
+```
+
+Elements from the `@changelog` annotation value must always be prefixed by the association name. The same caveats as for the identifiers for the entities apply here.
+
+In case of the annotated compositions with the identifier, the change log will contain the entry with identifier's value and will include the change log entries for all annotated elements of the composition's target entity.
```
agoerler, line 194:
```suggestion
If you annotate a composition with an identifier, the change log will contain an entry with the identifier's value. Additionally, it will include change log entries for all annotated elements of the composition's target entity.
```

or

```suggestion
If you annotate a composition with an identifier, the change log will contain an entry with the identifier's value. Additionally, it will include change log entries for all annotated elements of the composition's target entities.
```

?

java/change-tracking.md
```diff
@@ -158,6 +163,46 @@ not be formatted per user locale or some requirements, for example, different un
 You should consider this when you decide what to include in the identifier.
 :::
 
+:::danger Sensitive Data
+Do not use sensitive data as the identifiers.
```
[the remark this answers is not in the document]
vmikhailenko: I added more precise description.

java/change-tracking.md
```diff
@@ -164,14 +164,14 @@ You should consider this when you decide what to include in the identifier.
...
 
 :::danger Sensitive Data
-Do not use sensitive data as the identifiers.
+Do not use an element with [personal data](../guides/data-privacy/annotations#personaldata) as the identifier. Such elements will be ignored.
 :::
 
 ### Identifiers for Associated Entities
 
-When your entity has an association to other entity, you might want to log the changes in the relationships between them.
+When your entity has an association to other entity, you might want to log the changes in their relationship.
```
agoerler, line 172:
```suggestion
When your entity has an association to an other entity, you might want to log the changes in their relationship.
```

agoerler (approved):

java/change-tracking.md
```diff
@@ -158,6 +163,46 @@ not be formatted per user locale or some requirements, for example, different un
 You should consider this when you decide what to include in the identifier.
 :::
 
+:::danger Sensitive Data
+Do not use an element with [personal data](../guides/data-privacy/annotations#personaldata) as the identifier. Such elements will be ignored.
+:::
```
renejeglinsky, line 168: The same note is in the previous section.
Is there a reason to repeat it?
If so, why is it a danger note this time?

vmikhailenko: We ignore them anyway. So the previous warning holds true here as well. Removed.

renejeglinsky (approved):

PR #881, opened 2024-04-30
clarify comp / assoc w.r.t. deep ops

David-Kunz:
What's still missing / not clear:
- The target rows of a composition cannot exist without the parent (they're deleted automatically).
- It's disallowed to have multiple compositions pointing to the same rows as they take ownership over them.

Please let me know what you think!

### DIFF
--- a/cds/cdl.md
+++ b/cds/cdl.md
@@ -684,6 +684,7 @@ entity Orders.Items {
 
 :::info Contained-in relationship
 Essentially, Compositions are the same as _[associations](#associations)_, just with the additional information that this association represents a _contained-in_ relationship so the same syntax and rules apply in their base form.
+Deep operations permit modifications to the composition's target entity, unlike associations.
 :::
 
 ::: warning Limitations of Compositions of one


cds/cdl.md
```diff
@@ -684,6 +684,7 @@ entity Orders.Items {
 
 :::info Contained-in relationship
 Essentially, Compositions are the same as _[associations](#associations)_, just with the additional information that this association represents a _contained-in_ relationship so the same syntax and rules apply in their base form.
+Deep operations permit modifications to the composition's target entity, unlike associations.
```
beckermarc, line 687: How does this relate to the docs here: https://pages.github.tools.sap/cap/docs/guides/domain-modeling#compositions?
Seems like this kind of information is already mentioned there. Maybe we should rather link this other section here and extend the other section with additional information that we feel is missing?

renejeglinsky: @David-Kunz ping :)

David-Kunz: @beckermarc , true, there seems to be some kind of overlap, but that also applies to all other features. I think I would leave it as is for now.

renejeglinsky:
@David-Kunz @beckermarc @danjoa How to proceed here?

renejeglinsky:
2nd try: @David-Kunz @beckermarc @danjoa How to proceed here?

PR #884, opened 2024-04-30
messaging: cds.User.privileged for received messages

David-Kunz:


node.js/messaging.md
```diff
@@ -157,6 +157,9 @@ If you want to receive all messages without creating topic subscriptions, you ca
 messaging.on('*', async msg => { /*...*/ })
 ```
 
+::: tip
+In general, messages do not contain user information, hence `cds.context.user` is set to [`cds.User.privileged`](../node.js/authentication#privileged-user).
```
sjvans, line 161:
```suggestion
In general, messages do not contain user information but operate with a technical user (i.e., service instance credentials).
As a consequence, the user of the message processing context (i.e., `cds.context.user`) is set to [`cds.User.privileged`](../node.js/authentication#privileged-user) and, hence, any necessary authorization checks must be done in custom handlers.
```

sjvans: we use the credentials from the service instance binding

renejeglinsky: @David-Kunz ping :)

David-Kunz: @sjvans 
> we use the credentials from the service instance binding

which service instance? I think I would just skip that - it's just the privileged user

renejeglinsky: @sjvans Do you agree?

sjvans: sure, we can remove the `(i.e., service instance credentials)` part

renejeglinsky: Like this?
```suggestion
In general, messages do not contain user information but operate with a technical user. As a consequence, the user of the message processing context (`cds.context.user`) is set to [`cds.User.privileged`](/node.js/authentication#privileged-user) and, hence, any necessary authorization checks must be done in custom handlers.
```

smahati (approved):

PR #889, opened 2024-05-02
add impl() noop wrapper to simple sample

qmacro:
The sample given had a `const cds = require('@sap/cds')` but didn't use the `cds` after that. Moreover, the constant was then highlighted in VS Code as unused, with a rather urgent red squiggly line, which no-one wants :-)

I suggest that rather than omit this line, perhaps it's better to leave it in and use it to provide the noop wrapper that enables Intellisense.

### DIFF
--- a/get-started/in-a-nutshell.md
+++ b/get-started/in-a-nutshell.md
@@ -614,14 +614,14 @@ Copy this into _srv/cat-service.js_ to add custom event handlers:
 ::: code-group
 ```js [srv/cat-service.js]
 const cds = require('@sap/cds')
-module.exports = function (){
+module.exports = cds.service.impl (function (){
   // Register your event handlers in here, for example, ...
   this.after ('each','Books', book => {
     if (book.stock > 111) {
       book.title += ` -- 11% discount!`
     }
   })
-}
+})
 ```
 :::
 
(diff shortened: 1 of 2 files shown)


chgeo:
Thanks for the hint @qmacro.
Looks like a nice way out of the 'unused var' issue.  The thing that bothers me is that our docs [refer to `cds.service.impl` as 'old-style'](https://cap.cloud.sap/docs/node.js/core-services#how-to-provide-custom-service-implementations).  So I wonder if should rather use the `class` pattern (`cds.ApplicationService`) although that's a bit harder to grasp for beginners.
@qmacro @sjvans opinions?

sjvans:
i lean towards removing `cds` rather than showing the old style or the more complex class approach. the import is not really needed as we add `global.cds` anyways. and if your liniting rules don't consider that, then you can add it once you need it in your impl.

get-started/in-a-nutshell.md
```diff
@@ -613,7 +613,6 @@ Copy this into _srv/cat-service.js_ to add custom event handlers:
 
 ::: code-group
 ```js [srv/cat-service.js]
-const cds = require('@sap/cds')
 module.exports = function (){
   // Register your event handlers in here, for example, ...
   this.after ('each','Books', book => {
```
renejeglinsky, line 618: I couldn't add a suggestion, So I just did the change to show how I understood @sjvans.
Would this be the proposal?

PR #896, opened 2024-05-06
[cds^8] Node.js - rm legacy config for draft locks

johannes-vogel:
@renejeglinsky @smahati What do you think of introducing a different label for next major release? Or should I use this one?


renejeglinsky:
We're thinking about having a convention like I added it to the title now. That maybe makes more sense than lables. But that's not discussed yet, just for the time being. And I think it does the trick :)

renejeglinsky (approved):

PR #912, opened 2024-05-07
Improve `cds add` documentation

swaldmann:
- Also documented process for `cds add helm`
- Experimental: added styling for added files in code group:
Screenshot 2024-05-07 at 14 03 15


chgeo (approved):

renejeglinsky:
If we want to use that styling globally, we'd need to think about notebooks as well.

renejeglinsky (approved):

PR #940, opened 2024-05-15
Added section to describe how parts of the UI can be hidden based on user role

schiwekM:
Hi @renejeglinsky,

a few newcomers mentioned the lack of documentation on how parts of the UI can be hidden based on roles. This section should explain it.

@DanSchlachter Maybe it is also something to include in a tutorial or course / SFlight?

Hope the wording is understandable.

BR,
Marten


DanSchlachter:
maybe it should be mentioned, that there is a trade-off between having multiple apps for multiple roles or a single app for multiple roles. 
You don't want to duplicate the entire app for a small difference in UI -> use small amount of additional logic over creating multiple apps with duplicate code
You don't want to have a lot of logic in one app, when it's easier to define two apps -> use two apps over overcomplicating one app

schiwekM:
I agree but would argue that this should be part of Providing services or Modelling best practices. In general it is quite common that there is one app but create / edit / delete is restricted to an admin user, I'd say.

advanced/fiori.md
```diff
@@ -485,6 +485,47 @@ SELECT.from(Books.drafts) //returns all drafts of the Books entity
...
+});
+```
+
+and thirdly refer to the singleton in the annotation by using a [dynamic expression](../advanced/odata#dynamic-expressions)
+
+```cds
+annotate service.Books with @(
+    UI.CreateHidden : { $edmJson: {$Not: { $Path: '/CatalogService.EntityContainer/Configuration/isAdmin'} } },
+    UI.UpdateHidden : { $edmJson: {$Not: { $Path: '/CatalogService.EntityContainer/Configuration/isAdmin'} } },
+);
```
renejeglinsky, line 518: Hi @schiwekM ,
what is the result here? When I'm logged in as admin, I can't see the Books through this service?

advanced/fiori.md
```diff
@@ -485,6 +485,47 @@ SELECT.from(Books.drafts) //returns all drafts of the Books entity
...
+```
+
+The Entity Container is OData specific and refers to the '$metadata' of the OData service in which all accessible entities are located within the Entity Container. SAP Fiori elements also allows to not include it in the path:
+
+```cds
+annotate service.Books with @(
+    UI.CreateHidden : { $edmJson: {$Not: { $Path: '/Configuration/isAdmin'} } },
+    UI.UpdateHidden : { $edmJson: {$Not: { $Path: '/Configuration/isAdmin'} } },
+);
+```
```
renejeglinsky, line 528: This is doing the same as the previous snippet when using Fiori Elements, right? In which cases am I not allowed to do omit the entity container?

advanced/fiori.md
schiwekM: essentially if you are not the admin than you do not see the EDIT or Create buttons

advanced/fiori.md
schiwekM: Yes - EntityContainer is syntactically from an OData PoV correct however in most cases Fiori elements allows you to leave it out. For compatibility I added the entity container notation if something does not work that users can try the syntactically correct one.

renejeglinsky (approved):
Hi Marten,
thank you!
Please check my edits and merge if you agree. I'll take it as a follow-up to link it from other places and have an introduction to the topic and when to use what.

schiwekM:
Hi Rene,

LGTM - if you need any input from my side let me know. Moreover as an additional follow up might be expressions for annotation on which the compiler team is currently working on.

BR, Marten

PR #943, opened 2024-05-16
doc(features): sort alphabetically

joergmann:


chgeo (commented):
I wonder if this is helpful or makes things harder to find.
Atm the facets are grouped logically (kind of, see the DB block at the start, the messaging and MTX stuff together).

Shouldn't we go for a 'type' column or so, and then group it along this type?

@swaldmann @renejeglinsky what do you think?

swaldmann:
Yes, I also chose to group by type in `cds help add`, see cap/cds-dk/pull/2663#issuecomment-6430931.

Let's discuss this in the cds-dk sync and align accordingly.

PR #950, opened 2024-05-21
doc(CSN): Formatting; follow conventions for i18n

bugwelle:
This commit only includes some formatting changes, and adds a link to the CSN's `requires` equivalent in CDL.


stewsk (approved):

renejeglinsky (approved):
Thanks!

PR #966, opened 2024-05-29
[cds^8] No strings in INSERT.into and UPSERT.into

hm23:
With cds 8, we only want to allow refs, no strings in `INSERT.into` and `UPSERT.into` (as it is already the case for UPDATE and DELETE).


sjvans (approved):
lgtm

johannes-vogel (approved):

renejeglinsky (approved):

PR #973, opened 2024-06-03
fix: Move DB agnostic sections out of `sqlite` database guide

patricebender:
With this change, `Common Operators` and `Common Functions` have been moved from the `sqlite` guide to the generic database guide.

The function list is now including all functions which are supported by all `cap-js/dbs`.

Todos:

- [ ] align with Java colleagues, I think we can make this section a shared one for both stacks
- [ ] how to document the to-be-expected return types?

Also, it would be nice to remove the disclaimer `only supported in runtime queries`, but for that we first need compiler support for the functions as well as for the operators.


guides/databases.md
```diff
@@ -327,6 +327,85 @@ Select.from(AUTHOR)
...
+With special mappings for the following boolean operators:
+
+| Operator  | `@cap-js/sqlite` | `@cap-js/hana` | `@cap-js/postgres` |
+|-----------|------------------|----------------|---------------------|
+| `= NULL`  | `is NULL`        | `is NULL`      | `is NULL`           |
+| `!=`      | `<>`             | `!=`           | `<>`                |
+| `!= NULL` | `is not NULL`    | `is not NULL`  | `is not NULL`       |
+
+
+In particular, the translation of `!=` to `IS NOT` in SQLite — or to an equivalent polyfill in SAP HANA — greatly improves the portability of your code.
```
patricebender, line 345: what is the "equivalent polyfill"? Actually, `!=` is just passed to hana as-is.

guides/databases.md
```diff
@@ -327,6 +327,85 @@ Select.from(AUTHOR)
...
+
+> These operators are available for runtime queries, but not in CDS files.
```
patricebender, line 347: is that true?

guides/databases.md
```diff
@@ -327,6 +327,85 @@ Select.from(AUTHOR)
...
+* `startswith(x,y)` — checks whether `y` starts with `x`
+* `endswith(x,y)` — checks whether `y` ends with `x`
+* `matchesPattern(x,y)` — checks whether `x` matches regex `y`
+* `substring(x,i,n)` — extracts a substring from `x` starting at `i` with length `n` <sup>1</sup>
+* `indexof(x,y)` — returns the (zero-based) index of the first occurrence of `y` in `x`
+* `length(x)` — returns the length of string `x`
+* `tolower(x)` — returns all-lowercased `x`
+* `toupper(x)` — returns all-uppercased `x`
+* `ceiling(x)` — returns ceiled `x`
+* `session_context(v)` — with standard variable names → [see below](#session-variables)
```
renejeglinsky, line 366: Link is broken b/c the content has moved from one guide to another.
```suggestion
* `session_context(v)` — with standard variable names → [see below](/guides/databases-sqlite#session-variables)
```

patricebender:
ping @johannes-vogel @stewsk

johannes-vogel:
does that also apply to java?

guides/databases.md
BobdenOs: There is a polyfill for HANA [here](https://github.com/cap-js/cds-dbs/blob/c3c45630812539ff22c3fd7696e2c90ac973675e/hana/lib/HANAService.js#L786-L851)

guides/databases.md
BobdenOs: Are you sure this is using `@cap-js/hana` ? As it doesn't have `JSON` functions.

patricebender: No, it seems like my profile has not been setup correctly. What do you think would make sense to put into the hana column here? The query in the example actually leads to the "polyfill":
```sql
WHERE (
                                CASE
                                    WHEN Books.ID != ? THEN true
                                    WHEN NOT Books.ID != ? THEN false
                                    ELSE (
                                        CASE
                                            WHEN COALESCE(Books.ID, ?) IS NULL THEN false
                                            ELSE true
                                        END
                                    )
                                END
                            ) = TRUE
```

this does not really fit into the table :D

renejeglinsky:
> does that also apply to java?

@johannes-vogel Do you mean we'd need the equivalent docs for Java when moving those docs in the common guide?

@patricebender Is there anything I can support you with?

patricebender:
> > does that also apply to java?
> 
> @johannes-vogel Do you mean we'd need the equivalent docs for Java when moving those docs in the common guide?
> 
> @patricebender Is there anything I can support you with?

I think the "common operators" section must be moved to the generic part of the db guide, I would like to hear the POs opinions here. From my side, this change is fine (I only need to apply the suggestion from Bob).

johannes-vogel:
> > > does that also apply to java?
> > 
> > 
> > @johannes-vogel Do you mean we'd need the equivalent docs for Java when moving those docs in the common guide?
> > @patricebender Is there anything I can support you with?
> 
> I think the "common operators" section must be moved to the generic part of the db guide, I would like to hear the POs opinions here. From my side, this change is fine (I only need to apply the suggestion from Bob).

I was simply wondering whether this is node specific or also applicable for java. at the moment it's tagged as node specific. in general, fine for me:)

guides/databases.md
```diff
@@ -327,7 +327,80 @@ Select.from(AUTHOR)
...
+* `tolower(x)` — returns all-lowercased `x`
+* `toupper(x)` — returns all-uppercased `x`
+* `ceiling(x)` —  rounds the input numeric parameter up to the nearest numeric value
+* `floor(x)` — rounds the input numeric parameter down to the nearest numeric value
+* `round(x)` — rounds the input numeric parameter to the nearest numeric value.
+               The mid-point between two integers is rounded away from zero, i.e. 0.5 is rounded to 1 and ‑0.5 is rounded to -1.
+* `year(x)` `month(x)`, `day(x)`, `hour(x)`, `minute(x)`, `second(x)`, `fractionalseconds(x)`, `time(x)`, `date(x)` —
+  returns parts of a datetime for a given `cds.DateTime` / `cds.Date` / `cds.Time`
+* `maxdatetime(x)`, `mindatetime(x)` — return the maximum or minimum datetime for a given `cds.DateTime` / `cds.Date` / `cds.Time`
+* `totalseconds(x)` — returns the total seconds of a datetime a given `cds.DateTime` / `cds.Time`
```
patricebender, line 368:
```suggestion
* `totalseconds(x)` — returns the total seconds of a datetime for a given `cds.DateTime` / `cds.Time`
```

renejeglinsky:
@MattSchur @agoerler Could you please check this PR? It's about moving a topic and also what is probably Node/Java specific (or not!)

guides/databases.md
```diff
@@ -325,7 +325,80 @@ Select.from(AUTHOR)
...
+* `!=`, `<>` — with `!=` translated to `IS NOT` in SQLite, or to `IS DISTINCT FROM` in standard SQL, or to an equivalent polyfill in SAP HANA
+* `<`, `>`, `<=`, `>=`, `IN`, `LIKE` — are supported as is in standard SQL
+* `||` — concatenation operator
+
+In particular, the translation of `!=` to `IS NOT` in SQLite — or to `IS DISTINCT FROM` in standard SQL, or to an equivalent polyfill in SAP HANA — greatly improves the portability of your code.
+
+> These operators are available for runtime queries, but not in CDS files.
+
+
+### Functions Mappings for Runtime Queries {.impl .node}
```
agoerler, line 342: We have a general agreement on the semantic of the functions: https://github.tools.sap/cap/dev/issues/1111

We could create a table here to capture which functions are supported by CAP Java and which by CAP Node.js. SImilar to what @patricebender compiled here: https://github.tools.sap/cap/dev/issues/1111#issuecomment-7092917

guides/databases.md
```diff
@@ -325,7 +325,80 @@ Select.from(AUTHOR)
...
+* `seconds_between` — Computes the number of seconds between two specified dates.
+* `nano100_between` — Computes the time difference between two dates to the precision of 0.1 microseconds.
+
+The database service implementation translates these to the best-possible native SQL functions, thus enhancing the extent of **portable** queries.
+With open source and the new database service architecture, we also have methods in place to enhance this list by custom implementation.
+
+> For the SAP HANA functions, both usages are allowed: all-lowercase as given above, as well as all-uppercase.
+
+
+### Session Variables {.impl .node}
```
agoerler, line 389: There is a completely different API in CAP Java.

guides/databases.md
patricebender: @renejeglinsky As a first step, this list contains all functions which are supported by the node.js runtime. For months this list has been outdated **and** it was not at the right place of the documentation (those functions are supported by all db services, not only sqlite).

@agoerler what do you think about adding the functions which are supported by the java runtime (`{.impl .java}`) in a similar fashion? As we already went over the list, I could also contribute this list (assuming the semantics of the functions are the same).


I would probably refrain from using a table in the documentation. The list of supported functions could be toggled instead, for better readability.

renejeglinsky: I think adding the functions using the toggle would be great. Thanks Patrice!

guides/databases.md
```diff
@@ -325,7 +325,80 @@ Select.from(AUTHOR)
...
 </div>
 
+### Standard Operators {.impl .node}
 
+The database services guarantee identical behavior of these operators:
+
+* `==`, `=` — with `=` null being translated to `is null`
+* `!=`, `<>` — with `!=` translated to `IS NOT` in SQLite, or to `IS DISTINCT FROM` in standard SQL, or to an equivalent polyfill in SAP HANA
+* `<`, `>`, `<=`, `>=`, `IN`, `LIKE` — are supported as is in standard SQL
+* `||` — concatenation operator
```
patricebender, line 335: dbs might have different execution order wrt the concat operator → advertise only `concat` function for string concatenation

guides/databases.md
renejeglinsky: As this is toggled for Node we could provide the same section for Java. Would be really nice to have it in this PR but if capacity doesn't allow for that, I don't think that would be a blocker for this change, or?

guides/databases.md
renejeglinsky: @patricebender As this section is toggled for node, should we include the table for Java in this PR or do this as a follow-up?

patricebender: I'd be fine with following up in another change to add a java toggle here 👍

guides/databases.md
```diff
@@ -325,7 +325,85 @@ Select.from(AUTHOR)
...
+In particular, the translation of `!=` to `IS NOT` in SQLite — or to `IS DISTINCT FROM` in standard SQL, or to an equivalent polyfill in SAP HANA — greatly improves the portability of your code.
+
+::: warning Runtime Only
+The operator mappings are available for runtime queries only, but not in CDS files.
+:::
+
+
+### Functions Mappings for Runtime Queries {.impl .node}
+
+A specified set of standard functions is supported in a **database-agnostic**, hence portable way, and translated to database-specific variants or polyfills.
```
agoerler, line 345: Another thing to consider. These standard functions are not only _database_-agnostic but even _data-source_-agnostic. They can also be used with remote OData. Moreover, they there names, semantics and parameters are closer to OData then to SQL. Maybe we should move this section out of _database.md_ and rather to a CDS QL section.

renejeglinsky: @johannes-vogel What do you think?

patricebender:
@stewsk @johannes-vogel could you kindly review

danjoa (approved):

stewsk (approved):

agoerler:
> @agoerler what do you think about adding the functions which are supported by the java runtime ({.impl .java}) in a similar fashion?

For CAP Java, the set of supported functions is documented here:

* scalar functions: https://cap.cloud.sap/docs/java/working-with-cql/query-api#scalar-functions
* predicated: https://cap.cloud.sap/docs/java/working-with-cql/query-api#predicate-functions

As [mentioned earlier](#discussion_r1804303937), these functions can be used on the DBs but also in remote OData consumption. We could add a link in an ({.impl .java}) variant.

renejeglinsky (approved):

patricebender:
@renejeglinsky how can I fix the broken anchor links mentioned in the build:

```
in: /releases/jun24
Unresolved hash link: /guides/databases-sqlite #standard-operators
Unresolved hash link: /guides/databases-sqlite #standard-functions
Unresolved hash link: /guides/databases-sqlite #sap-hana-functions
```

I can't find them in the repository :(

PR #975, opened 2024-06-05
fix(databases): Really fix code comment in SQL snippet

bugwelle:
Follow up to #972.

### DIFF
--- a/guides/databases.md
+++ b/guides/databases.md
@@ -756,7 +756,7 @@ CREATE TABLE Books (
   author_ID INTEGER,    -- generated foreign key field
   ...,
   PRIMARY KEY(ID),
-  CONSTRAINT Books_author -- [!code focus]
+  CONSTRAINT Books_author -- // [!code focus]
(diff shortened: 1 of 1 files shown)


swaldmann:
This will render the `--` though, which is quite weird. So I'd either add an actual comment or use the line highlighting instead.

guides/databases.md
```diff
@@ -756,7 +756,7 @@ CREATE TABLE Books (
   author_ID INTEGER,    -- generated foreign key field
   ...,
   PRIMARY KEY(ID),
-  CONSTRAINT Books_author -- [!code focus]
+  CONSTRAINT Books_author -- // [!code focus]
```
bugwelle, line 759:
```suggestion
  CONSTRAINT Books_author   -- constraint is explicitly named // [!code focus]
```

swaldmann (approved):

renejeglinsky (approved):

PR #977, opened 2024-06-05
Use `schema.cds` naming throughout

swaldmann:
As reported by DJ, we were inconsistent in the way we name our schema file
- `cds add tiny-sample` created a `db/data-model.cds`
- `cds add sample` created a `db/schema.cds`

We settled for `schema.cds` for both in cds8. This should also be reflected in capire.


chgeo (approved):

renejeglinsky:
This is going to be published today. I hope that doesn't harm

swaldmann:
Nope, should be fine

PR #983, opened 2024-06-07
[CAPJAVA^3] Fix the wrong precondition in the list

vmikhailenko:
Fix a result of failed merge in change tracking.


agoerler (approved):
(we can remove this with CAP Java 3 as it's the default then)

vmikhailenko:
@renejeglinsky This is not relevant to CAP Java 3.0. This fixes precondition mentioned twice here  https://cap.cloud.sap/docs/java/change-tracking#enabling-change-tracking

renejeglinsky (approved):
Thanks!

PR #984, opened 2024-06-07
Mention that draft decorations can't be used in annotation expressions

hjboth:
Sth like `(IsActiveEntity = false)` doesn't work.

### DIFF
--- a/cds/cdl.md
+++ b/cds/cdl.md
@@ -1094,7 +1094,7 @@ Some advantages of using expressions as "first class" annotation values are:
 * [automatic translation of expressions in OData annotations](#odata-annotations)
 
 ::: info Limitations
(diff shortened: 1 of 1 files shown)


cds/cdl.md
```diff
@@ -1093,6 +1093,10 @@ Some advantages of using expressions as "first class" annotation values are:
 * [automatic path rewriting in propagated annotations](#propagation)
 * [automatic translation of expressions in OData annotations](#odata-annotations)
 
+::: info Limitations
+Elements that are not available to the compiler such as the OData draft decoration for example can't be used in annotation expressions.
```
stewsk, line 1097:
```suggestion
Elements that are not available to the compiler, for example the OData draft decoration, can't be used in annotation expressions.
```

stewsk (approved):

renejeglinsky (approved):

PR #986, opened 2024-06-10
replace `npm add` with `npm install -D` for typescript setup

ziyang-works:
This PR should close https://github.com/cap-js/docs/issues/985


daogrady (approved):
Makes sense, thanks for the contribution!

PR #1005, opened 2024-06-13
Remove CdsDataStore documentation from PersistenceService chapter

beckermarc:
PersistenceService is preferred over CdsDataStore. Also configuring custom CdsDataStoreConnector instances is not recommened anymore since we support additional PersistenceServices out-of-the-box meanwhile as well


agoerler (approved):
I think there are use cases when you want to use this API. But these cases are when cds4j is used as a library outside of CAP. Hence it's OK to remove the documentation here.

renejeglinsky (approved):

PR #1007, opened 2024-06-13
docs(Java): Adapt CQL counterparts: fix syntax

bugwelle:


java/working-with-cql/query-execution.md
```diff
@@ -428,8 +428,8 @@ Select BooksWithLowStock where author = 'Kafka'
 is executed against SQL databases as
 
 ```SQL
```
rjayasinghe, line 430: BTW: can't we use ````cds` here?

bugwelle: Because this is (not) just CDS, but SQL. Both would fit, but I think SQL will yield better syntax highlighting. :)

rjayasinghe: Ok 👍🏾

renejeglinsky (approved):

PR #1013, opened 2024-06-14
[3.x] cds-services-archetype improvements

mofterdinger:
This PR is related to CAP Java 3.x.


java/developing-applications/building.md
```diff
@@ -330,10 +330,11 @@ It supports the following command-line options:
 
 | Option | Description |
 | -- | -- |
-| `-DincludeModel=true` | Adds a minimalistic sample CDS model to the project |
-| `-DincludeIntegrationTest=true` | Adds an integration test module to the project |
-| `-DodataVersion=[v2\|v4]` | Specify which protocol adapter is activated by default |
-| `-DtargetPlatform=cloudfoundry` | Adds CloudFoundry target platform support to the project |
+| `-Dtype=[default|noPersistence]` | Choose the project type to generate. Defaults to `default`. |
```
renejeglinsky, line 333: "Defaults to `default`" sounds like stating the obvious. 
The value `default` is not really self.explanatory, isn't it? Only in comparison with ` noPersistence` you get an idea what default could be about.
Is it possible to give a more self-explanatory value?

mofterdinger: mhh, good question, by default, the `cds-services-archetype` creates a CAP Java **default** application. Should we name it `application` instead of `default` ? If we change it here in the docu, we need to change it also in the `cds-services-archetype`.

@BraunMatthias @rjayasinghe Do you have any suggestion for the name of the type ?

BraunMatthias (approved):

java/developing-applications/building.md
renejeglinsky: `withPersitence` and `noPersistence` might also be too close. Is it sth like `fullStack` vs. `noPersistence`?

mofterdinger: I have schedule a meeting monday to discuss the parameter name and values with the team. Will update this PR once we have clarified the names. Do not merge yet.

mofterdinger: Identified project types / flavor:
- application
  - extension (like application, but without persistence)
- plugin

I will change the parameter name to `persistence` with possible values: `true` / `false`.

mofterdinger: renamed in cds-services-archetype.

renejeglinsky (approved):
🤝

PR #1016, opened 2024-06-17
Destination caching is enabled by default

mariayord:


guides/using-services.md
```diff
@@ -1014,7 +1014,8 @@ Additionally, you can provide [destination options](https://sap.github.io/cloud-
 
 The `selectionStrategy` property controls how a [destination is resolved](#destination-resolution).
 
-The `useCache` option controls whether the SAP Cloud SDK caches the destination. Read [Destination Cache](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destination-cache#destination-cache) to learn more about how the cache works.
+The `useCache` option controls whether the SAP Cloud SDK caches the destination. In Node.js the destination caching is enabled by default, `useCache` is set to `true`.
```
renejeglinsky, line 1017:
```suggestion
The `useCache` option controls whether the SAP Cloud SDK caches the destination. In Node.js, the destination caching is enabled by default and `useCache` is set to `true`.
```

guides/using-services.md
```diff
@@ -1014,7 +1014,8 @@ Additionally, you can provide [destination options](https://sap.github.io/cloud-
 
 The `selectionStrategy` property controls how a [destination is resolved](#destination-resolution).
 
-The `useCache` option controls whether the SAP Cloud SDK caches the destination. Read [Destination Cache](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destination-cache#destination-cache) to learn more about how the cache works.
+The `useCache` option controls whether the SAP Cloud SDK caches the destination. In Node.js, the destination caching is enabled by default and `useCache` is set to `true`.
```
johannes-vogel, line 1017:
```suggestion
The `useCache` option controls whether the SAP Cloud SDK caches the destination. It's enabled by default but can be disabled by expliticly setting it to `false`.
```

We should also adapt the snippet above and set it to false.

mariayord: @johannes-vogel  I think it is good to point that in Node.js it is enabled by default , because in the [link]( https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destination-cache#destination-cache)(after that), the first sentence is that it is disabled by default and that can cause confusion.

mariayord: And why should we change the snippet ? As it is enabled by default? If the purpose of this snippet is to show how to disable it , I am OK, but this is not the case.

renejeglinsky: I agree with the proposed change to the sentence by Johannes.
For the snippet: Is the example for `selectionStrategy` also showing the default?

johannes-vogel: the default for selectionStrategy is `subscriberFirst`

mariayord: when Johannes says....

guides/using-services.md
```diff
@@ -1012,7 +1012,8 @@ Additionally, you can provide [destination options](https://sap.github.io/cloud-
 
 The `selectionStrategy` property controls how a [destination is resolved](#destination-resolution).
 
-The `useCache` option controls whether the SAP Cloud SDK caches the destination. Read [Destination Cache](https://sap.github.io/cloud-sdk/docs/js/features/connectivity/destination-cache#destination-cache) to learn more about how the cache works.
+The `useCache` option controls whether the SAP Cloud SDK caches the destination. It's enabled by default but can be disabled by expliticly setting it to `false`.
```
renejeglinsky, line 1015:
```suggestion
The `useCache` option controls whether the SAP Cloud SDK caches the destination. It's enabled by default but can be disabled by explicitly setting it to `false`.
```

johannes-vogel (approved):

renejeglinsky (approved):

PR #1017, opened 2024-06-17
hard coded vs usr defined strlength

hjboth:
... make foot note clearer

### DIFF
--- a/cds/types.md
+++ b/cds/types.md
@@ -47,7 +47,7 @@ The following built-in types are provided:
 
 > <sup>(4)</sup> Up to 7 digits of fractional seconds; if a data is given with higher precision truncation may occur
 
-> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. If no option is specified, a hard-coded default length of 5000 for SAP HANA and 255 for all other SQL backends applies. Note that this hard-coded default lengths are only applied on database level. Specify fixed lengths to get length checks on service level and/or inbound data.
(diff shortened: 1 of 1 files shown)


stewsk (approved):

cds/types.md
```diff
@@ -47,7 +47,7 @@ The following built-in types are provided:
 
 > <sup>(4)</sup> Up to 7 digits of fractional seconds; if a data is given with higher precision truncation may occur
 
-> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. The default length is 5000 for SAP HANA and 255 for all other SQL backends. Note that default lengths are only applied on database level. Specify fixed lengths to get length checks on service level and/or inbound data.
+> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. If no option is specified, a hard-coded default length of 5000 for SAP HANA and 255 for all other SQL backends applies. Note that this hard-coded default lengths are only applied on database level. Specify fixed lengths to get length checks on service level and/or inbound data.
```
bugwelle, line 50:
```suggestion
> <sup>(5)</sup> Argument `length` is optional → use options `cds.cdsc.defaultStringLength` and `cds.cdsc.defaultBinaryLength` to control the project-specific default length used for OData and SQL backends. If no option is specified, a hard-coded default length of 5000 for SAP HANA and 255 for all other SQL backends applies. Note that these hard-coded default lengths are only applied on database level. Specify fixed lengths to get length checks on service level and/or inbound data.
```

renejeglinsky:
We had a similar formulation [previously](https://github.com/cap-js/docs/commit/cb3e102d319644336845fda825fc9860d84979c0). But I don't remember exactly what the issues with "If not set" was. Are there restrictions wrt to length so that not every specified option is a valid option?
Might be, that back then the confusion was db layer vs. service layer. Does anyone remember? Most probably it's not an issue and nothing speaks against the way Hans-Joachim formulated it, just wanted to throw into the pot what I faintly remember :)

hjboth:
I can't remember...

renejeglinsky (approved):

PR #1019, opened 2024-06-18
docs(java outbox): Improve description for `.cdsrc.json`

bugwelle:
While the text mentions `.cdsrc.json`, it could just as well show the file using the code-group feature.

This commit does exactly that.

### DIFF
--- a/java/outbox.md
+++ b/java/outbox.md
@@ -34,10 +34,9 @@ Once the transaction succeeds, the messages are read from the database table and
 
 To enable the persistence for the outbox, you need to add the service `outbox` of kind `persistent-outbox` to the `cds.requires` section in the _package.json_ or _cdsrc.json_:
 
-::: code-group
-```jsonc [package.json]
+```jsonc 
 {
-  // […]
+  // ...
   "cds": {
     "requires": {
       "outbox": {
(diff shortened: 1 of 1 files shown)


java/outbox.md
```diff
@@ -34,22 +34,35 @@ Once the transaction succeeds, the messages are read from the database table and
 
 To enable the persistence for the outbox, you need to add the service `outbox` of kind `persistent-outbox` to the `cds.requires` section in the _package.json_ or _cdsrc.json_. Please note that the _cdsrc.json_ file represents already the `cds` section and only the `requires` section should be added to the _cdsrc.json_ file:
```
renejeglinsky, line 35: I just recently learned that the _.cdsrc.json_ can also hold a cds section. So we don't need to differentiate and keep it simpler.
```suggestion
To enable the persistence for the outbox, you need to add the service `outbox` of kind `persistent-outbox` to the `cds.requires` section in the _package.json_ or _cdsrc.json_:
```

bugwelle: It can? Great!

renejeglinsky: @chgeo My statement that _.cdsrc.json_ can hold a `cds` section is correct, or?

chgeo: Yes, nowadays you also can have such a section in `.cdsrc.json` while traditionally you couldn't.

renejeglinsky: Thanks! Then I'll remove the code-groups here as we don't need them anymore.

renejeglinsky (approved):

PR #1027, opened 2024-06-18
docs(java outbox): Show hard-coded value `ordered` for default outboxes

bugwelle:
Users can't change `ordered` of `DefaultOutboxOrdered` and `DefaultOutboxUnordered`.  But the documentation is not _that_ clear about that the first time reading this document.

This commit adds the default and non-changable values for `ordered` for these two default outboxes and mentions that they can't be changed in the option description.


t-bonk (approved):
LGTM

renejeglinsky (approved):

PR #1036, opened 2024-06-19
feat: add disclaimer to change tracking plugin

patricebender:
this was already done in https://cap.cloud.sap/docs/java/change-tracking

but there is no disclaimer for this in the node world (besides a warning in the readme of the plugin) 

I think it makes sense to address this concern prominently.

### DIFF
--- a/plugins/index.md
+++ b/plugins/index.md
@@ -217,6 +217,10 @@ Learn more about audit logging in [Node.js](../guides/data-privacy/audit-logging
 
 ## Change Tracking
 
+:::warning Personal data is ignored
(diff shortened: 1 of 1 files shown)


renejeglinsky:
Well, I think the disclaimer is already where it should be. Adding it here is a redundancy that we should only accept in certain situations. Therefore let me enhance the reviewer list before closing or merging.

nkaputnik:
I would go with @renejeglinsky in this matter. I think the disclaimer in the Plugin-Readme is good enough. 
I tend to see this capire section as a marketing/entry page which should just tease what the plugins can do, not much more...

patricebender:
Alright fine by me :)

PR #1049, opened 2024-06-24
Update learning-sources.md

markushaug:
Adding official SAP Learning Journeys on learning.sap.com.


DanSchlachter (approved):
looks good, I think putting them on top makes sense as they are full learning journeys and being kept up to date.

PR #1050, opened 2024-06-24
fix typo and rearrange to avoid multiple plural-possessive-quotes in …

qmacro:
…core-services.md

Spotted "servce" [sic] then thought I'd rearrange the phrase anyway to help avoid two plural-possessive-quotes in one phrase :-)


renejeglinsky (approved):
👍

PR #1053, opened 2024-06-25
fix: it's not propagated from msg.id

David-Kunz:


renejeglinsky:
Should there be a replacement for this sentence?

David-Kunz:
> Should there be a replacement for this sentence?

No I don't think so, it was simply wrong.

renejeglinsky (approved):

PR #1060, opened 2024-06-26
docs(Java): Limit project layout to stable-files; remove volatile stuff

bugwelle:
The issue with the project layout is: it was outdated, see https://github.com/cap-js/docs/pull/978

However, I forgot to adapt the table!

IMHO, let's remove all files that are volatile, i.e. that may change in the future. That includes example files (i.e. their filenames). That only stable stuff are the Java project folders. Let's keep those.


bugwelle:
Oh! Duplicate of https://github.com/cap-js/docs/pull/1009

PR #1069, opened 2024-06-27
troubleshooting: mention logging for cloud sdk

David-Kunz:


hm23 (approved):

renejeglinsky (approved):

renejeglinsky:
Is there a typical error code or error message that we could add to improve findability?

David-Kunz:
> Is there a typical error code or error message that we could add to improve findability?

The outer error is probably `502 Bad Gateway` (unless the customer handler changes it), but I think I wouldn't mention it because the 'real' error is the one from the remote system, that's not generally known

PR #1073, opened 2024-06-28
docs: Remove usage of Node.js/Java code groups; use toggles instead

bugwelle:
The Node/Java toggles do not toggle the code snippets.  That's annoying, as I now have to manually select the proper/correct snippets each time.

A task, which users have failed in the past, which then lead to issues.

Now, let's make the snippets part of the implementation.


rjayasinghe:
I can remember that we already had it like you propose here. @renejeglinsky maybe we can have a quick sync on that. The problem is that the java/nodejs toggle for the whole site does not toggle the code samples at the moment.

bugwelle:
@rjayasinghe / @renejeglinsky What do you think?

renejeglinsky:
This has been part of discussions for several times. I need to align with Daniel on that and get back to you.

PR #1089, opened 2024-07-08
Avoid `cap>=ire` in the window title

chgeo:
`cap>=ire` is hard as a search term in browser history, so use a simpler text in the window title.


chgeo:
@qmacro FYI

swaldmann (approved):

renejeglinsky (approved):

PR #1093, opened 2024-07-10
Update many to many relation example

schiwekM:
Hi @renejeglinsky,

while helping colleagues, I noticed that the many to many example is chosen poorly. The current example does not consider that this won't work with Fiori elements and I adjusted the sample + added a description for Fiori.

BR, Marten

### DIFF
--- a/guides/domain-modeling.md
+++ b/guides/domain-modeling.md
@@ -529,7 +529,7 @@ entity Users { ...
 ```
 
 Behind the scenes the equivalent of the model above would be generated, with the link table called `Projects.members` and the backlink association to `Projects` in there called `up_`.
-Consider that for SAP Fiori elements 'project' and 'user' shall not be keys, even if their combination is unique, because as keys those fields can't be edited on the UI. In this case a different key (e.g. a UUID) is required, and the unique constraint for `project` and `user` can be expressed via `@assert.unique`.
+Consider that for SAP Fiori elements 'project' and 'user' shall not be keys, even if their combination is unique, because as keys those fields can't be edited on the UI. In this case a different key is required, for example a UUID, and the unique constraint for `project` and `user` can be expressed via `@assert.unique`.
 
 ### Compositions
 
--- a/guides/domain-modeling.md
+++ b/guides/domain-modeling.md
@@ -511,9 +511,9 @@ entity Projects { ...
 entity Users { ...
   projects : Composition of many Members on projects.user = $self;
 }
-entity Members { // link table
-  key project : Association to Projects;
-  key user : Association to Users;
+entity Members: cuid { // link table
+  project : Association to Projects;
+  user : Association to Users;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


guides/domain-modeling.md
```diff
@@ -529,6 +529,7 @@ entity Users { ...
 ```
 
 Behind the scenes the equivalent of the model above would be generated, with the link table called `Projects.members` and the backlink association to `Projects` in there called `up_`.
+Please consider that for SAP Fiori elements you should not use [_Compositions of Aspects_](#composition-of-aspects) because necessary annotations are hard to add. Moreover 'project' and 'user' fields shall not be keys, even if any combination shall be unique, because as keys those fields could not be edited on the UI, thus requiring a different key and the unique constraint being expressed via `@assert.unique`.
```
stewsk, line 532: Hi @schiwekM , can you please explain what you mean by "necessary annotations are hard to add". Can you provide an example?

schiwekM: I checked again and with `annotate Orders.Items with @()`. I think in the past this was not possible, if I remember correctly (or at least Code completion did brought it up) and probably I missed the update. The line is now removed

guides/domain-modeling.md
```diff
@@ -529,6 +529,7 @@ entity Users { ...
 ```
 
 Behind the scenes the equivalent of the model above would be generated, with the link table called `Projects.members` and the backlink association to `Projects` in there called `up_`.
+Please consider that for SAP Fiori elements 'project' and 'user' fields shall not be keys, even if any combination shall be unique, because as keys those fields could not be edited on the UI, thus requiring a different key and the unique constraint being expressed via `@assert.unique`.
```
stewsk, line 532:
```suggestion
Consider that for SAP Fiori elements 'project' and 'user' shall not be keys, even if their combination is unique, because as keys those fields can't be edited on the UI. In this case a different key (e.g. a UUID) is required, and the unique constraint for `project` and `user` can be expressed via `@assert.unique`.
```

guides/domain-modeling.md
```diff
@@ -529,6 +529,7 @@ entity Users { ...
 ```
 
 Behind the scenes the equivalent of the model above would be generated, with the link table called `Projects.members` and the backlink association to `Projects` in there called `up_`.
+Consider that for SAP Fiori elements 'project' and 'user' shall not be keys, even if their combination is unique, because as keys those fields can't be edited on the UI. In this case a different key (e.g. a UUID) is required, and the unique constraint for `project` and `user` can be expressed via `@assert.unique`.
```
renejeglinsky, line 532:
```suggestion
Consider that for SAP Fiori elements 'project' and 'user' shall not be keys, even if their combination is unique, because as keys those fields can't be edited on the UI. In this case a different key is required, for example a UUID, and the unique constraint for `project` and `user` can be expressed via `@assert.unique`.
```

renejeglinsky (approved):

guides/domain-modeling.md
```diff
@@ -570,7 +571,16 @@ entity Orders { ...
...
-Behind the scenes this will add an entity named `Orders.Items` with a backlink association named `up_`, so effectively generating the same model as above.
+Behind the scenes this will add an entity named `Orders.Items` with a backlink association named `up_`, so effectively generating the same model as above. You can annotate the inline composition with UI annotations as follows:
+
+```cds
+annotate Orders.Items with @(
+   UI.LineItem : [
+      {Value: pos},
+      {Value: quantity}
+   ]
+)
```
renejeglinsky, line 582:
```suggestion
      {Value: quantity},
   ],
);
```

stewsk (approved):

PR #1104, opened 2024-07-12
Fix element reference in CQN example on avoiding SQL injection

qmacro:
Pretty sure this should be `ID` not `title`, right?


johannes-vogel (approved):

qmacro:
Hope it's OK to add new minor fixes to this PR on the same file ...

renejeglinsky (approved):

PR #1116, opened 2024-07-18
Increase specificity for start page heading

swaldmann:
We have this problem again on our start page for small screens:

Screenshot 2024-07-18 at 10 13 23

Increasing the specificity of the `.VPHome .VPHero h1.name` styling helped reset it to the smaller font on mobile.


chgeo (approved):
Thanks!

PR #1131, opened 2024-07-23
Update hybrid-testing.md

preetamkajalrout:
Replace `-p` (unsupported) to `-4` to use the correct profile during `bind`


swaldmann:
Thanks @preetamkajalrout for the contribution!

PR #1134, opened 2024-07-23
cds: use "IsActiveEntity" in anno expressions

stewsk:


cds/cdl.md
```diff
@@ -1105,6 +1105,7 @@ Each path in the expression is checked:
 * If the annotation is assigned to a subelement of a structured element, the top level
   elements of the entity can be accessed via `$self`.
 * A parameter `par` can be accessed via `:par`, just like parameters of a parametrized entity in queries.
+* For a draft-enabled entity, element `IsActiveEntity` can be accessed via prefix `$draft`.
```
smahati, line 1108: would using the word 'property' instead of 'element' here be better?

stewsk: In CAP we usually use "element" (or maybe "field") rather than "property", so I wouldn't change it.

hjboth (approved):

bugwelle (approved):

cds/cdl.md
```diff
@@ -1105,8 +1105,10 @@ Each path in the expression is checked:
 * If the annotation is assigned to a subelement of a structured element, the top level
   elements of the entity can be accessed via `$self`.
 * A parameter `par` can be accessed via `:par`, just like parameters of a parametrized entity in queries.
-* For a draft-enabled entity, element `IsActiveEntity` can be accessed via prefix `$draft`.
-* If a path cannot be resolved successfully, compilation fails with an error.
+* The draft specific element `IsActiveEntity` can be referred to with the magic variable `$draft.IsActiveEntity`.
```
stewsk, line 1108: hm, compared to the other entries in this list, this is a rather long and detailed description for a special case

hjboth: yes that's right, it's a special case that (what I think) needs some more detailed explanation (all others in this list are straight forward and will always work)

stewsk: ok

cds/cdl.md
```diff
@@ -1105,7 +1105,10 @@ Each path in the expression is checked:
 * If the annotation is assigned to a subelement of a structured element, the top level
   elements of the entity can be accessed via `$self`.
 * A parameter `par` can be accessed via `:par`, just like parameters of a parametrized entity in queries.
-* If a path cannot be resolved successfully, compilation fails with an error.
+* The draft specific element `IsActiveEntity` can be referred to with the magic variable `$draft.IsActiveEntity`.
+  During draft augmentation `$draft.IsActiveEntity` is rewritten to `$self.IsActiveEntity` for all draft enabled
```
smahati, line 1109:
```suggestion
  During draft augmentation `$draft.IsActiveEntity` is rewritten to `$self.IsActiveEntity` for all root and sub nodes of draft enabled
```
does this rewrite mean that it's $draft.IAE in some handlers and then it's always $self.IAE?

cds/cdl.md
```diff
@@ -1105,7 +1105,10 @@ Each path in the expression is checked:
...
+  entities (root and sub nodes but not for named types or entity parameters).
```
smahati, line 1110:
```suggestion
  entities (but not for named types or entity parameters).
```

hjboth: sorry, no in a draft context the state elements are available in both the root and the sub nodes, that's important to know.

cds/cdl.md
hjboth: it has nothing to do with handlers...

cds/cdl.md
smahati: I mentioned that in the previous line.

hjboth: so what's the difference? In fact the content in parentheses are more like an additional information... Just leave it as it is, it's good enough

cds/cdl.md
smahati:
```suggestion
 It can be reffered to by `$self.IsActiveEntity` in all root and sub nodes of draft enabled
```

hjboth: no please not. `$self.IsActiveEntity` IS A REFERENCE, a reference can't be referenced.... Why is there always the urge to improve something that is already exactly what it should be?

smahati (approved):

PR #1140, opened 2024-07-25
Set `add -p` as beta

chgeo:


chgeo:
Move to here as beta as follow-up from our discussion.
@swaldmann what do you think?

swaldmann (approved):

PR #1144, opened 2024-07-26
Avoiding Transactions for Select

beckermarc:
Docs for new property `cds.persistence.changeSet.enforceTransactional: false`


java/event-handlers/changeset-contexts.md
```diff
@@ -169,3 +169,17 @@ public class SessionContextHandler implements EventHandler {
 
 }
 ```
+
+## Avoiding Transactions for Select
+
+CAP ensures that every interaction with a service is inside of a ChangeSet Context. However transactions are not started at that point in time yet.
+By default any kind of first interaction with the Persistence Service will begin the transaction. Once a transaction has been started a connection for that transaction is reserved from the connection pool. This connection is only returned to the connection pool on commit or rollback of the transaction.
```
BraunMatthias, line 176:
```suggestion
By default, any kind of first interaction with the Persistence Service will begin the transaction. Once a transaction has been started, a connection for that transaction is reserved from the connection pool. This connection is only returned to the connection pool on commit or rollback of the transaction.
```

java/event-handlers/changeset-contexts.md
```diff
@@ -169,3 +169,17 @@ public class SessionContextHandler implements EventHandler {
...
+
+However READ events, which run simple Select queries don't actually require transactions in most cases. When setting the property `cds.persistence.changeSet.enforceTransactional` to `false` most Select queries do not cause a transaction to be started any longer. A connection for these queries is obtained from the connection pool and returned immediately after executing the queries on the database. This can increase throughput of an application, by making connections available to other threads faster. As soon as a modifying statement is executed on the Persistence Service a transaction is started. All subsequent Select queries will participate in that transaction. Note, that this behaviour is only transparent when using the default transaction isolation level "Read Committed".
```
BraunMatthias, line 178:
```suggestion
However, `READ` events which run simple Select queries don't actually require transactions in most cases. When setting the property `cds.persistence.changeSet.enforceTransactional` to `false` most Select queries do not cause a transaction to be started any longer. A connection for these queries is obtained from the connection pool and returned immediately after executing the queries on the database. This can increase throughput of an application, by making connections available for concurrent requests faster. As soon as a modifying statement is executed on the Persistence Service, a transaction is started. All subsequent Select queries will participate in that transaction. Note, that this behaviour is only transparent when using the default transaction isolation level "Read Committed".
```

java/event-handlers/changeset-contexts.md
```diff
@@ -169,3 +169,17 @@ public class SessionContextHandler implements EventHandler {
...
+
+A ChangeSet Context can always be marked as requiring a transaction, by calling the `markTransactional` on the `ChangeSetContext` or `ChangeSetContextRunner`. The next interaction with the Persistence Service will guarantee to start a transaction in that case. Alternatively Spring Boot annotations `@Transactional` can be used to eagerly start a transaction.
```
BraunMatthias, line 180:
```suggestion
A ChangeSet Context can always be marked as requiring a transaction, by calling the `markTransactional` on the `ChangeSetContext` or `ChangeSetContextRunner`. The next interaction with the Persistence Service will guarantee to start a transaction in that case. Alternatively, Spring Boot annotations `@Transactional` can be used to eagerly start a transaction.
```

BraunMatthias (approved):

java/event-handlers/changeset-contexts.md
```diff
@@ -169,3 +169,17 @@ public class SessionContextHandler implements EventHandler {
 
 }
 ```
+
+## Avoiding Transactions for Select
```
BraunMatthias, line 173:
```suggestion
## Avoiding Transactions for Select { #avoid-transactions }
```

smahati (approved):

PR #1188, opened 2024-08-12
fix: @topic should use cloudevents format

sjvans:


David-Kunz:
lgtm, but for `format: 'cloudevents'`, Event Mesh needs to be configured properly, the subscription rules must allow `+/+/+` in its service descriptor, let me check if that's mentioned in the guide

David-Kunz (approved):

sjvans:
@smahati @renejeglinsky could you please have a look asap? this is needed for event broker docs

PR #1209, opened 2024-08-26
add(theme): add global 'scroll to top' button

chgeo:
* add "scroll to top" button
* only shows if scrolled down


chgeo:
@joergmann I have move the scroll to top feature to the external repo.

PR #1228, opened 2024-09-03
Hide scroll-to-top on iOS

swaldmann:
iOS already has a huge system-wide scroll-to-top button (the orange area here), so the button is a bit redundant. On small screens it also overlays the content quite a bit →  I'd disable it for iOS.

Screenshot 2024-09-03 at 13 10 29

### DIFF
--- a/.vitepress/theme/components/ScrollToTop.vue
+++ b/.vitepress/theme/components/ScrollToTop.vue
@@ -28,7 +28,7 @@ export default {
     },
   },
   mounted() {
-    const isIOS = /iPad|iPhone|iPod/.test(window.navigator.userAgent);
+    const isIOS = /iPad|iPhone|iPod/i.test(window.navigator.userAgent);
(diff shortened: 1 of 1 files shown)


.vitepress/theme/components/ScrollToTop.vue
```diff
@@ -28,11 +28,16 @@ export default {
     },
   },
   mounted() {
-    window.addEventListener('scroll', () => {
-      const isVisible = window.scrollY > 240;
-      this.enableFadeInClass = isVisible;
-      this.enableFadeOutClass = !isVisible;
-    });
+    const isIOS = /iPad|iPhone|iPod/.test(window.navigator.userAgent);
```
joergmann, line 31:
```suggestion
    const isIOS = /iPad|iPhone|iPod/i.test(window.navigator.userAgent);
```
Paranoic Regex ;-)

joergmann: besides this remark: +1

PR #1233, opened 2024-09-04
Update cds-server.md with working examples

mika-espoo:
Instructions are missing `env` property from the chain and thus have no effect. This patch fixes the documentation from this part.

### DIFF
--- a/node.js/cds-server.md
+++ b/node.js/cds-server.md
@@ -237,7 +237,7 @@ The behavior of the built-in `server.js` can be customized through the options d
 
 ### CORS Middleware
 
-The built-in CORS middleware can be enabled explicitly with `cds.server.cors = true`.  By default, this is `false` if in production.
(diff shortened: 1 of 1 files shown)


smahati:
adding @johannes-vogel to review this PR

chgeo (commented):
Well, `cds.env.server.cors = true` could mean
- Programmatic config with `cds.env.server.cors = true` (the literal expression, but used rarely)
- Config through `package.json` etc. with `cds: {server: { cors: true }}` with the shorthand notation `cds.server.cors:true` (much more common)

Looking at our docs, it's really a mixed bag with one or the other of the above options.  Whatever we currently do, there are users that will get confused.

This is why I plan to add a little 'config inspector' that would appear as a popover for such a config key showing the user all config options.

For now, the suggested fix is ok, I'd say.

smahati (approved):

PR #1237, opened 2024-09-05
feat: add disclaimer for deep search

patricebender:
Useful additions for the time being, some people are still on CDS 7 with `@sap/cds-hana` for production while using `@cap-js/sqlite` for development. In this scenario, deep search in production will not work, while it does during development.


guides/providing-services.md
```diff
@@ -383,6 +383,10 @@ Searches the `title` element only.
 
 ##### Extend Search to *Associated* Entities
 
+::: warning
```
renejeglinsky, line 386:
```suggestion
::: warning Node.js: Only w/ streamlined database services
```

renejeglinsky: No, I just want to give it a header so that it doesn't say "Warning".

renejeglinsky (approved):

PR #1258, opened 2024-09-17
Use `deploy --out` instead of stdout

chgeo:
The new param `cds deploy --out <file>` allows writing DDL statements to a separate file.

This is needed by CAP Java to replace their stdout redirection (`cds deploy --dry > schema.sql`). That one fails if debug traces are on as these mess up the output.

So,

```sh
cds deploy --to h2 --dry > schema.sql
```

can change to

```sh
cds deploy --to h2 --dry --out schema.sql
```

Also, one can now request the log independently of the `dry` mode to get an additional log:

```sh
cds deploy --to h2 --out schema.sql
```

### DIFF
--- a/guides/databases-postgres.md
+++ b/guides/databases-postgres.md
@@ -410,7 +410,7 @@ You can use `cds deploy` with option `--dry` to simulate and inspect how things 
 1. Capture your current model in a CSN file:
 
    ```sh
-   cds deploy --dry --model-only > cds-model.csn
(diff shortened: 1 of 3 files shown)


chgeo:
BTW: we have also added a check in `cds deploy` that will error if the `sap/cds` version is too old to understand `--out` (versions < 8.3.0).  This way we detect mismatches between `@sap/cds-dk` and `@sap/cds` versions.

mofterdinger (approved):

PR #1263, opened 2024-09-19
Update cds-typer doc

daogrady:
1. describe newly added `--useEntitiesProxy` to facilitate static top-level imports
2. add "Configuration" section to describe passing configuration parameters to cds-typer vis `cds.env`
3. update synopsis to latest version of cds-typer to include snake_cased parameter names

### DIFF
--- a/tools/cds-typer.md
+++ b/tools/cds-typer.md
@@ -400,7 +400,6 @@ Any CLI parameter described [above](#typer-cli) can also be passed to cds-typer 
   …
   "cds": {
     "typer": {
-      "propertiesOptional": true,
       "log_level": "DEBUG"
     }
   }
--- a/tools/cds-typer.md
+++ b/tools/cds-typer.md
@@ -313,9 +313,8 @@ The CLI offers several parameters which you can list using the `--help` paramete
 
 <!-- TODO: automatically pull command line options from cds-typer --help -->
 ```log
-> @cap-js/cds-typer@0.22.0 cli
+> @cap-js/cds-typer@0.26.0 cli
 > node lib/cli.js --help
-
 SYNOPSIS
 
   cds-typer [cds file | "*"]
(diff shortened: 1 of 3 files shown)


daogrady:
Not sure about the failing CI. Can reproduce locally, but 

```
(!) Found dead link ./.../plugins/index in file @external/about/index.md
(!) Found dead link ./../plugins in file @external/about/index.md
```

doesn't seem related to this PR.

chgeo:
Link errors are fixed

tools/cds-typer.md
```diff
@@ -326,52 +325,89 @@ SYNOPSIS
...
+
+Any CLI parameter described [above](#typer-cli) can also be passed to cds-typer via [`cds.env`](#cds-env), for example via your project's _package.json_:
+
+::: code-group
+```json [package.json]
+{
+  …
+  "cds": {
+    "typer": {
+      "propertiesOptional": true,
```
chgeo, line 403: Should promote the snake-case options:

```suggestion
      "properties_optional": true,
```

tools/cds-typer.md
```diff
@@ -326,52 +325,89 @@ SYNOPSIS
...
+
   --version
 
     Prints the version of this tool.
 ```
 :::
 
+### Configuration
+
+Any CLI parameter described [above](#typer-cli) can also be passed to cds-typer via [`cds.env`](#cds-env), for example via your project's _package.json_:
```
chgeo, line 395:
```suggestion
Any CLI parameter described [above](#typer-cli) can also be passed to cds-typer via [`cds.env`](../node.js/cds-env), for example via your project's _package.json_:
```

tools/cds-typer.md
daogrady: I wanted to showcase that both forms are still valid. Is there harm to be expected from users using camelCase?

chgeo: > I wanted to showcase that both forms are still valid. Is there harm to be expected from users using camelCase?

No harm.  Just wondered if it would look inconsistent to see both styles.  In the end, users will ask which style they are supposed to use.

daogrady: > which style they are supposed to use
Whichever they fancy. 🙂

But if we want to nudge them to snake_case, let's remove the one with camelCase altogether from the sample. No need for two examples then.

chgeo: > But if we want to nudge them to snake_case,

Yes, kind of, for `cds.env`.  Because we tell in other areas to use this style to be able to use env vars, i.e. `CDS_TYPER_PROPERTIES__OPTIONAL` here.

daogrady: ACK, I have removed the camelCase example accordingly.

chgeo (approved):

tools/cds-typer.md
```diff
@@ -511,7 +546,7 @@ class CatalogService extends cds.ApplicationService { init(){
 })
 ```
 
-Just as with `cds.entities(…)`, these imports can't be static, but need to be dynamic:
+Just as with `cds.entities(…)`, these imports can't be static, but need to be dynamic (caveat, [see below](#typer-top-level-imports)):
```
renejeglinsky, line 549:
```suggestion
Similar to `cds.entities(…)`, you can't use static imports here. Instead, you need to use dynamic imports. However, there's an exception for [static top-level imports](#typer-top-level-imports).
```

tools/cds-typer.md
```diff
@@ -543,3 +578,25 @@ class CatalogService extends cds.ApplicationService { async init(){
   const { Book } = await import('#cds-models/sap/capire/bookshop')
 }}
 ```
+
+### Static Top-Level Imports {#typer-top-level-imports}
+Starting with `cds-typer@0.26.0`, a new option `useEntitiesProxy` can be passed to `cds-typer` that allows you to statically import your entities on top-level, as you intuitively would. You can still only _use them_ in a context where the cds runtime is fully booted, like in a service definition:
```
renejeglinsky, line 583:
```suggestion
Starting with `cds-typer@0.26.0`, you can pass a new option, `useEntitiesProxy`, to `cds-typer`. This option allows you to statically import your entities at the top level, as you intuitively would. However, you can still only _use these entities_ in a context where the CDS runtime is fully booted, like in a service definition:
```

renejeglinsky (approved):

PR #1265, opened 2024-09-20
Clarify text

tim-sh:
When first reading this, I got 'previous' wrong…

### DIFF
--- a/guides/databases.md
+++ b/guides/databases.md
@@ -72,7 +72,7 @@ The afore-mentioned packages use `cds-plugin` techniques to automatically config
 }}
 ```
 
-::: details In contrast to pre CDS 7 setups this means...
+::: details In contrast to pre-CDS 7 setups this means...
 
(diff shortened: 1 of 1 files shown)


guides/databases.md
```diff
@@ -83,7 +83,7 @@ The afore-mentioned packages use `cds-plugin` techniques to automatically config
 
 ### Custom Configuration  {.impl .node}
 
-The previous setups auto-wire things through configuration presets, which are automatically enabled via `cds-plugin` techniques. You can always use the basic configurations for other setups, or override individual properties as follows:
+The above-mentioned setup auto-wires things through configuration presets, which are automatically enabled via `cds-plugin` techniques. You can always use the basic configuration and override individual properties to create a different setup:
```
renejeglinsky, line 86: I like this sentence! Only "above" is sth we need to change due to accessibility.

```suggestion
The previously mentioned setup auto-wires things through configuration presets, which are automatically enabled via `cds-plugin` techniques. You can always use the basic configuration and override individual properties to create a different setup:
```

or

```suggestion
The auto-wired configuration uses configuration presets, which are automatically enabled via `cds-plugin` techniques. You can always use the basic configuration and override individual properties to create a different setup:
```

tim-sh: Thanks :) The lower sounds simple and good :+1:

renejeglinsky (approved):

PR #1274, opened 2024-09-24
Adjust docs for @cds.java.name

beckermarc:


java/cds-data.md
```diff
@@ -512,48 +512,118 @@ See the following example:
...
   void setClazz(String clazz);
-	...
+
 }
 ```
 
 #### Renaming Types in Java
 
-You might also want to rename the type of the entity. For this you can use annotation `@cds.java.this.name` to specify alternative name for the  accessor interfaces and [static model](./cqn-services/persistence-services#staticmodel) interfaces. This annotation can be used only on definitions and is ignored everywhere else.
+You might also want to rename the interface for an entity or type. You can also use the annotation `@cds.java.name` to specify an alternative name for the accessor interfaces and [static model](./cqn-services/persistence-services#staticmodel) interfaces.
```
agoerler, line 532: We should not promote using `@cds.java.name` to rename types. I would not given such an example here. Let's rather put a warning box that using `@cds.java.name` to rename types might get you into trouble.

java/cds-data.md
```diff
@@ -512,48 +512,118 @@ See the following example:
...
+```
 
-  @CdsName(ID)
-  String id();
+Annotations are by default propagated along projections or from types to (flattened) elements using that type.
+This can create unexpected behaviour when renaming Java interfaces or attributes, especially when using (structured) types.
 
-  @CdsName(ID)
-  MyJavaClass id(String id);
+Therefore an alternative annotation `@cds.java.this.name` can be used as well.
```
agoerler, line 553:
```suggestion
Therefore the `@cds.java.this.name` annotation should be used instead.
```

java/cds-data.md
```diff
@@ -512,48 +512,118 @@ See the following example:
...
+This annotation is not propagated and therefore only applies at exactly the CDS definition where it was originally placed.
```
agoerler, line 554:
```suggestion
This annotation is not propagated and therefore only applies at exactly the CDS definition where it is placed.
```

java/cds-data.md
beckermarc: I rephrased a little to make this more clear.

java/cds-data.md
```diff
@@ -512,48 +512,117 @@ See the following example:
...
-	...
+
 }
 ```
 
-#### Renaming Types in Java
+#### Renaming Entities in Java
 
-You might also want to rename the type of the entity. For this you can use annotation `@cds.java.this.name` to specify alternative name for the  accessor interfaces and [static model](./cqn-services/persistence-services#staticmodel) interfaces. This annotation can be used only on definitions and is ignored everywhere else.
+You can also use the annotation `@cds.java.name` to specify an alternative name for the accessor interfaces and [static model](./cqn-services/persistence-services#staticmodel) interfaces.
```
agoerler, line 532: I don't think we should promote this. Besides the issue with types which you explain below we also have the issues with includes and projections:

```
@cds.java.name: 'X'
entity E;

@cds.java.name: 'U'
type T; 

entity F : T;  -> F "inherits" the Java name 'U'

entity G as projection on E; -> G gets the Java name 'X'
```

beckermarc: For projections I think it is very reasonable, as the case you are mentioning here is rather unlikely. Most often it is `entity E as projection on db.E`, for which the propagation is actually pretty useful. The include is a good example for using @cds.java.this.name, that's right. 

I think in general the recommendation would be @cds.java.this.name on types. But the problems with propagation are also mentioned in the docs explicitly. Why not explain the differences like I do right now? Is this not explicit enough for you?

agoerler: So your recommendation would be:

element -> @cds.java.name
entity 
  if used in typical 1:1 projection -> @cds.java.name 
  if used in weird semantic-changing projection -> @cds.java.this.name
type -> @cds.java.this.name (element's type) 
aspect 
  as "includee" -> @cds.java.this.name (as the aspect is typically "Included" with `:`)
  as composition target @cds.java.name (as you can't redefine it, I think)
  
Overall I think it depends on how the annotated definition is to be used:

element -> @cds.java.name
as select from / as projection on -> depends on how fancy the projection is
inheritance `:` -> @cds.java.this.name
element type `:`-> @cds.java.this.name

beckermarc: I wouldn't make this differentiation on projections. First of all an annotation can always also explicitly be stopped from propagation by using something like `@cds.java.name: null` or `@cds.java.name: 'SomeOtherName'`. This behaviour can be easily explained and is also like that for most other annotations.

@cds.java.this.name from my perspective really helps in the situations where annotation propagation is not so intuitive. This mostly happens with types from my perspective, where annotations move from the type to the element.

agoerler: > is not so intuitive. This mostly happens with types

and with aspects

beckermarc: I rewrote the docs to prefer @cds.java.this.name over @cds.java.name.

java/cds-data.md
```diff
@@ -512,48 +512,119 @@ See the following example:
...
+
+  MyName getMyName();
+  void setMyName(MyName myName);
+
+  MyName getMyName();
+  void setMyName(MyName myName);
+}
+```
+
+Notice, that the propagated annotation `@cds.java.name` creates attribute and method conflicts in `Person`.
```
agoerler, line 625: Put this in a warning box?

agoerler (approved):
Looks good to me. :-)

java/cds-data.md
beckermarc: I'd keep it a simple text, as it is already within a details box, that you need to explicitly expand.

renejeglinsky (approved):

PR #1276, opened 2024-09-25
fix sample audit log

DanSchlachter:
those additional lines are needed, otherwise no audit log event is written.
It seems that if those lines are missing, no audit log is written but also no error / warning is shown? -> Maybe that should be there in the plugin, too.


sjvans:
> Maybe that should be there in the plugin, too.

what do you mean by that? https://github.com/cap-js/audit-logging?tab=readme-ov-file#annotate-personal-data is correct...

DanSchlachter:
the sample on the docs is missing `DataSubjectRole` and `EntitySemantics`. Without it no audit log is written.
The sample on github is correct.

sjvans:
> the sample on the docs is missing `DataSubjectRole` and `EntitySemantics`. Without it no audit log is written. The sample on github is correct.

got it, sorry. yes, should probably be there.

sjvans (approved):

plugins/index.md
```diff
@@ -219,7 +219,10 @@ Available for:
 The new Audit Log plugin provides out-of-the box support for logging personal data-related operations with the [SAP Audit Log Service](https://discovery-center.cloud.sap/serviceCatalog/audit-log-service). All we need is annotations of respective entities and fields like that:
 
 ```cds
-annotate my.Customers with @PersonalData {
+annotate my.Customers with @PersonalData  : {
```
sjvans, line 222:
```suggestion
annotate my.Customers with @PersonalData : {
```

renejeglinsky (approved):

PR #1294, opened 2024-10-01
Workflow to add the output for `cds-typer --help` in typer section

daogrady:
The [documentation for cds-typer](https://cap.cloud.sap/docs/tools/cds-typer#usage-options) contains the output of calling `cds-typer --help`. This improves searchability for both search engines and AI. We currently update this part of the docs from time to time, so as of now, the help text could be out of date for some time before being brought up to date.

This PR introduces an action that can be triggered manually and automatically runs every Wednesday morning to automatically grab the latest version of the help text. If there are any updates, it will open a PR with the new content.


renejeglinsky:
This PR made me look for other places where we document the CLI help. There's not a lot but I found that we used specific styling which isn't used here:
image
Source: https://github.com/cap-js/docs/blob/main/tools/cds-cli.md#L94

@chgeo I'm not sure if the green highlighting was how it used to look. Maybe sth has changed?

General question: Should add or remove this styling at all places?

daogrady:
> we used specific styling

Good catch, I wasn't even aware!

A few thoughts on that:

If we find that we'd like to have `--help` doc for other CLI tools in CAPire as well (esp. for Google, AI), we could also integrate them into this script in the future, which would result in the same colouring-discrepancy.
While the styling that is currently used for the output of `cds help` is obviously more colourful, code fences with `log` also come with a certain level of colouring (which is the style that is currently in use in the cds-typer section.

Screenshot 2024-10-01 at 12 19 56

As this is also the style currently used in the cds-typer section, I guess we can at least postpone the style adjustments?
If we want to stick with the current styling and also have it streamlined throughout CAPire, then I will gladly look into doing so in a follow-up.

tl;dr: I'd be in favour of dropping the custom style for `log` fences, but not at all opposed to sticking with the existing style.

.github/workflows/extract-docs.yml
```diff
@@ -0,0 +1,52 @@
...
+    # Runs every Wednesday at 02:45 AM (UTC)
+    - cron: '45 2 * * 3'
+
+jobs:
+  run_script_and_create_pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
chgeo, line 15:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/extract-docs.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
```
chgeo, line 18:
```suggestion
        uses: actions/setup-node@v4
```

.github/workflows/extract-docs.yml
```diff
@@ -0,0 +1,52 @@
...
+          BRANCH_NAME="update-cds-typer-synopsis-$(date +'%Y%m%d%H%M%S')"
+          git checkout -b "$BRANCH_NAME"
+          git push origin "$BRANCH_NAME"
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ steps.push_changes.outputs.BRANCH_NAME }}
+          title: "Automated PR: Update synopsis for cds-typer"
```
chgeo, line 50:
```suggestion
          title: "chore: Update synopsis for cds-typer"
```

.github/workflows/extract-docs.yml
```diff
@@ -0,0 +1,52 @@
...
+          body: "This PR was automatically generated and updates the output of `cds-typer --help` to the latest version."
```
chgeo, line 51:
```suggestion
          body: "Updates the output of `cds-typer --help` to the latest version."
```

chgeo (approved):

chgeo:
Let's merge this to see if the workflow worlks.

chgeo:
It failed: https://github.com/cap-js/docs/actions/runs/11128146329/job/30922182608
@daogrady can you check?

chgeo:
Continued in #1301

PR #1301, opened 2024-10-01
Update more CLI texts

chgeo:
Make #1294 more generic and update more CLI texts.

- Use the `<pre class="log">` approach instead of `log` code fence as it allows us to style the output more.
- Convert known terminal escape sequences to formats (mainly `<em>`, `<i>`, `<strong>`).
- Remove the versions numbers from the output as these would produce diffs for each new version, although the texts have not changed.
- Fix the workflow and produce a PR with a stable title. Parallel PRs seem unnecessary.
  - GH limitation: that PR cannot trigger new workflows. Workaround is to close and reopen the PR.
  - Example is #1303 

This also gets rid of the [custom renovate config to bump inner-md versions](https://github.com/cap-js/docs/blob/810cd2beedab1f0d9aa6ae0777c8b79a69f8f067/.github/renovate.json#L58-L117) and the corresponding PRs.


renejeglinsky:
I'd like to have a diff even if only the version number changed and not the text. Why not keep the versions up to date?

.github/cli/grab-cli-texts.js
```diff
@@ -0,0 +1,53 @@
...
+
+    // some very basic plausibility checks to make sure we don't
+    // end up with garbage or npx errors in the markdown
+    if (!/\d+\.\d+\.\d+/.test(version)) {
+        throw new Error(`unexpected version: ${version}`)
+    }
+    if (!cmdOut) {
+        throw new Error(`no output from: ${cmdString}`)
+    }
+    if (cmd.includes('help') && !/SYNOPSIS|USAGE/.test(cmdOut)) {
```
daogrady, line 47: Seeing that this function is now being generalised, I'd be in favour of either dropping this check or making it very generous. Maybe something like "contains line breaks". Otherwise, this will be a high-maintenance piece of code whenever a new command is covered by the workflow.

.github/cli/grab-cli-texts.js
```diff
@@ -0,0 +1,53 @@
...
+        .replace(/^\x1b\[1m(.*?)\x1b\[0m\n/gm, '<strong>$1</strong>') // bold at beginning of line -> strong
+        .replace(/(\s*)\x1b\[4m(.*?)\x1b\[0m/g, '$1<i>$2</i>') // underline -> i
+        .replace(/(\s*)\x1b\[\d+m(.*?)\x1b\[0m/g, '$1<em>$2</em>') // other colors -> em
+    , `${unstyled ? '```' : '</pre>'}`
+].join('\n')
+
+try {
+    const version = (await exec(`npm view ${pkg} version`)).stdout.trim()
+    const cmdString = `npm exec --package=${pkg}@${version} -c "${cmd}"`
+    console.error(`> ${cmdString}`)
```
daogrady, line 36: why `console.error`?

daogrady:
> I'd like to have a diff even if only the version number changed and not the text. Why not keep the versions up to date?

Was also my first impulse, but I see that the version number is no longer part of the output, so there wouldn't be any confusion for the user if they're looking at outdated versions. Plus if the version was still part of the output, this workflow would produce PRs for every patch version of our tools.
So I think this change makes sense, but if we do want to include the version number again, we should at least restrict it to major.minor.

daogrady (approved):
Looks great, thanks for making this generalisation!

I played around with the script and it looks fine to me. My gripe with contributing workflows is that they're hard to test, so I'd say merge and see what happens[^1]? 🥲

[^1]: unless the other reviewers have change-requests, ofc.

.github/cli/grab-cli-texts.js
chgeo: To not interfere w/ the output on stdout as this will be grabbed by the caller.

daogrady:
```suggestion
    // stdout is redirected to file -> use stderr for debugging output
    console.error(`> ${cmdString}`)
```

chgeo:
> Was also my first impulse, but I see that the version number is no longer part of the output, so there wouldn't be any confusion for the user if they're looking at outdated versions. 

Do you think the user is interested in the exact version?  I doubt that and think they just assume they're looking at the latest text.  Plus, the former output was a bit distracting like

```
> @sap/cds-dk@8.3.0 cds --help
```

while this is what users know (and type):

```
> cds --help
```


> Plus if the version was still part of the output, this workflow would produce PRs for every patch version of our tools. 

This is what happens today through a [custom renovate config](https://github.com/cap-js/docs/blob/810cd2beedab1f0d9aa6ae0777c8b79a69f8f067/.github/renovate.json#L58-L117), which turned out to produce quite a lot of rather dump 'just bump the version' PRs ([example](https://github.com/cap-js/docs/pull/1297/files)).  This is what I want to reduce.

chgeo:
> My gripe with contributing workflows is that they're hard to test, so I'd say merge and see what happens? 🥲

I already tested the workflow.  Turns out you can trigger workflows from non-main branches through the `gh` command line.

```sh
gh workflow run 'PR Latest cds-typer Output' --ref cli-out
```

What is strange though that I sometimes needed to use the old name _PR Latest cds-typer Output_ while in some instances I could use the name name _Update CLI Texts_.  Looks like a GH bug or so.  See the latest [run](https://github.com/cap-js/docs/actions/runs/11139662677/job/30956790519).

.github/cli/grab-cli-texts.js
chgeo: Done

daogrady:
Sorry, my comment was a bit vague. I initially had the same reaction as René, but after some consideration I came to the same conclusions as you did. So I think removing the version from the output is fine and even if we should reintroduce them, we should at least leave out the patch version.

daogrady:
> Turns out you can trigger workflows from non-main branches through the [`gh` command line tool](https://github.com/cli/cli).
> 
> ```shell
> gh workflow run 'PR Latest cds-typer Output' --ref cli-out
> ```

Awesome! 😮

chgeo:
> I'd like to have a diff even if only the version number changed and not the text. Why not keep the versions up to date?

In the files where versions explicitly appear (like `@sap/cds: 8.3.0`), we do keep them up to date now, just not by single PRs from renovate but by only one for all texts.

The other thing @daogrady and I discussed was whether they should appear in command prompts like `cds@8.3.0 --help`, and here I'd say that's too confusing to read,

chgeo:
Can we merge this?

renejeglinsky (approved):

swaldmann (approved):

PR #1305, opened 2024-10-02
Event Broker: ORD is available

renejeglinsky:


sjvans:
@renejeglinsky we should verify first ;)

sjvans:
> Alright, then please verify what we assumed could be true in the future, as the future is now :)

yes and no, as other components (e.g., BTP Cockpit) are involved in the end-to-end process

PR #1308, opened 2024-10-02
rename "SAP Event Broker" to "SAP Cloud Application Event Hub"

David-Kunz:


David-Kunz:
Ok, we can merge this. @renejeglinsky , can we also add this to the September  release?

renejeglinsky:
> Ok, we can merge this. @renejeglinsky , can we also add this to the September release?

It is already :)

PR #1328, opened 2024-10-09
docs(java): Add `@CdsName` to example

bugwelle:
A section in "Event Handlers" mentions `@CdsName`, but does not show any example.  This commit adds an example and fixes a few typos/grammar issues.

It also fixes the name `AddReviewContext`, which either changed in CAP Java or was wrong before, because for the bound action, `BooksAddReviewContext` is generated.


java/event-handlers/indicating-errors.md
```diff
@@ -42,7 +42,7 @@ throw new ServiceException(ErrorStatuses.CONFLICT, "Not enough stock available")
 throw new ServiceException(ErrorStatuses.BAD_REQUEST, "No book title specified", originalException);
 ```
 
-The OData V4 adapter turns all exceptions into an OData error response to indicate the error to the client.
+The OData adapters turn all exceptions into an OData error response to indicate the error to the client.
```
bugwelle, line 45: I clarified this with @eugene-andreev : it's not just the OData v4 adapter.

java/event-handlers/indicating-errors.md
```diff
@@ -273,7 +273,7 @@ The same applies to message targets that refer to an action or function input pa
 
 ``` java
 @Before
-public void validateReview(AddReviewContext context) {
+public void validateReview(BooksAddReviewContext context) {
```
bugwelle, line 276: tested locally: `BooksAddReviewContext` is generated

java/event-handlers/request-contexts.md
```diff
@@ -279,7 +279,8 @@ public class CustomUserInfoProvider implements UserInfoProvider {
             }
         }
         if (userInfo != null) {
-            userInfo.setName(userInfo.getName().toLowerCase()); // Normalize user name
+            // Normalize user name
+            userInfo.setName(userInfo.getName().toLowerCase(Locale.US));
```
bugwelle, line 283: `getName()` returns a string, right? If so, then the standard method `toLowerCase()` should get a locale

rjayasinghe: > `getName()` returns a string, right? If so, then the standard method `toLowerCase()` should get a locale

You're right. But I would take Locale.ENGLISH.

rjayasinghe (commented):
small inline comment regarding locale for string operations.

java/event-handlers/request-contexts.md
bugwelle: I used what I found in `cds-services` :D

But sure!

```suggestion
            userInfo.setName(userInfo.getName().toLowerCase(Locale.ENGLISH));
```

java/event-handlers/request-contexts.md
```diff
@@ -162,7 +162,7 @@ public void afterHandler(EventContext context){
...
 
-The application offers an action for one of its CDS entities. Within the action, a communication happens with a remote CAP service using an internal technical user from the provider account. The corresponding `on` handler of the action needs to create a new Request Context by calling `requestContext()`. Using the `systemUserProvider()` method, the existing user information is removed and the tenant is automatically set to the provider tenant. This allows the application to perform an HTTP call to the remote CAP service, which is secured using the pseudo-role `internal-user`.
+The application offers an action for one of its CDS entities. Within the action, the application communicate with a remote CAP service using an internal technical user from the provider account. The corresponding `on` handler of the action needs to create a new Request Context by calling `requestContext()`. Using the `systemUserProvider()` method, the existing user information is removed and the tenant is automatically set to the provider tenant. This allows the application to perform an HTTP call to the remote CAP service, which is secured using the pseudo-role `internal-user`.
```
rjayasinghe, line 165: "the application communicates with .."

rjayasinghe (commented):
only one small typo (comment inline)

java/event-handlers/request-contexts.md
bugwelle:
```suggestion
The application offers an action for one of its CDS entities. Within the action, the application communicates with a remote CAP service using an internal technical user from the provider account. The corresponding `on` handler of the action needs to create a new Request Context by calling `requestContext()`. Using the `systemUserProvider()` method, the existing user information is removed and the tenant is automatically set to the provider tenant. This allows the application to perform an HTTP call to the remote CAP service, which is secured using the pseudo-role `internal-user`.
```

bugwelle: Oh, thanks! :)

smahati (approved):

PR #1341, opened 2024-10-14
event-broker: add xsuaa-cross-consumption

David-Kunz:


johannes-vogel:
ok for now, once available we should use cds facet `cds add ias`

johannes-vogel (approved):

renejeglinsky (approved):

PR #1356, opened 2024-10-17
fix(odata): Fix `<Record>` example for `null` value

bugwelle:
Due to annotation flattening, the value `null` would _not_ be rendered as `<Null/>` in OData EDMX.  Instead, as the section before that mentions, we need to use special `$edmJson` syntax or the new expression syntax `(null)`.

Tested with cds-compiler v5.3.2

### DIFF
--- a/advanced/odata.md
+++ b/advanced/odata.md
@@ -336,18 +336,14 @@ Primitive annotation values, meaning Strings, Numbers, `true`, and `false` are m
 <Annotation Term="Some.String" String="foo"/>
 ```
 
-Rendering a `null` value must be done as dynamic expression:
+Rendering a `null` value must be done as dynamic expression or as an [annotation expression](#expression-annotations):
 
 ```cds
 @Some.Null: { $edmJson: { $Null } }
-```
-
-or as an [annotation expression](#expression-annotations):
-
-```cds
+// or
 @Some.Null: (null)
 ```
-
+Both result in the following:
 ```xml
 <Annotation Term="Some.Null">
   <Null/>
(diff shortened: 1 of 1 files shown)


advanced/odata.md
```diff
@@ -360,7 +360,7 @@ Record-like source structures are mapped to `<Record>` nodes in EDMX, with primi
 
 ```cds
 @Some.Record: {
-  Null: null,
+  Null: { $edmJson: { $Null } },
```
bugwelle, line 363: If there is special handling of record-annotation-types, I stand corrected. However, due to annotation flattening, such entries wouldn't be propagated properly anyway.

hjboth: yes that's wrong especially as a couple of lines `@Some.Null` explains how it should be. However, all this was written before we had this:

```cds
 @UI.rec: { Null: (null) }
```
renders into:

```
 <Annotation Term="UI.rec">
          <Record>
            <PropertyValue Property="Null">
              <Null/>
            </PropertyValue>
          </Record>
        </Annotation>
```

hjboth (approved):
lgtm, you probably want to rewrite the null anno with expression syntax?

advanced/odata.md
```diff
@@ -342,6 +342,12 @@ Rendering a `null` value must be done as dynamic expression:
 @Some.Null: { $edmJson: { $Null } }
 ```
 
+or as an [annotation expression](#expression-annotations):
+
+```cds
+@Some.Null: (null)
+```
+
```
renejeglinsky, line 350: How about making those two variants in a code-group? The following XML is the result, right? I guess this would be more obvious if we have the code-group. Otherwise I'd like to have a sentence like "Both result in the following:"

bugwelle: Updated. :)

renejeglinsky (approved):

PR #1357, opened 2024-10-17
style(odata): Use blockquote-style note for `@Some`, like other sections

bugwelle:
There are four notes about `@Some`, but only one uses the "tip" box style.  The others use the blockquote-style.


renejeglinsky (approved):
Thank you!

PR #1359, opened 2024-10-18
explain garbage collection in more detail

johannes-vogel:
@David-Kunz Does it make sense to you to have a more detailed explanation when drafts are deleted?
It was not clear to https://github.com/cap-js/cds-dbs/issues/839


node.js/fiori.md
```diff
@@ -146,6 +146,10 @@ You can set the property to one of the following:
 - number of hours like `'72h'`
 - number of milliseconds like `1000`
 
+::: info Technical background
+It can occur that inactive drafts are still in the database after the configured timeout. The deletion is implemented as a side effect of creating new drafts and there is no periodic job that does the garbage collection.
```
smahati, line 150:
```suggestion
It can occur that inactive drafts are still in the database after the configured timeout. The deletion is implemented as a side effect of creating new drafts and there's no periodic job that does the garbage collection.
```

smahati (approved):

renejeglinsky:
Hi @David-Kunz ,
can you quickly have a look?

David-Kunz (approved):

PR #1360, opened 2024-10-18
Config inspector

chgeo:
image

Usage: `<Config>cds.foo=bar</Config>` shows possible configuration variants
for this config key and value.

### DIFF
--- a/.vitepress/theme/components/ConfigInspect.vue
+++ b/.vitepress/theme/components/ConfigInspect.vue
@@ -1,0 +1,149 @@
+<template>
+  <VDropdown theme="cfgPopper" :distance="6" :triggers="['click', 'hover']" :delay="300" :popperTriggers="['hover']"
+  >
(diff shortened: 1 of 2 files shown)


chgeo:
@renejeglinsky @swaldmann the config inspector that I mentioned some time ago.

What do you think?
Samples in cap/docs#1110, but you can easily add you own.

renejeglinsky:
If I select one of the options, the pop-over collapses. Tested it in Edge and Chrome. Is that expected b/c it's not 100% finished yet?
Overall I like the idea. And if it's that simple to use: very nice!

chgeo:
> If I select one of the options, the pop-over collapses. 

This should be fixed by now.  

It's not finished indeed.  Needs some more cleanup and styling.  Syntax highlighting would be nice.
Can also discuss this today, time-permitting.

chgeo:
Let's merge this to get more feedback by more usages.

PR #1368, opened 2024-10-23
postgres: Make it clear that manual changes break automatic schema ev…

simonoswald:
…olution

### DIFF
--- a/guides/databases-postgres.md
+++ b/guides/databases-postgres.md
@@ -370,7 +370,7 @@ When redeploying after you changed your CDS models, like adding fields, automati
 
 > You can disable automatic schema evolution, if necessary, by setting `cds.requires.db.schema_evolution = false`.
 
-::: danger
+::: danger No manual altering
 
 Manually altering the database will most likely break automatic schema evolution!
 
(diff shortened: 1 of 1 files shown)


guides/databases-postgres.md
```diff
@@ -355,6 +355,12 @@ You can package and deploy that application, for example using [MTA-based deploy
 
 ## Automatic Schema Evolution { #schema-evolution }
 
+::: danger
+
+Manually altering the database will most likely break automatic schema evolution!
+
+:::
```
renejeglinsky, line 362: Maybe we can put it more like a hint, that if one wants to benefit from automatic schema evolution not manual changes are allowed. Otherwise it looks more like "Do not use it!" if we start the section with a danger note. And that's not what we want to convey, right?
We could in addition add a troubleshooting entry if you see this as an option.

renejeglinsky: Then the placement needs to be different. We cannot start with a danger note.

simonoswald: I've changed the placement - still prominent enough IMO.

Big problem: I think a lot of people don't read about schema evolution before using it (since it is enabled by default). But we can't really solve that I think.

guides/databases-postgres.md
```diff
@@ -370,6 +370,12 @@ When redeploying after you changed your CDS models, like adding fields, automati
 
 > You can disable automatic schema evolution, if necessary, by setting `cds.requires.db.schema_evolution = false`.
 
+::: danger
```
renejeglinsky, line 373:
```suggestion
::: danger No manual altering
```

renejeglinsky (approved):

PR #1369, opened 2024-10-23
Updated the default behaviour of the messaging error handler

dimamost:


java/messaging.md
```diff
@@ -679,6 +679,44 @@ private void handleError(MessagingErrorEventContext ctx) {
       ctx.setResult(true); // acknowledge
     }
 }
+```
+
+ In the multitenant application, there may be scenarios in which the client is not yet or no longer available. In this case, the message cannot be processed for the client because the client context is not available. In this case, the standard error handler acknowledges the message to prevent it from getting stuck in the message sequence. To change this behavior, the custom error handler from the example above can be extended by checking the exception type of the unknown tenant.
```
BraunMatthias, line 684: I think you should rather name the problem with subscriptions/races etc. here. Otherwise it's not clear for customers what it is actually about. The error message also explicitly refers to the tenant.

java/messaging.md
```diff
@@ -681,6 +681,44 @@ private void handleError(MessagingErrorEventContext ctx) {
...
+      // error handling for infrastructure error
+      ctx.setResult(false); // no acknowledgement
+    
+    } else if (errorCode.equals(CdsErrorStatuses.TENANT_NOT_EXISTS.getCodeString())) {
+      // error handling for unknown tenant context
+      
+       // tenant of the received message
+      String tenant = ctx.getTenant();
+
+      // receieved message
```
renejeglinsky, line 703:
```suggestion
      // received message
```

smahati (approved):

java/messaging.md
```diff
@@ -681,6 +681,44 @@ private void handleError(MessagingErrorEventContext ctx) {
 }
 ```
 
+In the multitenant application, there may be race conditions in the subscription process. In this case, the message cannot be processed for the tenant because the tenant context is not available. In this case, the standard error handler acknowledges the message to prevent it from getting stuck in the message sequence. To change this behavior, the custom error handler from the example above can be extended by checking the exception type of the unknown tenant.
```
BraunMatthias, line 684:
```suggestion
In a multi-tenant setup with several microservices, messages of a tenant not yet subscribed to the own microservice would be already received from the message queue. In this case, the message cannot be processed for the tenant because the tenant context is not yet available. By default, the standard error handler still acknowledges the message to prevent it from getting stuck in the message sequence. To change this behavior, the custom error handler from the example above can be extended by checking the exception type of the unknown tenant. 
```

BraunMatthias (approved):

PR #1371, opened 2024-10-23
Use more config inspector

chgeo:


chgeo:
@renejeglinsky this should also fix the warning with the config inspector

PR #1377, opened 2024-10-24
IN subquery; case expression

agoerler:


java/working-with-cql/query-api.md
```diff
@@ -1431,6 +1431,23 @@ Scalar functions are values that are calculated from other values. This calculat
...
+Select.from(PERSONS).columns(
+  p -> p.name(),
+  p -> p.when(p.age().lt(25)).then("young")
+        .when(p.age().ge(75)).then("elderly")
+        .orElse("adult").as("ageTxt").type(CdsBaseType.String));
+```
+
+::: tip
+You should specify the result type so that the runtime case respect it accordingly.
+:::
```
MattSchur, line 1448:
```suggestion
#### Case-When-Then Expressions

Use a case expression to compute a value based on the evaluation of conditions. The following query converts the stock of Books into a textual representation as 'stockLevel':

```java
Select.from(BOOKS).columns(
  b -> b.title(),
  b -> b.when(b.stock().lt(10)).then("low")
        .when(b.stock().gt(100)).then("high")
        .orElse("medium").as("stockLevel").type(CdsBaseType.STRING));
```

::: tip
Specify the result type so that the runtime case respects it accordingly.
:::
```

File: java/working-with-cql/query-api.md [hunk 2]
```diff
@@ -1908,7 +1920,24 @@ Select.from("Authors").where(CQL.exists(subquery));
...
+An `IN` subquery is used to test if an element (or tuple of elements) of an outer query is contained in the result of a subquery. You can use an `IN` subquery in fluent style as well as in fluent style:
 
+```java
+// fluent style
+Select.from(AUTHORS).where(author -> author.name().in(
+    Select.from(ASTRONAUTS).columns(astro -> astro.name())
+));
+```
+
+or in tree style. In this example we check wether the tuple (`firstName`, `lastName`) is contained in the result of the subquery:
```
MattSchur, line 1934:
```suggestion
or in tree style. In this example we check whether the tuple (`firstName`, `lastName`) is contained in the result of the subquery:
```

MattSchur (approved):

smahati (approved):

PR #1382, opened 2024-10-25
Avoid wording about 'lock-in to BTP'

chgeo:
The current language of a 'lock-in to SAP BTP' is unfortunate and could be misunderstood.  Make it a bit more neutral.

### DIFF
--- a/node.js/authentication.md
+++ b/node.js/authentication.md
@@ -344,7 +344,7 @@ npm add @sap/xssec
 [See **XSUAA in Hybrid Setup** below for additional information of how to test this](#xsuaa-setup){.learn-more}
 
 ::: warning
-Only use this authentication kind if it's required for your use case, as it binds you to a specific authentication service on SAP BTP.  Adjustments might be necessary if you want to change to a different authentication service in the future.
+Only use this authentication kind if it's required for your use case, as it binds you to a specific authentication service on SAP BTP.  Adjustments might be necessary if you want to change to a different BTP authentication service in the future.
 :::
 
 
--- a/node.js/authentication.md
+++ b/node.js/authentication.md
@@ -344,7 +344,7 @@ npm add @sap/xssec
 [See **XSUAA in Hybrid Setup** below for additional information of how to test this](#xsuaa-setup){.learn-more}
 
 ::: warning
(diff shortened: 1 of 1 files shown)


chgeo:
@D033039 what do you think?

knutheusermann:
Hi colleagues, SAP wants to promote BTP and therefore I would rather outline this approach as the preferred approach with a sentence like: "NOTE: This approach is recommended as it follows the SAP BTP Developer's Guideline and ensures that your application conforms to the SAP standard".

node.js/authentication.md
```diff
@@ -344,7 +344,7 @@ npm add @sap/xssec
 [See **XSUAA in Hybrid Setup** below for additional information of how to test this](#xsuaa-setup){.learn-more}
 
 ::: warning
-It's recommended to only use this authentication kind if it's necessary for your use case, as it denotes a lock-in to SAP BTP.
+Only use this authentication kind if it's required for your use case, as it binds you to a specific authentication service on SAP BTP.  Adjustments might be necessary if you want to change to a different authentication service in the future.
```
smahati, line 347:
```suggestion
Only use this authentication kind if it's required for your use case, as it binds you to a specific authentication service on SAP BTP.  Adjustments might be necessary if you want to change to a different BTP authentication service in the future.
```

chgeo:
> I would rather outline BTP value propositions and XSUAA and IAS as preferred approaches with a sentence like:
"NOTE: This approach is recommended as it follows the SAP BTP Developer's Guideline and ensures that your application conforms to SAP product standards".

Indeed, we seem to have a mismatch here:
- `cds add xsuaa` sets `auth: xsuaa` straight away
- This page says only set `xsuaa` if needed, otherwise keep to the default of `jwt` (which is almost the same but allows no programmatic access to xsuaa-specific attributes)

Need to discuss this internally.  @renejeglinsky something for our next sync maybe.

smahati:
Added to the [agenda](https://github.tools.sap/orgs/cap/projects/21/views/2?pane=issue&itemId=49439)

chgeo:
We came to the conclusion that it's best to remove the note altogether.
In the light of upcoming IAS, it's not stressing a third 'JWT' option.

PR #1387, opened 2024-10-29
Import from `cds` rather than `cds-dk` for `cds add` plugins

swaldmann:


swaldmann:
Tiny change, merging this :)

PR #1391, opened 2024-10-29
Fix long form --production option example in databases.md

qmacro:
Looks like the original was an en-dash (`–`) possibly converted from two original dashes (`--`) from some copy-pasta action starting in a WYSIWYG doc tool?


renejeglinsky (approved):
Thanks a lot!

PR #1419, opened 2024-11-05
doc(CDS): Update SQLite mapping for `cds.Map` to `JSON_TEXT`

bugwelle:
The next cds-compiler version will change the mapping.


renejeglinsky:
Is that included in a patch or the next minor?

agoerler (approved):

agoerler:
@stewsk - we don't need this as a quick fix. SQLite is typically used in-memory hence I am not concerned about persistent data. Nevertheless, it would be _nice to have_ this in a patch release.

stewsk (approved):

renejeglinsky (approved):

PR #1446, opened 2024-11-15
update audit-log presets

sjvans:
re https://github.com/cap-js/docs/pull/1443


guides/data-privacy/audit-logging.md
```diff
@@ -486,13 +490,12 @@ This provides an ultimate level of resiliency, plus additional benefits:
 
 - **False log messages are avoided** &mdash;  messages are forwarded to the audit log service on successfully committed requests; and skipped in case of rollbacks.
 
-This transparently applies to all implementations, even [custom implementations](#custom-implementation). You can opt out of this default by configuring outbox: false in the configuration, for example, as we do in the default configuration for development:
+This transparently applies to all implementations, even [custom implementations](#custom-implementation). You can opt out of this default by configuring outbox: false in the configuration, for example, as shown in the following snippet for profile `development`:
```
renejeglinsky, line 493: How about using this format instead of the code block?
image

renejeglinsky (approved):

grenik:
@renejeglinsky thanks for fixing it, but I think there are two open points left:

1. I think https://github.com/cap-js/docs/pull/1443#issuecomment-2478252616 from @sjvans regarding default outboxed was not reflected (it's not false anymore)

2. I thought `kind` is the new level of abstraction from the internal implementation, which is there exactly to avoid such errors (plugin impl changed > config doesn't work anymore > documentation must be changed / code throws errors)
See the audit logging plugin [impl](https://github.com/cap-js/audit-logging/blob/9bbb1b1f370e00e39559a3507c7cb091bcfff8af/package.json#L45)
This is why I suggested to replace `impl` with `kind` in https://github.com/cap-js/docs/pull/1443
Is my understanding wrong?

renejeglinsky:
Hi  @grenik 
for the latter: As this kind is the implicit default (as you can see from the source you linked or by executing `cds env requires.audit-log`), there's no need to list it here. That way, you'd not benefit from any changes done within the plugin. If you stick to the defaults, you benefit from all the changes there. At least that is my understanding. Maybe @sjvans can confirm or improve my answer.
Wrt your first point, thanks for catching this! I'll fix it.

sjvans:
i'm a little torn on this one. i don't really see the value in `[development]: { kind: 'audit-log-to-console' }` without also listing what this predefined kind contains. but if we also add the three kinds, then the snippet gets way too long. that's why i actually kind of like the current state, because it reflects the effective env. but it really is confusing for the readers...

sjvans:
```
{
  "audit-log": {
    "handle": ["READ", "WRITE"],
    "outbox": true,
    "[development]": {
      "kind": "audit-log-to-console"
    },
    "[hybrid]": {
      "kind": "audit-log-to-restv2"
    },
    "[production]": {
      "kind": "audit-log-to-restv2"
    }
  },
  "kinds": {
    "audit-log-to-console": {
      "impl": "@cap-js/audit-logging/srv/log2console"
    },
    "audit-log-to-restv2": {
      "impl": "@cap-js/audit-logging/srv/log2restv2",
      "vcap": {
        "label": "auditlog"
      }
    }
  }
}
```

PR #1451, opened 2024-11-19
Super minor typo fixes for new Aspect-oriented Modelling section

qmacro:
In addition, do you @danjoa (or @renejeglinsky ) think that it would help to emphasise that the third [Single table strategy](https://pages.github.tools.sap/cap/docs/cds/aspects#single-table-strategy) in the Inheritance Hierarchies section is the "recommended one" here?


renejeglinsky (approved):
@danjoa I'll bring DJ's question to our sync. 
@qmacro Someone will get back to you then :)

PR #1452, opened 2024-11-19
Native functions w/ diff DBs for CAP Java

rjayasinghe:
In order to use database dependent CDS models with CAP Java applications you need to consider some additional steps. Those are added to the database guide in this commit.


guides/databases.md
```diff
@@ -968,6 +968,54 @@ In case of conflicts, follow these steps to provide different models for differe
     }
    }}}
    ```
+<div markdown="1" class="impl java">
+
+3. For CAP Java setups you might need to reflect the different profiles in your CDS Maven plugin configuration. This might not be needed for all setups. For mixed local DB (sqlite, H2, or PostgreSQL) and hybrid HANA setup nothing special needs to be done since the local build defaults to the `development` profile. But for other setups like e.g. local PostgreSQL and SQLite you'll need two (profiled) `cds deploy` commands:
```
renejeglinsky, line 973: If I understand it correctly, this is about using two different DBs locally. If it's only one DB locally and one DB (HANA, Postgres) for production, then you can stay with the defaults, right?

```suggestion
3. For CAP Java setups you might need to reflect the different profiles in your CDS Maven plugin configuration. This might not be needed for all setups, like using a standard local database (sqlite, H2, or PostgreSQL) and a production SAP HANA setup. In that case the local build defaults to the `development` profile. But for other setups, like using a local PostgreSQL and a local SQLite you'll need two (profiled) `cds deploy` commands:
```

guides/databases.md
```diff
@@ -968,6 +968,54 @@ In case of conflicts, follow these steps to provide different models for differe
...
+    <execution>
+      <id>cds.build</id>
+      <goals>
+        <goal>cds</goal>
+      </goals>
+      <configuration>
+        <commands>
+          <command>build --for java</command>
+          <command>deploy --profile development --dry --out "${project.basedir}/src/main/resources/schema-h2.sql"</command>
+          <command>deploy --profile production --dry --out "${project.basedir}/src/main/resources/schema-postresql.sql"</command>
```
renejeglinsky, line 985: If my assumption (two local DBs) is correct, why is the production profile needed here?

guides/databases.md
```diff
@@ -968,6 +968,54 @@ In case of conflicts, follow these steps to provide different models for differe
...
+        </commands>
+      </configuration>
+    </execution>
+   ```
+
+4. For the Spring Boot side it's similar. If you have a local development DB and a hybrid profile with a remote HANA DB you only need to run in default (or any other) profile and you get the local DB and it's SQL schema that was written to your `src/main/resources`. For the HANA part the build and deploy part is done seperately and the application just needs to be started with a local `default-env.json` file and `cloud` profile or via `cds bind`.
```
renejeglinsky, line 991: I guess we should not promote the `default-env.json` case anymore when we have the more secure `cds bind`. But it's not started using cds bind, right? We need to rephrase that.
I removed some parts as I thought that is too detailed and probably not needed. But of course you can change my mind ;)

```suggestion
4. For the Spring Boot side it's similar. If you have a local development database and a hybrid profile with a remote SAP HANA database, you only need to run in default (or any other) profile. For the SAP HANA part, the build and deploy part is done separately and the application just needs to be started using `cds bind`.
```

renejeglinsky:
@rjayasinghe Could you have a look at my comments?

guides/databases.md
rjayasinghe: > If I understand it correctly, this is about using two different DBs locally. If it's only one DB locally and one DB (HANA, Postgres) for production, then you can stay with the defaults, right?

yes, exactly.

guides/databases.md
rjayasinghe: IIRC, for PostgreSQL this needed to be set explicitly. Without the production profile it did not work.

guides/databases.md
rjayasinghe: > I guess we should not promote the `default-env.json` case anymore when we have the more secure `cds bind`. But it's not started using cds bind, right? We need to rephrase that. I removed some parts as I thought that is too detailed and probably not needed. But of course you can change my mind ;)

You're right. We should not promote this anymore. Initially, I still liked it because it's easier to use it with run configurations of your IDE i.e. more flexibility. But recently I saw a neat trick how to combine `cds bind` and IDE run configurations for debugging: Instead of letting `cds bind` call some Maven plugin you can also use it to start the complete IDE. So, we can take your suggestion. 👍🏽

guides/databases.md
```diff
@@ -968,6 +968,54 @@ In case of conflicts, follow these steps to provide different models for differe
...
+          schema-locations: classpath:schema-postgresql.sql
+      datasource:
+        url: "jdbc:postgresql://localhost:5432/my_schema"
+        driver-class-name: org.postgresql.Driver
+        hikari:
+          maximum-pool-size: 1
+          max-lifetime: 0
+    ```
+  In case you use 2 different databases you also need to make sure that you have the JDBC drivers configured (on the classpath).
+</div>
```
renejeglinsky, line 1018:
```suggestion

</div>
```

rjayasinghe:
@renejeglinsky all checks green. Shall we merge? 🙃

renejeglinsky:
I know it might look a little unexpected, but when reading it locally, I felt that we need to make it clearer, that these steps is only apply to using two different local DBs. Here are my two suggestions:

1.
image

2.
image

WDYT?

rjayasinghe:
> I know it might look a little unexpected, but when reading it locally, I felt that we need to make it clearer, that these steps is only apply to using two different local DBs. Here are my two suggestions:
> ...
> 
> WDYT?

This is only a layout change, right? If yes, I would prefer the first one.

renejeglinsky (approved):

guides/databases.md
```diff
@@ -996,6 +996,59 @@ In case of conflicts, follow these steps to provide different models for differe
     }
    }}}
    ```
+<div markdown="1" class="impl java">
+
+::: info The following steps are only needed when you use two different local databases.
```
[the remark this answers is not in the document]
renejeglinsky:
```suggestion
:::info The following steps are only needed when you use two different local databases.
```

PR #1453, opened 2024-11-19
Add migration plugin-name to hdiconfig

Nils-Schiwek:
add plugin-name to `.hdiconfig` for migration

Without this, the migration will not work and deployment will show the following error:
`Error: "src/.hdiconfig": Configuration does not define a build plugin for file suffix "hdbmigrationtable"`

For me it was hard to find this information and I think it is missing in the documentation.

### DIFF
--- a/guides/databases-hana.md
+++ b/guides/databases-hana.md
@@ -449,6 +449,8 @@ Direct migration from _.hdbcds_ to _.hdbmigrationtable_ isn't supported by HDI. 
 [Learn more in the **Enhance Project Configuration for SAP HANA Cloud** section.](#configure-hana){.learn-more}
 
 During the transition from _.hdbtable_ to _.hdbmigrationtable_ you have to deploy version=1 of the _.hdbmigrationtable_ artifact, which must not include any migration steps.
+
+Ensure to add `"hdbmigrationtable": { "plugin_name": "com.sap.hana.di.table.migration" }` to your `.hdiconfig` file.
 :::
 
 HDI supports the _hdbcds → hdbtable → hdbmigrationtable_ migration flow without data loss. Even going back from _.hdbmigrationtable_ to _.hdbtable_ is possible. Keep in mind that you lose the migration history in this case.


renejeglinsky:
Thanks @Nils-Schiwek for opening this PR. I assigned a reviewer and will keep you posted.

LotharBender:
> Thanks @LotharBender ! In this guide we already mention `cds add hana`: https://cap.cloud.sap/docs/guides/databases-hana#configure-hana
> 
> Should we add a hint at the Schema Evolution?

I think this would be a good idea. To be honest, I was struggling, why Capire suggests to add `hana` for the `[hybrid]` profile instead of the `[production]` profile - this might totally confuse users...

Nils-Schiwek:
> On the other hand, the existing one should already contain the `.hdbmigrationtable` entry - do you have an idea why this isn't the case?

In our project, we did not yet have a migration, also no `@cds.persistence.journal`. It was a bit fiddly to build it in at this stage of the project. We are still on v7.9.5 so, that might also have some influence. I was impended by error messages on jenkins as my build was failing due to missing .hdiconfig entry for migration and it was hard to find that information. So I would like to improve doc with the information to keep in mind, that the .hdiconfig needs that information to run migrations.

Nils-Schiwek:
> > In our project, we did not yet have a migration, also no `@cds.persistence.journal`. It was a bit fiddly to build it in at this stage of the project. We are still on v7.9.5 so, that might also have some influence. I was impended by error messages on jenkins as my build was failing due to missing .hdiconfig entry for migration and it was hard to find that information. So I would like to improve doc with the information to keep in mind, that the .hdiconfig needs that information to run migrations.
> 
> Maybe we could add that as troubleshooting information in our troubleshooting section. Could you provide an error message we could attach this information to?

Yes, the logs from btp look like this:
```
Error: "src/secretEntity.hdbmigrationtable": could not create a compile unit for the file [8211714] (STDERR, APP/TASK/deploy)#

Error: "src/.hdiconfig": Configuration does not define a build plugin for file suffix "hdbmigrationtable" [8210015] (STDERR, APP/TASK/deploy)#

Error: Adding "src/secretEntity.hdbmigrationtable" for deploy... failed [8211700] (STDERR, APP/TASK/deploy)#
```

renejeglinsky:
Hi @Nils-Schiwek ,
I'm still struggling to understand what happened. Have you ever executed `cds add hana` in your project?
I mean even if my project is new, if I add hana, then I get an `db/src/.hdiconfig` file with `hdbmigrationtable` in it. And I think this has also been the case for cds7. So does it really play a role that your project hasn't had a migration yet?
To understand it helps to write a proper troubleshooting information, so thanks for taking the time to engage with us.
Best,
René

Nils-Schiwek:
Hey René,
unfortunately I am not sure, if we ever did a cds add hana command. I joined the project when the entities were already set and the repo was well established. So I am unable to tell. We did use cds-ts in the past, but since we did have some issues with it we now use cds again, maybe that was also part of the problem.

kind regards,
Nils

renejeglinsky:
Thanks @Nils-Schiwek !
In that case I'd really go for a short troubleshooting, so that users can find the error message and a hint. I'd not put it in the general guide as I'd say it's an edge case for now.
Will do that in an own PR and close yours then.

PR #1483, opened 2024-11-29
Auth explanation request

Nils-Schiwek:
This is unfinished text, so please do not merge.

When migrating to CDSv8 (from 7) we did have a few issues, especially on middleware related to the user.

Please update your documentation, it is missing these points currently afaik.

> We cannot (or should not) use `req.user` any more. Instead we use `cds.context.user` in the middleware now. 
> We need some explanation on the different users here. Today I had to debug and identify an anonymous user. This is not documented yet, so please elaborate here or somewhere else.

### DIFF
--- a/node.js/authentication.md
+++ b/node.js/authentication.md
@@ -18,6 +18,9 @@ uacp: This page is linked from the Help Portal at https://help.sap.com/products/
 [user]: #cds-user
 [`req.user`]: #cds-user
 
+afaik we cannot (or should not) use req.user any more. Instead we use cds.context.user in the middleware now. 
+We need some explanation on the different users here. Today I had to debug and identify an anonymous user. This is not documented yet, so please elaborte here or somewhere else.
+
 Represents the currently logged-in user as filled into [`req.user`](events#user) by authentication middlewares.
 Simply create instances of `cds.User` or of subclasses thereof in custom middlewares.
 For example:


sjvans:
hi @Nils-Schiwek 

thanks! some references where really outdated + i tried to clarify in https://github.com/cap-js/docs/pull/1493 does that make it clearer?

best,
sebastian

Nils-Schiwek:
Thanks @sjvans this looks a lot better! I just miss some pointers on the `cds.context.user.id` - afaik it can have the static ids `'anonymous'` and `'system'` (maybe more). In the past our middleware tried to check if the user object (or id) was empty, but I noticed this has changed, as since the update to cds v8 I get `cds.context.user.id = 'anonymous'` when before I got empty user object. Is this observation correct? 
Could you please add some pointers to the documentation describing the described "named" users `'anonymous'` and `'system'`?

sjvans:
hi @Nils-Schiwek 

the anonymous and privileged users should be checked via `instanceof cds.User.Anonymous/Privileged`. the matching strings are not sufficient.

system/ technical users, on the other hand, are identified via a [pseudo role](https://cap.cloud.sap/docs/guides/security/authorization#pseudo-roles) and should be checked via `cds.context.user.is('system-user')`.

best,
sebastian

sjvans:
+ for a privileged user, the `is(<role>)` check always returns true.

btw, why do you need to know if it is the anonymous user? non-anonymous users have pseudo role `authenticated-user`, so `user.is('authenticated-user')` should be sufficient.

Nils-Schiwek:
I tried to use cds.User.Anonymous in a test today, but I am having trouble.

In the middleware I check if user is anonymous with this code:
```typescript
    if (!cds.context?.user.is('authenticated-user')) {
      next();
      return;
    }
```
In the test I would expect to use something like 
```typescript
    it('handles request with anonymous user correctly', async () => {
      cds.context = { user: new cds.User('anonymous') } as EventContext;
      
      const nextMock = jest.fn();

      await initUser({} as XssecExpressRequest, {}, nextMock);

      expect(cds.context.user.id).toBe('anonymous');
      expect(nextMock).toHaveBeenCalledWith();
    });
```

It would be more convenient to use `cds.User.Anonymous` directly, but I was not able to. 
How do you expect us to mock a cds user? Or do I need to mock` user.is() `?

sjvans:
> I tried to use cds.User.Anonymous in a test today, but I am having trouble.
> 
> In the middleware I check if user is anonymous with this code:
> 
> ```ts
>     if (!cds.context?.user.is('authenticated-user')) {
>       next();
>       return;
>     }
> ```
> 
> In the test I would expect to use something like
> 
> ```ts
>     it('handles request with anonymous user correctly', async () => {
>       cds.context = { user: new cds.User('anonymous') } as EventContext;
>       
>       const nextMock = jest.fn();
> 
>       await initUser({} as XssecExpressRequest, {}, nextMock);
> 
>       expect(cds.context.user.id).toBe('anonymous');
>       expect(nextMock).toHaveBeenCalledWith();
>     });
> ```
> 
> It would be more convenient to use `cds.User.Anonymous` directly, but I was not able to. How do you expect us to mock a cds user? Or do I need to mock`user.is()`?

`new cds.User('anonymous')` creates a user that has the id `anonymous` but is an authenticated user nevertheless. the user's id is an opaque string to us, i.e., it carries no meaning. for a real anonymous user, do `user: new cds.User.Anonymous()` or `user: cds.User.anonymous` (the ready-to-use sealed instance).

Nils-Schiwek:
> `new cds.User('anonymous')` creates a user that has the id `anonymous` but is an authenticated user nevertheless. the user's id is an opaque string to us, i.e., it carries no meaning. for a real anonymous user, do `user: new cds.User.Anonymous()` or `user: cds.User.anonymous` (the ready-to-use sealed instance).

@sjvans is this a new feature? We are on cds 8.5.0 and I get `Property 'Anonymous' does not exist on type 'typeof User'`
It does not seem like the two samples you described above are available in typescript.

sjvans:
> @sjvans is this a new feature? We are on cds 8.5.0 and I get `Property 'Anonymous' does not exist on type 'typeof User'` It does not seem like the two samples you described above are available in typescript.

no. that's been there a long time. the sealed instances were added in v6.0.0, the classes are even a lot older.

cf. node repl:
```sh
Welcome to Node.js v22.8.0.
Type ".help" for more information.
> const cds = require('@sap/cds')
undefined
> cds.User
[Function: exports] {
  default: [Getter/Setter],
  anonymous: Anonymous {},
  Anonymous: [class Anonymous extends User],
  privileged: Privileged {},
  Privileged: [class Privileged extends User],
  _default: Anonymous {}
}
> cds.User.Anonymous
[class Anonymous extends User]
> cds.User.anonymous
Anonymous {}
```

@daogrady could you please assist re typescript?

daogrady:
Gladly! Feel free to just send me a meeting request so we can work this out.

Nils-Schiwek:
thanks for the help @sjvans  and @daogrady  - this really helped me out! Closing this PR as everything important is in #1493

PR #1493, opened 2024-12-03
req.user vs req.user

sjvans:
re https://github.com/cap-js/docs/pull/1483


node.js/authentication.md
```diff
@@ -43,6 +43,8 @@ const yetAnotherUser = new cds.User({id: user.id, roles: user.roles, attr: user.
 ```
 
 
+
+
```
sjvans, line 47:
```suggestion
```

node.js/events.md
```diff
@@ -137,6 +137,11 @@ The current user, an instance of `cds.User` as identified and verified by the au
 
 [See reference docs for `cds.User`.](authentication#cds-user){.learn-more .indent}
 
+::: tip
+Please note the difference between `req` in a service handler (instance of `cds.EventContext`) and `req` in an express middleware (instance of `http.IncomingMessage`).
+Case in point, `req.user` in a service handler is an official API and, if not explicitely set, points to `cds.context.user`.
+On the other hand, setting `req.user` in a custom authentication middleware is deprecated.
```
Nils-Schiwek, line 143: do you generally discourage the use use of `req.user` in middleware or only for setting properties?

smahati (approved):

node.js/events.md
johannes-vogel: We do recommend the usage of `<cds.Request>.user` or `cds.context.user`.

Our authentication middlewares only set `cds.context.user` which later on is also used when constructing a `cds.Request`.

`<express.Request>.user` is not set by us but if custom authentication middlewares set it, it's still working but deprecated.  Custom authentication middlewares should also set `cds.context.user` instead.

danjoa (approved):

PR #1495, opened 2024-12-04
Update typescript.md

czindel-sap:
Sorry I know this is not a proper text, but I wanted to highlight the misleading description

### DIFF
--- a/node.js/typescript.md
+++ b/node.js/typescript.md
@@ -124,7 +124,7 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
 
 ## TypeScript APIs in `@sap/cds`
 
-The package `@sap/cds` is shipped with TypeScript declarations. These declarations are used automatically when you write TypeScript files, but also enable IntelliSense and type checking for standard JavaScript development in Visual Studio Code.
+~The package `@sap/cds` is shipped with TypeScript declarations~ [See release notes](https://cap.cloud.sap/docs/releases/jun24#cds-types-dependency). These declarations are used automatically when you write TypeScript files, but also enable IntelliSense and type checking for standard JavaScript development in Visual Studio Code.
 
 Use them like this:
 
--- a/node.js/typescript.md
+++ b/node.js/typescript.md
@@ -122,11 +122,14 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
 
 [Learn more about integrating it into your build process.](../tools/cds-typer#integrate-into-your-build-process){.learn-more}
 
-## TypeScript APIs in `@sap/cds`
+## TypeScript APIs in `@sap/cds` <Since version="8.0.0" of="@sap/cds" />
 
-~The package `@sap/cds` is shipped with TypeScript declarations~ [See release notes](https://cap.cloud.sap/docs/releases/jun24#cds-types-dependency). These declarations are used automatically when you write TypeScript files, but also enable IntelliSense and type checking for standard JavaScript development in Visual Studio Code.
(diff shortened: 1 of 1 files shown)


daogrady (approved):

renejeglinsky:
I'm sorry but I'm not getting the point here. Do you want to point out, that this sentence is only valid since a certain release?

renejeglinsky:
I understand now, thanks!
@daogrady That means we need to adapt the whole section, right?

daogrady:
> I understand now, thanks! @daogrady That means we need to adapt the whole section, right?

I am not aware of any other places where it is claimed that the types are automatically installed

renejeglinsky:
So imports are still done through `@sap/cds`? The only thing one needs to do is `npm add @cap-js/cds-types`?

daogrady:
That is correct, yes. We outlined this as part of the [migration guide](https://cap.cloud.sap/docs/releases/jun24#cds-types-dependency).

node.js/typescript.md
```diff
@@ -124,7 +124,7 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
 
 ## TypeScript APIs in `@sap/cds`
 
-The package `@sap/cds` is shipped with TypeScript declarations. These declarations are used automatically when you write TypeScript files, but also enable IntelliSense and type checking for standard JavaScript development in Visual Studio Code.
+~The package `@sap/cds` is shipped with TypeScript declarations~ [See release notes](https://cap.cloud.sap/docs/releases/jun24#cds-types-dependency). These declarations are used automatically when you write TypeScript files, but also enable IntelliSense and type checking for standard JavaScript development in Visual Studio Code.
```
renejeglinsky, line 127: I'll make that code a codeblock later. That doesn't work in suggestions.

```suggestion
The package `@cap-js/cds-types` contains all TypeScript declarations for `@sap/cds` APIs. These declarations are used automatically when you write TypeScript files, but also enable IntelliSense and type checking for standard JavaScript development in Visual Studio Code. Just add the `@cap-js/cds-types` package to your project as follows:

`npm add @cap-ja/cds-types`

```

node.js/typescript.md
```diff
@@ -124,7 +124,7 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
...
 
 Use them like this:
```
renejeglinsky, line 129:
```suggestion
Use the Typescript declarations like this:
```

node.js/typescript.md
daogrady: Reads much better! I'd just like to add for consideration that cds7 actually still comes with type-batteries included. "Only" users of cds8 have to manually install cds-types. But I believe CAPire always targets the latest version anyway?

renejeglinsky: Yes, we target always the latest version. What would happen if a project on cds7 installs the cds-types package standalone?

We could add a badge to that section for clarity, like `<Since version="8.0.0" of="@sap/cds" />`. But I'd only like to do it if a cds7 user breaks his project using the direct dependency.

daogrady: The types users would see from explicitly installing cds-types at its latest version alongside a cds7 project would probably show the API for cds8. This isn't a problem for parts of cds that have not changed between v7 and v8. While it likely wouldn't _break_ their project, it could lead to confusion when users try to use cds8 API in their cds7 project, which then crashes on runtime, where that API is not actually present.

renejeglinsky: Alright, then let's do it :)

node.js/typescript.md
```diff
@@ -124,7 +124,7 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
 
 ## TypeScript APIs in `@sap/cds`
```
renejeglinsky, line 125:
```suggestion
## TypeScript APIs in `@sap/cds` <Since version="8.0.0" of="@sap/cds" />
```

renejeglinsky (approved):

node.js/typescript.md
daogrady: Awesome, thanks for making these amendments! 🙂

PR #1505, opened 2024-12-09
fix: audit logging for defined services only

sjvans:


johannes-vogel (approved):

guides/data-privacy/audit-logging.md
```diff
@@ -182,7 +182,7 @@ There are two options to access audit logs:
 
 ### Behind the Scenes...
 
-The generic audit logging implementation does the following:
+The generic audit logging implementation does the following for all [defined services](../providing-services#service-definitions):
```
renejeglinsky, line 185: I'm not sure if this isn't stating the obvious. but if we need that, then let's reorder it , so that the necessary condition is the first part:
```suggestion
For all [defined services](../providing-services#service-definitions), the generic audit logging implementation does the following:
```

renejeglinsky (approved):

PR #1513, opened 2024-12-12
move BTP-related docs to open source repo

renejeglinsky:


guides/extensibility/assets/customization-old.md
```diff
@@ -0,0 +1,249 @@
+---
```
chgeo, line 1: Let's leave the 'old' docs internal

guides/multitenancy/old-mtx-apis.md
```diff
@@ -0,0 +1,518 @@
+---
```
chgeo, line 1: Let's leave the 'old' docs internal

guides/multitenancy/old-mtx-migration.md
```diff
@@ -0,0 +1,632 @@
+---
```
chgeo, line 1: This is fine I'd say because the info there are not old, but only refer to old mtx.

java/multitenancy-classic.md
```diff
@@ -0,0 +1,950 @@
+---
```
chgeo, line 1: Let's leave the 'old' docs internal

guides/extensibility/assets/customization-old.md
renejeglinsky: Agreed to keep it here.

guides/multitenancy/old-mtx-apis.md
renejeglinsky: Agreed to keep it here.

java/multitenancy-classic.md
renejeglinsky: Agreed to keep it here.

renejeglinsky:
I guess we can ignore the failing build. The review step breaks often in the past days. Should we check that?

chgeo:
> I guess we can ignore the failing build. The review step breaks often in the past days. Should we check that?

Let's ignore that here I'd say

PR #1523, opened 2024-12-16
added troubleshooting for hana migration

renejeglinsky:
Closes https://github.com/cap-js/docs/pull/1453

### DIFF
--- a/get-started/troubleshooting.md
+++ b/get-started/troubleshooting.md
@@ -498,6 +498,15 @@ Options in [Saas Provisioning Service upgrade API](../guides/multitenancy/mtxs#e
 
 <div id="hana-ips" />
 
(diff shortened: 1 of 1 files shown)


LotharBender (approved):
I've again tried to reproduce the problem. My test showed that `cds build` is creating a `.hdiconfig` file in the `db/src` folder of your project if none exists.  This `.hdiconfig` file has an entry for the `hdbmigrationtable` build plugin.

Thus, the only scenario that could cause the described issue might be that your project already had a `.hdiconfig` file that didn't have this entry which is strange as it is included in the default one generated by `cds add hana`.

The troubleshooting hint can be seen as a hint for fixing an existing `.hdiconfig` file as  `cds add hana` can be used to add any missing build plugin entries (not only hdbmigrationtable) - not sure whether users have handcraftet versions of it.

PR #1524, opened 2024-12-16
Feedback button for Matomo

swaldmann:


chgeo (commented):
As discussed:
- Disable the Send button if there is no text.
- In the placeholder text add that a comment is optional.
- Add info about not sending personal information.

chgeo (approved):

PR #1546, opened 2025-01-08
`cds watch --include/--exclude`

swaldmann:


tools/cds-cli.md
```diff
@@ -428,6 +428,18 @@ To customize the diagram layout, use these settings in the _Cds > Preview_ categ
...
+
+Use `cds watch` to watch for changed files, restarting your server.
+
+Additional watched or ignored paths can be specified via CLI options:
+
+```sh
+cds watch --include ../other-app --exclude .stuff
+```
+
+Alternatively, you can add these paths through settings <Config keyOnly>cds.watch.include: ["../other-app"]</Config> and <Config keyOnly>cds.watch.exclude: [".stuff"]</Config> to your project configuration.
```
chgeo, line 441: I have used <Config> tool here instead of the lengthy config block.

Also, not use `.idea` as an example here as one could expect this is ignored by default.

chgeo (approved):

tools/cds-cli.md
```diff
@@ -428,6 +428,24 @@ To customize the diagram layout, use these settings in the _Cds > Preview_ categ
...
+::: details See the full help text of `cds watch`
+<!--@include: ./assets/help/cds-watch.out.md-->
+:::
+
+### Includes and Excludes <Since version="8.7.0" of="@sap/cds-dk" />
+
+Additional watched or ignored paths can be specified via CLI options:
+
+```sh
+cds watch --include ../other-app --exclude .stuff
```
renejeglinsky, line 444: Does that mean, thinking about our samples, that I can run bookstore, orders and review in one go?

renejeglinsky: What would be a use case for exclude, again using our samples as example?

chgeo: > Does that mean, thinking about our samples, that I can run bookstore, orders and review in one go?

No, you still need to run _one_ of these apps, but then can choose to have `cds watch` restart on changes in other (reused) modules.  In cap-samples, this is useful:

```sh
cd bookshop
cds watch --include ..
```

chgeo: > What would be a use case for exclude, again using our samples as example?

`exclude` is useful for folders that get changed e.g. by your IDE every now and then, like `.idea/` or `.vscode/`.  Maybe we should add this example back.
All in all, `exclude` is much rarer to be used, given that we exclude the usual stuff by default (like _node_modules/_, _gen/_ etc).

renejeglinsky: Ok, so still multiple terminals but when I've started it correctly I'd get the restart on changes from one terminal. Would it work if I use `cds watch` in all of the terminals or do I need to use `cds run` for the others?

chgeo: > Ok, so still multiple terminals but when I've started it correctly I'd get the restart on changes from one terminal.

This sounds like a different scenario where you want to have multiple parallel apps running (that potentially connect to each other).

This `include` feature is only about what could trigger a restart of one of these apps. By default it's a source code change in the app's folder, and with `--include` you can extend to more folders.

> Would it work if I use `cds watch` in all of the terminals or do I need to use `cds run` for the others?

This is also an orthogonal aspect: you use `watch` whenever you want to have the fast restart loop and `run` if you don't.

renejeglinsky: It seems that this is not sth you use in an easy getting started scenario. So it's fine if I don't fully understand it. Just tried to understand it and find an example based on our samples. But that's a different scenario, I understand that now.
Thanks @chgeo!

renejeglinsky (approved):

tools/cds-cli.md
chgeo: We can also have a talk on it if you want.

PR #1559, opened 2025-01-14
add a box with clarification for java changes and restart

rjayasinghe:


java/developing-applications/running.md
```diff
@@ -31,6 +31,12 @@ Once this is added, you can use the restart capabilities of the Spring Boot Devt
   * Artifacts generated from CDS (schema.sql, CSN, EDMX)
   * Any other static resource
 
+::: warning Restart for changed Java classes
+Sometimes, it happens that a (saved) change to a Java class does not result in an application restart even though you would have expected it. The reason is that the restart support for the Spring Boot Devtools only listens to changed .class files. So, a changed class only is detected as changed after it has been built.
```
bugwelle, line 35:
```suggestion
Sometimes, it happens that a (saved) change to a Java class does not result in an application restart even though you would have expected it. The reason is that the restart support for the Spring Boot Devtools only listens to changed `.class` files. So, a changed class only is detected as changed after it has been built.
```

java/developing-applications/running.md
```diff
@@ -31,6 +31,12 @@ Once this is added, you can use the restart capabilities of the Spring Boot Devt
   * Artifacts generated from CDS (schema.sql, CSN, EDMX)
   * Any other static resource
 
+::: warning Restart for changed Java classes
+Sometimes, it happens that a (saved) change to a Java class does not result in an application restart even though you would have expected it. The reason is that the restart support for the Spring Boot Devtools only listens to changed `.class` files. So, a changed class only is detected as changed after it has been built.
```
bugwelle, line 35:
```suggestion
Sometimes, it happens that a (saved) change to a Java class does not result in an application restart even though you would have expected it. The reason is that the restart support for the Spring Boot Devtools only listens to changed `.class` files. So, a changed class only is detected it has been built.
```

bugwelle (approved):

smahati (approved):

PR #1574, opened 2025-01-20
add cds repl and minor edits

renejeglinsky:


cds/cxn.md
```diff
@@ -137,17 +137,18 @@ _operator = string
 Examples:
 
 ```js
-cds.parse.expr(`x<9`)  ==//> returns:
+[dev] cds repl
+> cds.parse.expr(`x<9`)  ==//> returns:  // [!code focus]
```
chgeo, line 141: The code focus makes it worse for me as the result is blurred.  The result is equally important here as the input expression though.

Same for all the snippets below.

renejeglinsky: Ok. I took that idea from the [cds facade](https://pages.github.tools.sap/cap/docs/node.js/cds-facade#cds-home) docs. But it seems the response here is more important than over there. Will fix that.

renejeglinsky: Do we need the `//> returns` still if we have now more clarity through the `[dev] cds repl` hint?

cds/cxn.md
```diff
@@ -137,17 +137,18 @@ _operator = string
...
-cds.parse.expr(`x<9 and (y=1 or z=2)`)  ==//> returns:
+> cds.parse.expr(`x<9 and (y=1 or z=2)`)  ==//> returns: 
 {xpr:[
   {ref:['x']}, '<', {val:9}, 'and', {xpr:[
     {ref:['y']}, '=', {val:1}, 'or', {ref:['z']}, '=', {val:2}
   ]}
 ]}
 
-cds.parse.expr(`exists books[year = 2000]`)  ==//> returns:
+> cds.parse.expr(`exists books[year = 2000]`)  ==//>
```
renejeglinsky, line 151:
```suggestion
> cds.parse.expr(`exists books[year = 2000]`)  ==//> returns:
```

cds/cxn.md
chgeo: > Do we need the `//> returns` still if we have now more clarity through the `[dev] cds repl` hint?

No, I think we can omit it then.

PR #1595, opened 2025-01-24
Java Runtime views: Filtered Associations & Expressions

MattSchur:


java/working-with-cql/query-execution.md
```diff
@@ -434,7 +434,7 @@ WHERE B.STOCK < 10 AND A.NAME = ?
 ```
 
 ::: warning Limitations
-Runtime views are supported for simple [CDS projections](../../cds/cdl#as-projection-on). Constant values, expressions such as *case when* and [association filters](../../cds/cdl#publish-associations-with-filter) are currently ignored.
+Runtime views are supported for simple [CDS projections](../../cds/cdl#as-projection-on). Constant values and expressions such as *case when* are currently ignored. Expands of [filtered associations](../../cds/cdl#publish-associations-with-filter) are only supported since `3.7.0`.
```
agoerler, line 437: - Not sure we need this as the docu is valid for 3.7.0 only anyhow. On the other hand, there is no documentation for the LTS release. @renejeglinsky - what do you think.

> Constant values and expressions such as *case when* are currently ignored. 
@MattSchur - we also need to tackle this, I'm afraid.

renejeglinsky: We also have badges signaling a certain release which can be applied on header level. As we can't use a header here, I think doing this is perfectly fine. Sure, usually we go for "latest greatest" in the docs, but there can be circumstances where we want to avoid confusion. Probably after next major we can think about removing this version hint here.

agoerler (approved):

renejeglinsky (approved):

agoerler (approved):

MattSchur:
@smahati can we merge this PR?

PR #1611, opened 2025-01-31
chore: add limiation section for deep write

johannes-vogel:
@agoerler @beckermarc Please double check whether these limitations also apply to CAP Java.

@renejeglinsky Do we have a better pattern to signal good/bad?


guides/providing-services.md
```diff
@@ -298,9 +298,41 @@ DELETE .../Orders/1  -- would also delete all headers and items
 ```
 :::
 
+#### Limitations
 
+Note that deep `WRITE` operations are only supported out of the box if the following conditions are met:
 
+1. The on-condition of the composition only uses `=` operator
+2. Expressions are combined with the logical operator `AND`
```
MattSchur, line 306: and no negation `not(...)`

guides/providing-services.md
```diff
@@ -298,9 +298,41 @@ DELETE .../Orders/1  -- would also delete all headers and items
...
+3. The operands are references or `$self`
```
MattSchur, line 307: In Java we also support literals and pseudo vars, e.g. `$user.locale`

guides/providing-services.md
johannes-vogel: also for deep writes?

MattSchur: We don't support writing via on-conditions using negation.
`not(...)` can be rewritten using `!=`, which is covered by condition 1.

johannes-vogel: somehow I misread the initial comment and though you do support `not(...)`.
In essence, you also only support `AND`. `OR/NOT` are not supported generically?!.

We could emphasize this with 

```suggestion
2. The predicates are only connected with the logical operator `AND`
```

guides/providing-services.md
johannes-vogel: For a model like this, you automatically, you do set the locale automatically? Does that apply only to `locale` or to all pseudo variables?

```cds
entity X {
    key ID: Integer;
    ys: Composition of many Y on ys.locale = $user.locale
}
```

MattSchur: Yes, we automatically set the value for the pseudo var and support all vars.

guides/providing-services.md
MattSchur: That's correct, writing via on-conditions using `OR` or `NOT` is not supported.

guides/providing-services.md
johannes-vogel: bullet point is updated

guides/providing-services.md
johannes-vogel: I've changed the sentence now to above proposal. fine for you?

guides/providing-services.md
```diff
@@ -298,9 +298,41 @@ DELETE .../Orders/1  -- would also delete all headers and items
...
+1. The on-condition of the composition only uses comparison predicates with an `=` operator
+2. The predicates are only connected with the logical operator `AND`
+3. The operands are references or `$self`. CAP Java also supports pseudo variables like `$user.locale`
 
+Good:
+```cds
+entity Orders {
+  key ID : UUID;
+  title  : String;
+  Items  : Composition of many OrderItems on Items.order = $self;
```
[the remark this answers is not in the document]
johannes-vogel: If we don't make a distinction of good/bad, I do not think that we need a sample.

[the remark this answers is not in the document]
renejeglinsky: I mean it does show "Don't do that, do that". The only difference in both samples is this one line, or?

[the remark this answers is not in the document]
johannes-vogel: i think my comment was placed at the wrong place. I was referring to adrians comment above. We can do it in one block of course.

renejeglinsky:
Looks now like this:

image

renejeglinsky (approved):

PR #1612, opened 2025-01-31
update docs on how to access public change tracking interfaces

JenniferStamm:


java/change-tracking.md
```diff
@@ -279,19 +279,28 @@ If you change the values of the `OrderItems` entity directly via an OData reques
...
+@ServiceName(ChangeTrackingService.DEFAULT_NAME)
 public class ChangeTrackingHandler implements EventHandler {
 
-  @After(event = "createChanges")
-  void afterCreate(EventContext context) {
-    Result result = (Result) context.get("result");
-    result.listOf(Changes.class).forEach(c -> {
+  @After(event = ChangeTrackingService.CREATE_CHANGES)
+  void afterCreate(CreateChangesEventContext context) {
+    context.getChanges().forEach(c -> {
```
beckermarc, line 301: Are we sure context.getChanges is the same as `context.get("result")` earlier? I see `CreateChangesEventContext` also has  a `getResult()` method. @vmikhailenko Can you please take a look at this?

vmikhailenko: Isn't result the only correct way to handle something on after-phase? 

`getResult().listOf(Changes.class)` is the right way to do it.

JenniferStamm: I'll open another PR to fix it 🫡

PR #1614, opened 2025-02-03
Clarify semantics of locks wrt SELECT/READ

beckermarc:
There is a common misunderstanding that pessimistic locking, prevents plain select queries without locks.

See also HANA docs https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20fcf24075191014a89e9dc7b8408b26.html

See also PostgreSQL docs
https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS

See also H2 docs
https://www.h2database.com/html/commands.html#select


guides/providing-services.md
```diff
@@ -709,15 +709,22 @@ If the ETag validation detects a conflict, the request typically needs to be ret
 
 _Pessimistic locking_ allows you to lock the selected records so that other transactions are blocked from changing the records in any way.
 
-Use _exclusive_ locks when reading entity data with the _intention to update_ it in the same transaction and you want to prevent the data to be read or updated in a concurrent transaction.
+Use _exclusive_ locks when reading entity data with the _intention to update_ it in the same transaction and you want to prevent the data to be locked or updated in a concurrent transaction.
 
-Use _shared_ locks if you only need to prevent the entity data to be updated in a concurrent transaction, but don't want to block concurrent read operations.
+Use _shared_ locks if you only need to prevent the entity data to be updated in a concurrent transaction, but don't want to block concurrent locking read operations.
```
agoerler, line 714:
```suggestion
Use _shared_ locks if you only need to prevent the entity data to be locked exclusively by an update in a concurrent transaction or by a read operation with lock mode _exclusive_. Non-locking read operations or read operations with lock mode _shared_ are not prevented.
```

guides/providing-services.md
```diff
@@ -709,15 +709,22 @@ If the ETag validation detects a conflict, the request typically needs to be ret
...
+::: info
+It is allowed to read **locked** entities with a `Select` query that doesn't define a lock as well.
+This applies both to _exclusive_ and _shared_ locks.
+:::
+
 [Learn more about using the `SELECT ... FOR UPDATE` statement in the Node.js runtime.](../node.js/cds-ql#forupdate){.learn-more}
 
 [Learn more about using the `Select.lock()` method in the Java runtime.](../java/working-with-cql/query-api#write-lock){.learn-more}
+
+[]
```
agoerler, line 727: what's this?

agoerler (commented):
Maybe we should have a table

| state | select without lock | select with shared lock |  select with exclusive lock / update |
| --- | --- | --- |  --- | 
| not locked | passes -> not locked | passes -> shared lock | passes -> exclusive lock |
| shared lock | passes -> not locked | passes -> shared lock | passes -> exclusive lock |
| exclusive lock | passes -> not locked | passes -> exclusive lock | waits -> exclusive lock |

beckermarc:
> Maybe we should have a table

Maybe, but I don't think your table is correct, is it?
 
- "shared lock" / "select with exclusive lock / update" should be "waits"
- "exclusive lock" / "select with shared lock" should be "waits"

agoerler:
> > Maybe we should have a table
> 
> Maybe, but I don't think your table is correct, is it?
> 
> * "shared lock" / "select with exclusive lock / update" should be "waits"

Yes, you are right. A shared lock can be promoted to an exclusive lock. But I think only in the transaction where the lock was originally created, I think. Others will wait.

> * "exclusive lock" / "select with shared lock" should be "waits"

Yes, you are right.

johannes-vogel:
Should we also link the db documentation?

beckermarc:
The db docs are also rather cryptic sometimes and not so clear in my opinion. Therefore not sure if that provides a lot of additional value. Do you think they would be helpful after taking a look at them?

guides/providing-services.md
```diff
@@ -709,15 +709,21 @@ If the ETag validation detects a conflict, the request typically needs to be ret
 
 _Pessimistic locking_ allows you to lock the selected records so that other transactions are blocked from changing the records in any way.
 
-Use _exclusive_ locks when reading entity data with the _intention to update_ it in the same transaction and you want to prevent the data to be read or updated in a concurrent transaction.
+Use _exclusive_ locks when reading entity data with the _intention to update_ it in the same transaction and you want to prevent the data to be locked or updated in a concurrent transaction.
 
-Use _shared_ locks if you only need to prevent the entity data to be updated in a concurrent transaction, but don't want to block concurrent read operations.
+Use _shared_ locks if you only need to prevent the entity data to be locked exclusively by an update in a concurrent transaction or by a read operation with lock mode _exclusive_. Non-locking read operations or read operations with lock mode _shared_ are not prevented.
 
 The records are locked until the end of the transaction by commit or rollback statement.
```
renejeglinsky, line 717:
```suggestion
Here's an overview table:

| State | Select Without Lock | Select With Shared Lock |  Select With Exclusive Lock/Update |
| --- | --- | --- |  --- | 
| not locked | passes -> not locked | passes -> shared lock | passes -> exclusive lock |
| shared lock | passes -> not locked | passes -> shared lock | waits -> exclusive lock |
| exclusive lock | passes -> not locked | waits -> exclusive lock | waits -> exclusive lock |

```

guides/providing-services.md
```diff
@@ -709,15 +709,28 @@ If the ETag validation detects a conflict, the request typically needs to be ret
...
+Here's an overview table:
+
+| State | Select Without Lock | Select With Shared Lock |  Select With Exclusive Lock/Update |
+| --- | --- | --- |  --- | 
+| not locked | passes -> not locked | passes -> shared lock | passes -> exclusive lock |
+| shared lock | passes -> not locked | passes -> shared lock | waits -> exclusive lock |
+| exclusive lock | passes -> not locked | waits -> exclusive lock | waits -> exclusive lock |
```
renejeglinsky, line 723:
```suggestion

Here's an overview table:

| State              | Select Without Lock   | Select With Shared Lock |  Select With Exclusive Lock/Update |
| --------------- | ----------------------- | -------------------------- |  ------------------------------------- | 
| not locked      | passes -> not locked | passes -> shared lock    | passes -> exclusive lock                  |
| shared lock     | passes -> not locked | passes -> shared lock    | waits -> exclusive lock                     |
| exclusive lock | passes -> not locked | waits -> exclusive lock   | waits -> exclusive lock                     |
```

guides/providing-services.md
```diff
@@ -709,15 +709,29 @@ If the ETag validation detects a conflict, the request typically needs to be ret
...
+| State              | Select Without Lock   | Select With Shared Lock |  Select With Exclusive Lock/Update |
+| --------------- | ----------------------- | -------------------------- |  ------------------------------------- | 
+| not locked      | passes -> not locked | passes -> shared lock    | passes -> exclusive lock                  |
+| shared lock     | passes -> not locked | passes -> shared lock    | waits -> exclusive lock                     |
+| exclusive lock | passes -> not locked | waits -> exclusive lock   | waits -> exclusive lock                     |
+
+::: info
+It is allowed to read **locked** entities with a `Select` query that doesn't define a lock as well.
+This applies both to _exclusive_ and _shared_ locks.
+:::
```
renejeglinsky, line 729: Can this then be removed?
```suggestion
```

guides/providing-services.md
```diff
@@ -709,15 +709,25 @@ If the ETag validation detects a conflict, the request typically needs to be ret
...
 
 The records are locked until the end of the transaction by commit or rollback statement.
 
+Here's an overview table:
+
+| State              | Select Without Lock   | Select With Shared Lock |  Select With Exclusive Lock/Update |
+| --------------- | ----------------------- | -------------------------- |  ------------------------------------- | 
+| not locked      | passes -> not locked | passes -> shared lock    | passes -> exclusive lock                  |
+| shared lock     | passes -> not locked | passes -> shared lock    | waits -> exclusive lock                     |
+| exclusive lock | passes -> not locked | waits -> exclusive lock   | waits -> exclusive lock                     |
```
beckermarc, line 724: To me it is not clear what the `->` in this table means. Does it mean "and this will be the new state"? If yes I think the table is not correct. In my opinion we should remove those hints completely and only keep "passes" or "waits".
@agoerler Can you clarify?

agoerler: The intent of `->` was to indicate the target state after execution of the select. I think the entry exclusive lock/select with shared lock is wrong as the select would wait until the exclusive lock is released and hence the traget state should be "shared lock". Also select without always leaves the lock state untouched. 

I agree that we should remove the `->`.

renejeglinsky (approved):

PR #1617, opened 2025-02-04
Troubleshooting: add header with error message to existing entry

renejeglinsky:
Closes: https://github.com/cap-js/docs/issues/1583

As we do have already a solution to this new problem, I wanted to add a header with the error message to the existing content. Similar to:
image

But when trying that, it didn't work. I added the method class (`{.method}`) just to test, and found that this works. Using any other class, like `{.class}` or `{.annotation}` didn't work.

@swaldmann @chgeo Do you have any idea why?


chgeo:
I don't get it → let's talk about it.

PR #1628, opened 2025-02-06
Added prerequisites for remote services in node

mariayord:


mariayord:
@renejeglinsky The package installation of sdks is needed not only for the queries, but it is required at the beginning for the initialization of the remote service.

renejeglinsky:
> @renejeglinsky The package installation of sdks is needed not only for the queries, but it is required at the beginning for the initialization of the remote service.

Understood. But where in this guide is the first time you really need it? Definitely not before getting an EDMX to import the API, or? If we should decide to start a prerequisite section, which undermines our grow as you go approach, it should then be complete. So that anyone just looking for prerequisites could do anything wrt remote service consumption w/o having a look at further sections.

mariayord:
@renejeglinsky 'Get and Import an External Service API' is a part of the remote service scenario. But the first thing that the user has to do to create or to connect to remote service is to have the installed dependencies. @David-Kunz could you please comment?

renejeglinsky:
> > So that anyone just looking for prerequisites could do anything wrt remote service consumption w/o having a look at further sections.
> 
> @renejeglinsky you mean also the part with the 'Get and Import an External Service API' should be in prerequisites ?

I mean that if we have a prerequisite section, then it should cover all prerequisites for the whole guide.
AFAIK we only have prerequisites in capire for thing like "Setup your HANA" that aren't directly CAP. Usually, if it's only a command, we let the user execute this command when the user needs it. So I still think, in this example here, we should not do a prerequisites section. Instead we should add the "Install Dependencies" step whenever it is needed (or cross reference within the guide).

mariayord:
@renejeglinsky Is it OK to rename 'Prerequisites' with 'Install Dependencies'? I think it doesn't matter if it is before or after 'Get and Import an External Service API' as these are not steps for some scenario.

renejeglinsky:
> @renejeglinsky Is it OK to rename 'Prerequisites' with 'Install Dependencies'? I think it doesn't matter if it is before or after 'Get and Import an External Service API' as these are not steps for some scenario.

Hi @mariayord , just renaming it from Prerequisites to Install Dependencies would not solve the issue I see. But the idea in general is good. So let me give it a try and I'll commit a proposal for you here. Ok?

renejeglinsky:
So let me explain my proposal:

1. Add the step as an "Install Dependencies" section (so that we can link to that later)
image

2. Refer to this step from other places in one of these two ways:
-> collapsed
image

-> expanded
image


What do you think overall? Which version do you like better the "Learn more" style link or the box that can be expanded?

guides/using-services.md
```diff
@@ -387,6 +388,16 @@ mv API_BUSINESS_PARTNER-new.cds API_BUSINESS_PARTNER-orig.cds
 
 To prevent accidental loss of modifications, the `cds import --as cds` command refuses to overwrite modified files based on a "checksum" that is included in the file.
 
+### Install Dependencies { .node }
+
+In the next section, you're going to mock a remote service as OData service. For this you need to install the required packages first:
```
mariayord, line 393: @renejeglinsky These dependencies are needed not only for the mock scenario,  they are needed for all remote service scenarios.

renejeglinsky:
```suggestion
In the next section, you're going to mock a remote service as OData service. To make this and all other remote service interaction possible, you need to install the required packages first:
```

David-Kunz: It's independent of mocking, you always need these packages.

renejeglinsky: The challenge is to find the best place for this step. Usually (grow as you go) we make the user execute such things when they first need it. In this guide, if you go from top to bottom, that would be this place, right? If we make it a step like a prerequisite, you'd install the dependencies and then? Why do it if you not need it immediately? Anyway the referencing part is much more important so that people jumping into the guide and not reading from top to bottom can find that information as well.

David-Kunz: Installing the dependencies is very important, otherwise you cannot consume remote services. If we bury it within some minor step, people will overlook it and create issues on our side (happened before many times).

guides/using-services.md
```diff
@@ -84,6 +84,15 @@ The user picks a supplier from the list. That list is coming [from the remote sy
 
 It should be also possible to search for suppliers and show the associated risks by extending the remote supplier service [with the local risk service](#extend-a-remote-by-a-local-service) and its risks.
 
+## Prerequisites { .node }
```
renejeglinsky, line 87:
```suggestion
## Install Dependencies { .node }
```

guides/using-services.md
```diff
@@ -84,6 +84,15 @@ The user picks a supplier from the list. That list is coming [from the remote sy
...
+
+First install the required packages:
+
```
renejeglinsky, line 90:
```suggestion
```

renejeglinsky (approved):

PR #1652, opened 2025-02-18
Aggregation.default is mandatory for custom aggregate

HeikoTheissen:
As discussed with @agoerler.

### DIFF
--- a/advanced/odata.md
+++ b/advanced/odata.md
@@ -1187,6 +1187,7 @@ GET /Books?$apply=aggregate(stock with sum as stock) HTTP/1.1
 #### Currencies and Units of Measure
 
 If a property represents a monetary amount, it may have a related property that indicates the amount's *currency code*. Analogously, a property representing a measured quantity can be related to a *unit of measure*. To indicate that a property is a currency code or a unit of measure it can be annotated with the [Semantics Annotations](https://help.sap.com/docs/SAP_NETWEAVER_750/cc0c305d2fab47bd808adcad3ca7ee9d/fbcd3a59a94148f6adad80b9c97304ff.html) `@Semantics.currencyCode` or `@Semantics.unitOfMeasure`.
+The aggregation method (typically, sum) is specified with the `@Aggregation.default` annotation.
 
 ```cds
 @Aggregation.CustomAggregate#amount   : 'Edm.Decimal'
@@ -1195,6 +1196,7 @@ entity Sales {
     key id        : GUID;
         productId : GUID;
         @Semantics.amount.currencyCode: 'currency'
+        @Aggregation.default: #SUM
         amount    : Decimal(10,2);
         @Semantics.currencyCode
         currency  : String(3);


agoerler (approved):

advanced/odata.md
```diff
@@ -1187,6 +1187,7 @@ GET /Books?$apply=aggregate(stock with sum as stock) HTTP/1.1
 #### Currencies and Units of Measure
 
 If a property represents a monetary amount, it may have a related property that indicates the amount's *currency code*. Analogously, a property representing a measured quantity can be related to a *unit of measure*. To indicate that a property is a currency code or a unit of measure it can be annotated with the [Semantics Annotations](https://help.sap.com/docs/SAP_NETWEAVER_750/cc0c305d2fab47bd808adcad3ca7ee9d/fbcd3a59a94148f6adad80b9c97304ff.html) `@Semantics.currencyCode` or `@Semantics.unitOfMeasure`.
```
renejeglinsky, line 1189: The documentation behind this link is not maintained anymore. We should replace the link with one of the options (at least that's what I found):

```suggestion
If a property represents a monetary amount, it may have a related property that indicates the amount's *currency code*. Analogously, a property representing a measured quantity can be related to a *unit of measure*. To indicate that a property is a currency code or a unit of measure it can be annotated with the [Semantics Annotations](https://help.sap.com/docs/ABAP_PLATFORM_NEW/fc4c71aa50014fd1b43721701471913d/3ccd5cf7b98248618b1013a4a296e415.html) `@Semantics.currencyCode` or `@Semantics.unitOfMeasure`.
```

or

```suggestion
If a property represents a monetary amount, it may have a related property that indicates the amount's *currency code*. Analogously, a property representing a measured quantity can be related to a *unit of measure*. To indicate that a property is a currency code or a unit of measure it can be annotated with the [Semantics Annotations](https://help.sap.com/docs/SAP_HANA_PLATFORM/691cb949c1034198800afde3e5be6570/57f00f85b92c4ee9a8b069f212b59573.html?locale=en-US&version=2.0.08&) `@Semantics.currencyCode` or `@Semantics.unitOfMeasure`.
```

agoerler: I honestly don't know where the `@Semantics` annotations are specified.

HeikoTheissen: https://github.com/SAP-samples/abap-platform-fiori-feature-showcase/wiki/Feature-Showcase-App-Guide#adding-currency-or-uom-fields-to-a-table

agoerler: `@Semantics.currencyCode` declares that the annotated element is a currency

`@Semantics.amount.currencyCode` relates an annotated element, the amount, to the element that holds the currency code for this amount

renejeglinsky: Well, none of the pages, except the outdated Netweaver docs, talks about `@Semantics.currencyCode`. How do we approach this then?

agoerler: I have also found this page: https://help.sap.com/doc/saphelp_nw75/7.5.5/en-US/fb/cd3a59a94148f6adad80b9c97304ff/frameset.htm

renejeglinsky: This is an old doc set as well :(

agoerler: These annotations determine how the backend (i.e. CAP) should aggregate values. They are no contract between the frontend and the UI. Maybe this is the reason why we can't find a backend-agnostic documentation. So we either link to the documentation of a different backen (RAP) or we duplicate the documentation here. Both is fine with me. But actually it would be cleaner to duplicate the doc for CAP.

renejeglinsky: @agoerler If there is no perfect fit, I think it's fair to think about documenting it ourselves. But that would be some effort, right? Who should be involved? I guess it's not just about duplicating the RAP docs, or?

agoerler: @HeikoTheissen 

What about 

[@Meassure.ISOCurrency](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Measures.V1.xml#L70) ?

Would this be a standard conform way to express the same intent?

agoerler: I thought as "replacement" for `@Semantics.amount.currencyCode`, which targets an amount and references the currency property.

renejeglinsky: We'll make it a separate activity to provide the documentation and just merge this PR as is, ok?

PR #1659, opened 2025-02-20
Node: Advanced Event Mesh

David-Kunz:


sjvans:
see plugin readme and https://github.com/cap-js/docs/pull/1871

PR #1663, opened 2025-02-24
Update link to ORD Specification

pavelkornev:


renejeglinsky (approved):
Thanks for providing this fix!

PR #1667, opened 2025-02-25
Node: Document Database Service

BobdenOs:


node.js/databases.md
```diff
@@ -12,18 +12,56 @@ status: released
...
 
-## cds.**DatabaseService**  <i>  class </i> { #cds-db-service}
+## cds.**DatabaseService**  <i>  class </i> {#cds-db-service }
 
 ### class cds.**DatabaseService**  <i>  extends cds.Service </i>
 
 
 
-
 ### <span style="color:grey">srv</span>.begin <i> () → this </i> {#db-begin }
```
renejeglinsky, line 21: I'm not sure, but maybe we should adapt this "new style" as shown here:
- Source: https://github.com/cap-js/docs/blob/49eae8f28a7f5206225b6c97da32ac23d7252706/node.js/core-services.md?plain=1#L427
- capire: https://cap.cloud.sap/docs/node.js/core-services#srv-init

node.js/databases.md
```diff
@@ -12,18 +12,56 @@ status: released
...
+
+Uses the current DatabaseService model to `infer` a provided `cds.ql.Query`. It is best to use `cds.infer` instead. (deprecated?)
+
+### <span style="color:grey">srv</span>.set <i> () → this </i> {#db-set }
+
+Sets the provided key value pairs as variables for the current connection. The values can be access by using the `session_context('<key>')` function inside any Queries. Is called by [begin](#db-begin) automatically.
+
+### <span style="color:grey">srv</span>.run <i> () → this </i> {#db-run }
+
+Runs the provided `cds.ql.Query` using the current DatabaseService.
```
sjvans, line 63: what about plain queries? we should mention that they are supported as fallback option, but should always use "for security reasons" prepared statement syntax, i.e., `run('INSERT into T values (?,?)', 1, 'alice')`

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
 
 This documentation is not complete yet, or the APIs are not released for general availability.
 Stay tuned to upcoming releases for further updates.
+
+
+## SAP HANA — Known Limitations {#hana-limitations}
+
+The following limitations apply when running on SAP HANA. Some apply regardless of which client driver is used, others depend on the chosen driver: [`hdb`](https://www.npmjs.com/package/hdb), [`@sap/hana-client`](https://www.npmjs.com/package/@sap/hana-client), or [`@cap-js/hana`](https://www.npmjs.com/package/@cap-js/hana).
```
BobdenOs: > or [`@cap-js/hana`](https://www.npmjs.com/package/@cap-js/hana).

Is not a driver.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+Please report additional limitations you encounter at [cap-js/cds-dbs](https://github.com/cap-js/cds-dbs/issues).
```
BobdenOs: Lets not promote this here.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+### General {#hana-limitations-general}
+
+These limitations apply to SAP HANA independently of the chosen driver.
+
+#### Virtual elements in expressions
+
+Virtual elements (`@cds.virtual`/`virtual`) **must not appear** in `WHERE`, `SELECT`, `ORDER BY`, or any other expression. Such queries are rejected with *"Virtual elements are not allowed in expressions"*.
+
+The runtime cannot transparently replace virtual conditions with `1 = 1` because it does not have enough information about expression boundaries (`xpr`).
```
BobdenOs: Not a HANA limitation.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `BLOB` columns cannot have default values
+
+SAP HANA does not support default values on `BLOB` types. Provide values explicitly on `INSERT`/`UPSERT` instead.
```
BobdenOs: Not a HANA limitation.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### Date/time range and formatting
+
+- `DATE`, `SECONDDATE`, and `TIMESTAMP` reject years ≤ 0 and years ≥ 10000.
+- The upper bound is `9999-12-31 23:59:59` — `24:00:00` is **not** accepted (it is on some other databases).
+- Truncated forms are not auto-expanded; date components must be fully specified when a time is included.
```
BobdenOs: Yes that is the case, but I would definitely not see this as a limitation that WE should document. This follows the ISO standards...

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `DateTime` / timestamp precision
+
+When using the [`hdb`](#hana-limitations-hdb) driver, only millisecond precision (3 fractional digits) is supported. See the [driver-specific section](#hana-limitations-hdb) below.
```
BobdenOs: This isn't really true. `@cap-js/hana` is capable of using 7 fractional precision timestamps just fine even when using `hdb`. The 3 fractional limitations is MORE a CAP limitation then a `@cap-js/hana` limitation.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `UNION`-based queries
+
+Queries with a top-level `UNION` (`SELECT.SET.op === 'union'`) are not supported.
```
BobdenOs: Not a HANA limitation. This is a `@cap-js/db-service` specific decided not supported feature.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `HIERARCHY`, `HIERARCHY_DESCENDANTS`, and `HIERARCHY_ANCESTORS`
+
+These HANA functions are currently not exposed via the unified CQL function framework.
```
BobdenOs: This is factionally false. These are directly being used through `cqn` inside the `SELECT_recurse` function.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `Binary` input restrictions
+
+Binary values are converted with `HEXTOBIN(...)` and therefore only work when the input is a string literal or a `NEW.<field>` reference. Arbitrary binary expressions cannot be embedded in DML.
```
BobdenOs: Purely internals. Doesn't impact the actual public APIs.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `Int64` and `Decimal` are returned as strings
+
+To prevent precision loss in JavaScript, `cds.Int64` and `cds.Decimal` values are returned as strings. Custom handlers must not assume `Number`. See the [Data structure](#databaseservice-data-structure) section above.
```
BobdenOs: Not a HANA limitation. This literally describes a feature.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### Streaming binary content uses `NCLOB`
+
+`NVARCHAR` parameters quickly hit size limits when streaming. The runtime therefore interprets the parameter as `NCLOB` so that values up to 2 GB per row can be streamed.
```
BobdenOs: What is this even on about ?

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### HANA Express (version ≤ 2)
+
+Older HANA Express versions cannot process large JSON documents (limit between 64 KB and 128 KB). The runtime falls back to inserting entries one at a time on such servers.
```
BobdenOs: As stated in the sentence. This is a limitation that is solved by `@cap-js/hana`. Which means it is not a limitation for the public APIs.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `not not` requires HANA ≥ 4
+
+Double-negation (`not not <expr>`) is not supported on HANA servers older than version 4.
```
BobdenOs: Sure, but officially HANA 2 is not supported by `@cap-js/hana` anyway.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `JSON_TABLE` does not produce `BOOLEAN`
+
+When working at SQL level: `JSON_TABLE` returns booleans via `NVARCHAR` + `CASE` workaround.
```
BobdenOs: Also doesn't impact the public API.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### Filtering on structured / JSON-array-valued elements
+
+Filters that compare against a structured element whose value is a JSON array return no results on HANA.
```
BobdenOs: Doesn't work on any database service.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `groupBy` with same-named fields from different associations
+
+A `groupBy` that includes elements with identical names from different paths (e.g. `author.ID` and `ID`) does not produce the expected result. See [cap/cdsnode#2366](https://github.tools.sap/cap/cdsnode/issues/2366).
```
BobdenOs: Who is going to look at the **internal** issue ? This is `github.com` you are not allowed to post internal urls.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `any()` over unmanaged-association navigation *(on `@cap-js/hana`)*
+
+OData `$filter` expressions using `any()` that navigate through an **unmanaged** association are currently not supported by the new `@cap-js/hana` driver. The legacy `hdb`-based stack supports this case.
```
BobdenOs: Applies to all `@cap-js/db-service`s and calling `legacy` as `hdb` based stack is very misleading. It is supposed to be called `@sap/cds-hana`.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+---
+
+### `hdb` driver {#hana-limitations-hdb}
+
+The following limitations are specific to the [`hdb`](https://www.npmjs.com/package/hdb) driver.
+
+#### Timestamp precision is limited to 3 fractional digits
+
+`hdb` does not yet support timestamps with more than 3 fractional digits of second precision (milliseconds, not microseconds). See [cap/cdsnode#2368](https://github.tools.sap/cap/cdsnode/issues/2368).
```
BobdenOs: Duplicate from above. Also doesn't really apply to the `@cap-js/hana`. It better applies to `@sap/cds` in general.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `$filter` with binary literals fails
+
+Queries such as `?$filter=binary eq binary'AB12…'` fail with `"Cannot set parameter at row: 1. Wrong input for BINARY type"`. See [cap/cdsnode#2357](https://github.tools.sap/cap/cdsnode/issues/2357).
```
BobdenOs: internal link. and the link provides no proof of this still being the case.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### Large strings may be returned as streams
+
+`hdb` occasionally returns large `NCLOB`/`NVARCHAR` values as `Readable` streams instead of plain strings. Consumer code that reads such values directly must be prepared to drain a stream.
```
BobdenOs: Problem of `hdb` that doesn't apply to `@cap-js/hana`.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+#### Drafts: `cds.LargeBinary` is not auto-copied from draft to active
+
+When activating a draft (`draftActivate`), unchanged LOB columns (`cds.LargeBinary`) are not automatically copied from the draft table to the active table.
+
+To opt-in to the database-side move (`UPDATE ... FROM SELECT`), enable the feature flag:
+
+```jsonc
+// package.json
+"cds": {
+  "features": { /* … */ },
```
BobdenOs: Why is `features` here ?

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+Queries such as `?$filter=binary eq binary'AB12…'` fail with `"Cannot set parameter at row: 1. Wrong input for BINARY type"`. See [cap/cdsnode#2357](https://github.tools.sap/cap/cdsnode/issues/2357).
+
+#### Large strings may be returned as streams
+
+`hdb` occasionally returns large `NCLOB`/`NVARCHAR` values as `Readable` streams instead of plain strings. Consumer code that reads such values directly must be prepared to drain a stream.
+
+#### Drafts: `cds.LargeBinary` is not auto-copied from draft to active
+
+When activating a draft (`draftActivate`), unchanged LOB columns (`cds.LargeBinary`) are not automatically copied from the draft table to the active table.
```
BobdenOs: That is the opposite of what actually happens. When an active is being made into a draft. The Blobs are not copied into the draft. So once the draft is activated again the non existent blobs will be skipped. Also this describes the opposite of the root cause of the `hdb` issue.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+```jsonc
+// package.json
+"cds": {
+  "features": { /* … */ },
+  "fiori": { "move_media_data_in_db": true }
+}
+```
+
+The legacy flag `cds.features.binary_draft_compat` restores the pre-optimization behavior (every BLOB copied into the draft entity).
```
BobdenOs: As mentioned before this feature flag is not related to the root cause. So it has to not be mentioned here it will only cause confusion.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### Draft + binary streaming combined is not supported
+
+End-to-end tests for streamed binaries inside drafts are currently excluded on the `hdb` driver.
```
BobdenOs: This is a duplicate of the previous issue. Not sure why customers would care that our `hdb` tests are disabled for a streaming scenario that doesn't work. I don't think most CAP customers are going to run out `cds-dbs` test suite.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+---
+
+### `@sap/hana-client` driver {#hana-limitations-hana-client}
+
+The following limitations are specific to the [`@sap/hana-client`](https://www.npmjs.com/package/@sap/hana-client) driver.
+
+#### `communicationTimeout` must be set explicitly
+
+Without an explicit `communicationTimeout` (in milliseconds), queries can hang for over ten minutes. The runtime sets `60000` by default — applications can override via pool/connection configuration.
```
BobdenOs: "if not set" ... "it is set" ... so it is not a problem.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `LOB`s cannot be streamed deferred
+
+The `ResultSet` API of `@sap/hana-client` only allows `getData()` on the current row and is forward-only. As a consequence, all LOB content of a result row is buffered in memory before processing continues — true deferred streaming is not possible with this driver.
```
BobdenOs: The start is true, because of a lack of imagination the secondary conclusion is completely wrong. In the case of `@cap-js/hana` the streams are NOT loaded into memory. Instead the `ResultSet` is `reset` and walked to the correct `row` to ensure that the current `LOB` stream is the active `row` of the `ResultSet`. Which allows for real streaming and out of order streaming, but comes with the overhead of the amount of rows that have to be walked from the current row to the target row. Which for most scenarios is either `1` or `0`.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### Result-set size ceiling (~4 GB)
+
+A consequence of the above limitation: result sets are capped at approximately `0xFFFFFFFB` bytes (~4 GB).
```
BobdenOs: As a consequence of the above statement being a lie. This one is also false.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### `HanaLobStream` cannot be consumed with `for await`
+
+The `HanaLobStream` returned by `@sap/hana-client` hangs when iterated with `for await ... of`. The runtime detects this case (via `result[key].constructor.name === 'HanaLobStream'`) and uses a `PassThrough`-based reader instead.
```
BobdenOs: The `@sap/hana-client` itself doesn't expose a `LOB` stream. Instead it exposes an `getData` function which functions like an `read` function of a node stream. So while `PassThrough` is used it is only used for `hdb`. It is not even imported in the `hana-client` driver wrapper.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+#### APM decorators (e.g. Dynatrace) cannot wrap internal methods
+
+`@sap/hana-client` does not allow wrapping/decorating `client.prepare()` and other internals. APM integrations that rely on method wrapping have reduced visibility on this driver.
+
+---
+
+### Reporting additional limitations
+
+If you hit a HANA-specific issue not listed here, please open an issue at [cap-js/cds-dbs](https://github.com/cap-js/cds-dbs/issues) (for new-stack issues) or at the [internal `cdsnode` tracker](https://github.tools.sap/cap/cdsnode/issues) (for legacy-stack issues). Driver name (`hdb`, `@sap/hana-client`, or `@cap-js/hana`), HANA version, and a minimal reproduction help triage.
```
BobdenOs: Duplicate from above. I wouldn't promote this here like this.
Also this one has an internal link. The sentence even says that it is an internal link.
Again `@cap-js/hana` is not a driver.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+The `HanaLobStream` returned by `@sap/hana-client` hangs when iterated with `for await ... of`. The runtime detects this case (via `result[key].constructor.name === 'HanaLobStream'`) and uses a `PassThrough`-based reader instead.
+
+#### UPDATE with streams does not report affected rows
+
+When an `UPDATE` includes streamed parameters (LOBs), `@sap/hana-client` returns `0` even if rows were modified. The runtime treats `streams.length && changes === 0` as success. As a side effect, ETag/affected-row checks during streaming UPDATEs are unreliable.
+
+#### APM decorators (e.g. Dynatrace) cannot wrap internal methods
+
+`@sap/hana-client` does not allow wrapping/decorating `client.prepare()` and other internals. APM integrations that rely on method wrapping have reduced visibility on this driver.
```
BobdenOs: Dynatrace is supported with both drivers.

node.js/databases.md
```diff
@@ -349,3 +349,152 @@ The way that the unified data structures is achieved is by leveraging the [`JSON
...
+
+A consequence of the above limitation: result sets are capped at approximately `0xFFFFFFFB` bytes (~4 GB).
+
+#### `HanaLobStream` cannot be consumed with `for await`
+
+The `HanaLobStream` returned by `@sap/hana-client` hangs when iterated with `for await ... of`. The runtime detects this case (via `result[key].constructor.name === 'HanaLobStream'`) and uses a `PassThrough`-based reader instead.
+
+#### UPDATE with streams does not report affected rows
+
+When an `UPDATE` includes streamed parameters (LOBs), `@sap/hana-client` returns `0` even if rows were modified. The runtime treats `streams.length && changes === 0` as success. As a side effect, ETag/affected-row checks during streaming UPDATEs are unreliable.
```
BobdenOs: This is a valid issue, but nothing like how this is describing it. It doesn't apply to `UPDATE`. It actually applies to `INSERT` queries, because we stream our rows. While the driver returns the number of affected rows before it is possible to upload the source of the entries. Which means that there is no way for HANA to determine how many rows are going to be uploaded afterwards.

Why this does not apply to `UPDATE` is that there is no way to use a `LOB` in the `WHERE` condition of the query, because the only check that is allowed to be done on a `LOB` is `IS NULL` which will be send along already if the value is `null` instead of `Readable`.

PR #1669, opened 2025-02-26
doc(cdl): Add upcoming type projection feature

bugwelle:


cds/cdl.md
```diff
@@ -426,6 +426,20 @@ entity Books {
 }
 ```
 
+Starting with CAP 8.8.0, you can declare structured types based on other
```
stewsk, line 429:
```suggestion
You can declare structured types based on other
```

For new features we never mention when it is available. capire always only reflects the newest version.

stewsk (approved):

renejeglinsky (approved):

PR #1674, opened 2025-02-28
Use enum symbols

stewsk:


bugwelle (approved):

cds/cql.md
```diff
@@ -409,6 +409,24 @@ You don't need a CDL cast if you already use a SQL cast. The compiler will extra
...
 
+## Use enums
+
+In queries, you can use enum symbols instead of the respective literals in places
+where the corresponding type can be deduced:
+
+```cds
+type Status : String enum {open; closed;}
+
+entity Order as projection on Order {
```
bugwelle, line 420:
```suggestion
entity Order as projection on db.Order {
```

cds/cql.md
```diff
@@ -409,6 +409,24 @@ You don't need a CDL cast if you already use a SQL cast. The compiler will extra
...
+  case status when #open    then 0
+              when #closed  then 1 end
+    as status_int : Integer,
+  (status = #closed ? 'is closed' : 'is open')
+    as status_txt : String,  
+  #open as state : Status
+} where status = #open;
```
bugwelle, line 427:
```suggestion
  case status when #open    then 0
              when #closed  then 1 end
    as status_int : Integer,

  (status = #closed ? 'is closed' : 'is open')
    as status_txt : String,  

  #open as state : Status
} where status = #open;
```

cds/cql.md
```diff
@@ -409,6 +409,26 @@ You don't need a CDL cast if you already use a SQL cast. The compiler will extra
...
+entity Order as projection on db.Order {
+  case status when #open    then 0
+              when #closed  then 1 end
+    as status_int : Integer,
+
+  (status = #closed ? 'is closed' : 'is open')
+    as status_txt : String,  
+
+  #open as state : Status
+} where status = #open;
```
bugwelle, line 429:
```suggestion
type Status : String enum { open; closed; in_progress; };

entity OpenOrder as projection on db.Order {
  
  case status when #open        then 0
              when #in_progress then 1 end
    as status_int : Integer,

  (status = #in_progress ? 'is in progress' : 'is open')
    as status_txt : String,  
    
} where status = #open or status = #in_progress;
```

The example didn't make much sense. You're selecting only open orders, but still check if the status is open or closed. Also selecting just `#open` in this example doesn't make much sense either.

Could you make it a more real-life example? I've added a suggestion. :)

renejeglinsky (approved):

PR #1693, opened 2025-03-10
The link to the section "Customization of cds.context.user" was broken

gregorwolf:
From section https://cap.cloud.sap/docs/node.js/authentication#custom there is a link with the text "If you want to customize the user ID, please also have a look at this example.". It wasn't pointing to the correct section.


renejeglinsky (approved):
Thanks a lot Gregor! Will publish that fix today.
🚀

PR #1695, opened 2025-03-10
Improve i18n docs

beckermarc:
`-` vs `_` leads to many confusions with stakeholders, see also:

- https://github.tools.sap/cap/issues/issues/18027
- https://sap.stackenterprise.co/questions/12460/12768
- https://sap.stackenterprise.co/questions/41859/42092

### DIFF
--- a/guides/i18n.md
+++ b/guides/i18n.md
@@ -238,7 +238,7 @@ In CAP Java the preserved locales can be configured via the <Config java>cds.loc
 However this list is configured, ensure to have translations for the listed locales, as the fallback language will otherwise be `en`.
 :::
 
-#### Use Hyphens in File Names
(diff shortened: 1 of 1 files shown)


vmikhailenko:
Maybe also change the config to support `pt_BR`. Initial issue was also very confusing to the stakeholders.

beckermarc:
We could discuss to add that to the default configuration, but that's out of scope of this docs PR.
I don't know the intentions behind the original list of normalized locales.

renejeglinsky:
Pls discuss changes to default localizations with Daniel. That has been changed in the past which had unwanted effects for existing projects and was rolled back if I remember correctly.

renejeglinsky (approved):

PR #1703, opened 2025-03-13
Update getting-started.md for java

ecklie:
cds init --java is wrong I guess


ecklie:
My fault, I accidentally had an outdated cds-dk installed

PR #1705, opened 2025-03-14
Make examples the same order and params

nkaputnik:
In the action/function exaxmples, I think for better understanding, it would be good to have them done consistently with the same order and parameters, or is it intentional that they differ?


guides/providing-services.md
```diff
@@ -1108,9 +1108,9 @@ module.exports = class Sue extends cds.Service {
 ```js
 GET .../sue/sum(x=1,y=2)              // unbound function
 GET .../sue/stock(id=2)               // unbound function
-POST .../sue/add {"x":1,"to":2}       // unbound action
+POST .../sue/add {"x":11,"to":2}       // unbound action
```
```suggestion
POST .../sue/add {"x":11,"to":2}      // unbound action
```

PR #1708, opened 2025-03-14
Java: connect to guides

renejeglinsky:
Hi Matthias,

I added some connections. 
Please let me know [1] if those links are the one you would also set and [2] if you see more topics where we should add such a connection.

### DIFF
--- a/java/cds-data.md
+++ b/java/cds-data.md
@@ -15,7 +15,7 @@ uacp: Used as link target from Help Portal at https://help.sap.com/products/BTP/
 
 {{ $frontmatter.synopsis }}
 
-
+[Before you continue reading, make sure to read the databased guide.](../guides/databases){.learn-more}
 
 ## Predefined Types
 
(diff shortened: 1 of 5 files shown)


java/cds-data.md
```diff
@@ -15,7 +15,7 @@ uacp: Used as link target from Help Portal at https://help.sap.com/products/BTP/
 
 {{ $frontmatter.synopsis }}
 
-
+[Before you continue reading, make sure to read the databased guide.](../guides/databases){.learn-more}
```
renejeglinsky, line 18:
```suggestion
[Before you continue reading, make sure to read the database guide.](../guides/databases){.learn-more}
```

java/cds-data.md
```diff
@@ -15,7 +15,7 @@ uacp: Used as link target from Help Portal at https://help.sap.com/products/BTP/
 
 {{ $frontmatter.synopsis }}
 
-
+[Before you continue reading, make sure to read the database guide.](../guides/databases){.learn-more}
```
BraunMatthias, line 18: I don't think that the DB guide content is base for this chapter. It's rather about core CDS types which is already linked

PR #1717, opened 2025-03-19
shared-db

DanSchlachter:


guides/deployment/shared-db.md
```diff
@@ -0,0 +1,723 @@
...
+    "source": "^/app/(.*)$", //[!code focus]
+    "target": "$1", //[!code focus]
+    "localDir": ".", //[!code focus]
+    "cacheControl": "no-cache, no-store, must-revalidate"
+  }
+]
+```
+:::
+
+#### static content route
```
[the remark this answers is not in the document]
vl-leon: app and appconfig will be needed when we deploy ./fiori

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,755 @@
...
+
+::: code-group
+```json [shared-db/package.json]
+{
+  "name": "@capire/samples-shared-db",
+  "version": "3.0.0",
+  "description": "CAP Sample CDS model deployment for shared-db scenario",
+  "cds": {
+    "sql": {
+      "native_hana_associations": false
```
DanSchlachter, line 134: some explanation of why this step is necessary?

vl-leon: added a docu link

guides/deployment/late-cut-microservices.md
```diff
@@ -0,0 +1,179 @@
+---
+breadcrumbs:
+  - Cookbook
+  - Deployment
+  - Late Cut Microservices
+synopsis: >
+  An overview on monoliths, moduliths and microservices - breaking down their aspects, benefits and drawbacks.
```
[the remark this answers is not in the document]
renejeglinsky: We can add it to the project-words.txt

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,755 @@
...
+```
+
+[Learn more about **how to deploy to Cloud Foundry**.](../deployment/to-cf){.learn-more}
+
+### Database
+
+Add initial database configuration using the command:
+
+```shell
+cds add hana
```
DanSchlachter, line 53: -- for production ?

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,755 @@
...
+```
+
+Delete the generated db folder as we do not need it on root level:
+
+```shell
+rm -r db
+```
+
+
+#### Initial Data
```
DanSchlachter, line 103: can we remove this section entirely?
This is not related to ms with shared db but would be needed on any deployment and is already there e.g. when generating own sample data with cds add data --records ..

guides/deployment/with-shared-db.md
vl-leon: I did not notice any difference when using the parameter

DanSchlachter: then I don't understand, why do we first do `cds add hana` and then set it to `true` in the step after? Can't we leave out one of the two commands?

Akatuoro: We need to set this for each module. The `cds add` command only adds the settings to the root folder.

guides/deployment/with-shared-db.md
Akatuoro: yes, this is related to deployment to hana in general, not to shared-db -> removed it

guides/deployment/with-shared-db.md
Akatuoro: Added a short explanation, particularly since it is the first instance of needing to configure each module

guides/deployment/late-cut-microservices.md
```diff
@@ -0,0 +1,217 @@
...
+status: released
+---
+
+# Late-Cut Microservices
```
renejeglinsky, line 11: I'm having a hard time getting my head around this guide. It discusses various options, benefits and disadvantages but I'm missing the guidance somehow. Is it needed to have that discussion first or can we turn it around (as usual in capire) and have the guidance with explanation first. And then we can show the other relevant aspects and discuss them in a addendum?

renejeglinsky: I tried to create a proposal in my last commit. Basically I switched the order of sections and I moved the Modulith and Microservices section under the first section in the appendix. There, now all concepts are explained and then the more specific aspects are discussed in the following sections. It's still pretty raw, so please let me know what you think about that direction and then we can polish and refine it accordingly.

guides/deployment/late-cut-microservices.md
```diff
@@ -0,0 +1,208 @@
...
+
+* Prefer a late cut
+* Stay flexible in where to cut
+* Prefer staying loosely coupled → e.g. ReviewsService → reviewed events → UPDATE avg ratings
+* Leverage db-level integration selectively → Prefer referring to (public) service entities, not (private) db entities
+
+
+## Appendix
+
+### Monolith, Modulith or Microservice
```
renejeglinsky, line 57: I choose the previous title so that the side menu is readable. I felt that modulith is a concept that is close to monolith and should be explained, but doesn't need to be part of the heading necessarily:
image

WDYT?

Akatuoro: Good point

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,842 @@
...
+   ```
+
+   > Note: the `using` directives above refer to `index.cds` files existing in the target packages. Your projects may have different entry points. 
+
+::: details Try it out
+
+With that we're basically done with the setup of the collector project. At the end of the day, it's just another CAP project with some cds models in it, which we can handle as usual. We can test whether it all works as expected, for example, we can test-compile and test-deploy it to sqlite and hana, build it, and deploy it to the cloud as usual:
+
+```sh
+cds db -2 sql
```
renejeglinsky, line 131: What is `cds db` about? And why is this step needed?

Akatuoro: This step is not needed -> it's just to show that it works just like you would use it in a standalone cap project... since these are not necessary steps, we put it inside a details block

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,842 @@
...
+      service: hana
+      service-plan: hdi-shared
+```
+
+
+
+
+#### Binding to shared db
+
+The only thing left to care about is to ensure all 3+1 projects will be bound and connected to the same dbs at deployment, subscription, and runtime.
```
renejeglinsky, line 225: dbs= database service?

guides/deployment/with-shared-db.md
renejeglinsky: Thanks. 
I've got cds-dk@8.8.1 and that doesn't know `cds db`. What does it do?

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,1082 @@
...
+A comprehensive guide on deploying your CAP application as microservices.
+
+[[toc]]
+
+
+## Create a Solution Monorepo
+
+Assumed we want to create a composite application consisting of two or more micro services, each living in a separate GitHub repository, for example:
+
+- https://github.com/capire/bookstore
```
DanSchlachter, line 18: None existing links - will be picked up by broken link checker? 
Repos don't exist or are outdated - isn't this confusing?
Maybe start with cap-samples directly and add a `details` section for setting up a new project?

guides/deployment/with-shared-db.md
```diff
@@ -0,0 +1,1082 @@
...
+```
+```sh
+cds build --for hana
+```
+
+> Note: As we can see in the output for `cds deploy` and `cds build`, it also correctly collects and adds all initial data from enclosed `.csv` files. 
+:::
+
+
+### Deployment as separate mta
```
DanSchlachter, line 173: shouldn't this be on the same heading level as `All-in-one Deployment`
Can we move it to the end of the guide and clearly mention, that it is an alternative approach? It's unclear that it's either one or the other

Akatuoro: Yes, my thoughts exactly - been wondering where it would be a good fit, let's do it like that

guides/deployment/with-shared-db.md
Akatuoro: Yes, either create the repos or describe some other way.
I think it's good to have this front and center (monorepo with submodules) as it would be the typical project setup.

We could have a script that copies the folders from cloud-cap-samples and keeps them up to date...

Akatuoro: Ideally there'd be a command like
```
git submodule add https://github.com/SAP-samples/cloud-cap-samples --folder bookstore
```

But there isn't as far as I'm aware. There seems to be some way to achieve something similar, but too complicated for a guide :D

renejeglinsky (approved):

PR #1756, opened 2025-04-03
Describe limitations of remote services

mofterdinger:


java/cqn-services/remote-services.md
```diff
@@ -486,3 +486,10 @@ OAuth2DestinationBuilder
         .property("name", "my-destination")
         .build();
 ```
+
+### Known Limitations
+
+The following limitations are currently known for Remote OData Services:
+
+- Streaming of media content is not supported. This means that elements of an entity annotated with `@Core.MediaType` cannot be accessed via Remote OData Services and will cause an error.
+- Accessing an OData Singleton via remote services is limited to read access. Other operations (insert/update/delete) are not supported and will result in an error.
```
beckermarc, line 495:
```suggestion
```

beckermarc (approved):

mofterdinger (commented):
Looks good to me. I can't approve because I'm the initial author of this PR.

java/cqn-services/remote-services.md
```diff
@@ -486,3 +486,9 @@ OAuth2DestinationBuilder
         .property("name", "my-destination")
         .build();
 ```
+
+### Limitations
+
+The following limitations are currently present for Remote OData Services in CAP Java:
+
+- Streaming of media content is not supported. This means that elements of an entity annotated with `@Core.MediaType` cannot be accessed via Remote OData Services and will cause an error. It is recommended to exclude these elements in your projections on the imported entity.
```
renejeglinsky, line 494: If we only describe one limitation, maybe we can get rid of the list.

```suggestion
Streaming of media content is not supported. This means that elements of an entity annotated with `@Core.MediaType` cannot be accessed via Remote OData Services and will cause an error. It is recommended to exclude these elements in your projections on the imported entity.
```

renejeglinsky (approved):

mofterdinger:
@renejeglinsky  Hi Rene, I wonder why I can't see this PR in current Capire docu: https://cap.cloud.sap/docs/java/cqn-services/remote-services#configuring-cds-service-name
There should be the chapter about limitations, but it's missing. Do you know why ?
Thanks,
Markus

renejeglinsky:
Hi @mofterdinger ,
this has not been published externally yet. It will be published with the major release.

PR #1775, opened 2025-04-16
Minor updates for to-cf.md

qmacro:
Fixed a typo ("monrepo") then also a few other very minor details as I was already editing this file. Great section, btw!


renejeglinsky (approved):
Thanks a lot :)

PR #1782, opened 2025-04-17
Update paramValues for querying parametrized views

PirvuCatalin:
creating a a map of parameters using Map.of(...) will actually not work, probably something inside the library doesn't like "ImmutableCollections" created by Map.of(...)


PirvuCatalin:
I can provide more details if necessary.
Also, if you want, I can recreate this change using my SAP github account...

agoerler:
I consider this a bug in the runtime, which we'll fix with the next minor release.

renejeglinsky:
Hi @agoerler @MattSchur Is this already fixed and we can close this PR?

renejeglinsky:
@agoerler Can you briefly comment here?

PR #1792, opened 2025-04-25
add spring-boot-starter-security note

Akatuoro:
Mock users are not initialized if this dependency is not present

### DIFF
--- a/java/security.md
+++ b/java/security.md
@@ -242,6 +242,13 @@ In the example, the `CustomUserInfoProvider` defines an overlay on the default X
 ### Mock User Authentication with Spring Boot { #mock-users}
 
 By default, CAP Java creates a security configuration, which accepts _mock users_ for test purposes.
+
+::: details Requirement
+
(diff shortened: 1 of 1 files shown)


renejeglinsky:
Hi @Akatuoro ,
this dependency is set when creating a project, right? So the scenario would be more like a troubleshooting when someone deleted that dependency, or?

Akatuoro:
Hi @renejeglinsky , no, not always. If you do a cds init --java or a cds init --java --add tiny-sample, it does not get added. It only gets added with sample or xsuaa.
This means that if someone is creating their model from scratch or from tiny-sample, they'll run into this.

And yes, for troubleshooting this info is of course also helpful.

renejeglinsky:
Thanks @Akatuoro for the explanation. Then I'm fine with having that note here. It will also suffice as troubleshooting for users who search for this.

renejeglinsky (approved):

PR #1793, opened 2025-04-25
Java: flows

JenniferStamm:
WIP, draft for early feedback

### DIFF
--- a/java/flows.md
+++ b/java/flows.md
@@ -1,7 +1,7 @@
 ---
 synopsis: >
   Learn about the flow feature in CAP Java.
-status: draft TODO: released (only valid status)
+status: released
 ---
 
 # Flow: Modeling State Transitions <Beta /> <Since version="4.0.0" of="com.sap.cds:cds-feature-flow" />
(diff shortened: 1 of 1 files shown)


java/flows.md
```diff
@@ -0,0 +1,126 @@
...
+        from:[ Open ],                     
+        to: Accepted 
+    })
+    action acceptTravel();
+  };
+
+}
+```
+
+We can simply remove the [AcceptRejectHandler](https://github.com/SAP-samples/cap-sflight/blob/main/srv/src/main/java/com/sap/cap/sflight/processor/AcceptRejectHandler.java). All of the logic is taken care of by the flow feature. The flow feature will validate that entry state is `Open`. The flow feature will transit the status to `Accepted` and `Canceled` respectively.
```
BraunMatthias, line 120: "You can simply remove ..."

I fear customers will not investigate existing SFlight code ni depth - you could still name the benefit, that a custom reject handler is not needed due to flow capabilities.

BraunMatthias (commented):
Looks good to me! We need to discuss whether Flows should be top level topic (keep it as it is for now)

java/flows.md
JenniferStamm: I moved it around, so that the benefit is named first and added a note that for those who like to follow along

JenniferStamm:
closing because release date has moved

PR #1798, opened 2025-04-28
Explain some caveats in change tracking

vmikhailenko:
Change tracking documentation is extended to attempt to clear two things that cause confusion between stakeholders: 

- the element-defined identifier should not be used for compositions as it does very little sense for them. 
- the stakeholders need to ensure that change-tracked entities are updated using deep updates from the root of the document (entity with compositions). Notably, the changes are not linked to their expected parents and this causes confusion.


java/change-tracking.md
```diff
@@ -195,18 +195,52 @@ annotate Orders {
...
+annotate OrderItems with @changelog: [
+  parent.OrderNo,
+  supplierName,
+];
+```
+
+Each entry in the changelog in case of the deep change will have appropriate identifier depending on the place of the changed element in the `Order` structure.
+
+You can also define the identifier on the element declaring the composition in the `Orders`, such identifier will not 
+be applied to the changes in the `OrderItems`. Identifiers on the elements are intended for the associations.
```
beckermarc, line 242: Here you are saying you can use, but it is not intended. Maybe don't mention it then at all. I also think it might make sense to explain the pattern for "Compositions" (which seems to be the same pattern also for single entities), before the additional handling for "Associations".

Currently from the documentation flow you are introducing a general pattern "Identifiers for Associations" and then have to argue against it again further down. I think it would be better if you change the order therefore.

java/change-tracking.md
```diff
@@ -195,18 +195,52 @@ annotate Orders {
...
+```
+
+You can assign identifiers on these entities so that changes in the `Orders` are annotated with some element from it and 
+the changes on the item of the order are annotated with the different set of values.
+
+```cds
+annotate Orders with @changelog: [OrderNo];
+
+annotate OrderItems with @changelog: [
+  parent.OrderNo,
```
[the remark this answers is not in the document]
vmikhailenko: Exactly.

java/change-tracking.md
```diff
@@ -268,10 +302,10 @@ the changes across the complete document and stores them in the change log with
...
 the deep update, for example, the customer name in the order and the quantity of the item, the change log contains
 two entries: one for the order and one for the item. The change log entry for the item will also reflect that
-the root of the change is an order.
+the root of the change is an order. Both changes will be reachable through the association `changes` of the order entity.
 
 :::warning Prefer deep updates for change tracked entities
-If you change the values of the `OrderItems` entity directly via an OData request or a CQL statement, the change log contains only one entry for the item and won't be associated with an order.
+If you change the values of the `OrderItems` entity directly via an OData request or a CQL statement, the change log contains only one entry for the item and won't be associated with an order and will not be reachable through the `changes` association. While the updates of composition targets directly is possible, the change tracking feature does not attempt to resolve the parent entity by itself, it requires that either OData request or a CQL statement provide the reference to the parent e.g. `Orders`.
```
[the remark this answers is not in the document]
vmikhailenko: Created the BLI to sort this out.

java/change-tracking.md
```diff
@@ -173,39 +173,61 @@ You should consider this when you decide what to include in the identifier.
...
 ```cds
 annotate Orders {
   customer @changelog: [ customer.name ]
 }
 ```
 
-Elements from the `@changelog` annotation value must always be prefixed by the association name. The same caveats as for the identifiers for the entities apply here.
-
-If you annotate a composition with an identifier, the change log will contain an entry with the identifier's value. Additionally, it will include change log entries for all annotated elements of the composition's target entity.
+Elements from the `@changelog` annotation value must always be prefixed by the association name and the identifier of the target entity is not considered at all.
```
beckermarc, line 220: > the identifier of the target entity is not considered at all

What is the reason for this?

java/change-tracking.md
```diff
@@ -173,39 +173,61 @@ You should consider this when you decide what to include in the identifier.
...
 annotate Orders {
   customer @changelog: [ customer.name ]
 }
 ```
 
-Elements from the `@changelog` annotation value must always be prefixed by the association name. The same caveats as for the identifiers for the entities apply here.
-
-If you annotate a composition with an identifier, the change log will contain an entry with the identifier's value. Additionally, it will include change log entries for all annotated elements of the composition's target entity.
+Elements from the `@changelog` annotation value must always be prefixed by the association name and the identifier of the target entity is not considered at all. 
+The same caveats as for the identifiers for the entities apply here.
```
beckermarc, line 221: What are these caveats?

java/change-tracking.md
vmikhailenko: That was always how I have understood the intention of syntax introduced by Nodies. FK by default (even if association target has own identifier) or own identifier if you wish. 
We can no longer look at their syntax I think so this now just documents the way it works.

java/change-tracking.md
beckermarc: Can we explain this in the text here, or where is this mentioned already earlier?

java/change-tracking.md
```diff
@@ -268,10 +290,10 @@ the changes across the complete document and stores them in the change log with
 For example, given the order and item model from above, if you change values for the tracked elements with
 the deep update, for example, the customer name in the order and the quantity of the item, the change log contains
 two entries: one for the order and one for the item. The change log entry for the item will also reflect that
-the root of the change is an order.
+the root of the change is an order. Both changes will be reachable through the association `changes` of the order entity.
```
beckermarc, line 293: In which scenarios would the items entity have a `changes` association at all? Wouldn't it be recommended to have the changeTracked aspect only on root entity. Do we explain that? Or are there good reasons to have it also on child entities?

java/change-tracking.md
vmikhailenko: https://github.com/cap-js/docs/blob/03a1dc0a91269898dd850e865e2d8f118f30d45a/java/change-tracking.md?plain=1#L161

beckermarc: I think that is not clear that this hint links to that section in the docs. I think we should make that more explicit, by either:

a) Mentioning the caveats explicitly again, e.g. by something like "be careful when using aliases in your projects as also explained in <link>" or
b) Moving that topic into a dedicated section or warning box and referring to that section from both places

vmikhailenko: Introduced new subsection. This is general limitation, hope that is clearer.

java/change-tracking.md
vmikhailenko: Added explicit instruction.

No, I consider dedicated changelogs for composition items as something very unusual. Not sure if that is even work. :)

java/change-tracking.md
```diff
@@ -158,54 +162,71 @@ annotate Bookshop.Book with @changelog: [
...
+annotate OrderItems with @changelog: [
+  parent.OrderNo,
+  supplierName,
+];
+```
+
+Changes for `Orders` and `OrderItems` will have their own respective target or root identifiers. Values of the changed fields are not affected, 
+such identifiers simply annotate them with additional context. When no annotation is present, identifier is empty.
+
+For associations, you might also replace the changed values with some value of the associated entity instead of the foreign key.
```
beckermarc, line 201: Maybe first explain that foreign key values are used by default for associations. This is somehow hidden between the lines in this statement. Once that is established you can mention that this can be changed by using explicit identifiers with the annotations on the association element.

What I am wondering: What is used for to one Compositions that also have a foreign key on the source side? Foreign key or identifier from target? What if the target doesn't have an identifier?

vmikhailenko: >What I am wondering: What is used for to one Compositions that also have a foreign key on the source side? Foreign key or identifier from target? 

Foreign key. If it is annotated and it is on the source side, then it will be stored as such. It does very little sense to track it, I think.

> What if the target doesn't have an identifier?

Then, its changes will have empty one.

java/_menu.md
```diff
@@ -19,7 +19,7 @@
 # [Fiori Drafts](../../java/fiori-drafts)
 # [Messaging](messaging)
 # [Audit Logging](auditlog)
-# [Change Tracking](../../java/change-tracking)
+# [Change Tracking](../../change-tracking)
```
beckermarc, line 22:
```suggestion
# [Change Tracking](../../java/change-tracking)
```

java/change-tracking.md
```diff
@@ -156,56 +159,74 @@ annotate Bookshop.Book with @changelog: [
...
+  supplierName,
+];
+```
+
+Changes for `Orders` and `OrderItems` will have their own respective target or root identifiers filled. 
+
+### Human-readable values for associations
+
+For associations, the value of the foreign key is stored in the changelog by default. You can change this and store values of the associated entity instead. 
+This kind of identifier changes the values stored in the changelog, while [entity identifiers](#identifiers-for-entities) annotate changed values.
```
beckermarc, line 202: This is very helpful. I think I only really understood the difference now after reading these two sentences :)

beckermarc (approved):

java/change-tracking.md
```diff
@@ -156,56 +159,74 @@ annotate Bookshop.Book with @changelog: [
...
 ```cds
 annotate Orders {
   customer @changelog: [ customer.name ]
 }
 ```
-
-Elements from the `@changelog` annotation value must always be prefixed by the association name. The same caveats as for the identifiers for the entities apply here.
-
-If you annotate a composition with an identifier, the change log will contain an entry with the identifier's value. Additionally, it will include change log entries for all annotated elements of the composition's target entity.
+Elements from the `@changelog` annotation value must always be prefixed by the association name. Identifier defined on the association target is never considered.
```
vmikhailenko, line 211:
```suggestion
Elements from the `@changelog` annotation value must always be prefixed by the association name.
```

vmikhailenko: Unnecessary.

renejeglinsky (approved):
Please check my edits and merge if there are no objections. Thanks :)

PR #1805, opened 2025-04-30
Fix Node.js event `code` argument type

swaldmann:
For cap/issues/issues/18314.


danjoa (changes requested):
We anyways need to thoroughly revise these docs, as we cleaned up respective implementations

swaldmann:
Fixed in https://github.com/cap-js/docs/commit/a16f555387a59be732a0cd347a1d8db737d2ad93, closing this PR.

PR #1809, opened 2025-05-05
cds 9: modified input validations

David-Kunz:
- [x] rm `@readonly` and `@insertonly` from Access Control Annotations
  - [x] rm `@Common.FieldControl` from Input Validation
  - [x] rm `@assert.notNull` from Input Validation
  - [x] mv `@assert.unique` to Database Constraints
  - [x] add link to database constraints in Input Validations
  - [x] document `not null` in Database Constraints
  - [x] document `not null` (in actions/functions)


advanced/hana.md
```diff
@@ -50,7 +50,7 @@ Steps to match the signature of a database object in a facade entity:
 * For a view, table function, or calculation view with parameters, check that the parameter names and types match, too.
   Functions with table-like input parameters are not supported.
 
-> Note: If a field of that entity is defined as `not null` and you want to disable its runtime check, you can add [`@assert.notNull: false`](../guides/providing-services#assert-notnull). This is important if you want to use, for example [SAP HANA history tables](https://help.sap.com/docs/SAP_HANA_PLATFORM/6b94445c94ae495c83a19646e7c3fd56/d0b2c5142a19405fb912f71782cd0a84.html).
+> Note: If a field of that entity is defined as `not null` and you want to disable its runtime check, you can add `@assert.notNull: false`. This is important if you want to use, for example [SAP HANA history tables](https://help.sap.com/docs/SAP_HANA_PLATFORM/6b94445c94ae495c83a19646e7c3fd56/d0b2c5142a19405fb912f71782cd0a84.html).
```
David-Kunz, line 53: I left this because it still applies to CAP Java, no?

agoerler (commented):
I am afraid that we have not sufficiently discussed the meaning of `not null` for calculated/computed elements.

guides/databases.md
```diff
@@ -841,6 +841,44 @@ Find here a collection of resources on selected databases and their reference do
...
+
+You can specify that a column's value must not be null by adding the [`not null` constraint](http://localhost:5173/docs/cds/cdl#null-values) to the element, for example:
+
+```cds
+entity Books {
+  key ID: Integer;
+  title: String not null;
+}
+```
+
```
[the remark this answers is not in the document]
David-Kunz: > If an element is defined in a view as not null

How does one achieve this? Is that even possible?

advanced/hana.md
agoerler: CAP Java does evaluate and enforce `not null`. As `not null` is converted in EDMX to `Nullable="false"`, which Olingo enforces. `@assert.notNull: false` has no effect on the EMDX.

The runtime also enforces that no null values are provided upon update for a `not null`-element - unless the element is additionally annotated with `@assert.notNull : false`.

As far as I understand the coding `@assert.notNull:true` has no effect unless the annotated element is `not null`.

@beckermarc - please double-check

guides/databases.md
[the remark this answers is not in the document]
agoerler: I checked and have the impression that it's not possible. 

But then, the question arises: How do I express that a virtual/calculated/@CoreComputed element is guaranteed to not be null?

danjoa (approved):
- did some more cleanup in https://github.com/cap-js/docs/pull/1809/commits/e66a9124823b8de9c6de2077e493ddac31c731d1 
- see the check marks in the [top comment](https://github.com/cap-js/docs/pull/1809#issue-3039208029)
- Fine for me with that.

guides/databases.md
```diff
@@ -784,6 +784,45 @@ Find here a collection of resources on selected databases and their reference do
 
 ## Database Constraints
 
+### Not Null
+
+You can specify that a column's value must not be `NULL` by adding the [`not null` constraint](http://localhost:5173/docs/cds/cdl#null-values) to the element, for example:
```
renejeglinsky, line 789:
```suggestion
You can specify that a column's value must not be `NULL` by adding the [`not null` constraint](../cds/cdl#null-values) to the element, for example:
```

guides/databases.md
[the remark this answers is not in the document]
agoerler: I have double-checked with @stewsk - you can actually define a virtual element, which is `not null`:

```cds
type NotNullableString : String not null;

view V as select from E { *, 'x' as nns : NotNullableString };
```

this is **not** possible, however:

```cds
view V as select from E { *, 'x' as nns : String not null };
```

agoerler (approved):

guides/databases.md
```diff
@@ -784,6 +784,45 @@ Find here a collection of resources on selected databases and their reference do
...
+```cds
+@assert.unique: {
+  locale: [ parent, locale ],
+  timeslice: [ parent, validFrom ],
+}
+entity LocalizedTemporalData {
+  key record_ID : UUID; // technical primary key
+  parent    : Association to Data;
+  locale    : String;
+  validFrom : Date;  validTo : Date;
```
renejeglinsky, line 812:
```suggestion
  validFrom : Date;  
  validTo : Date;
```

renejeglinsky (approved):

PR #1810, opened 2025-05-05
cds^9: chunkSize: parallel ? 10 : 1

sjvans:


David-Kunz:
Already added this in https://github.com/cap-js/docs/pull/1790

PR #1821, opened 2025-05-08
Clarifications for CF deployment guide

swaldmann:


guides/deployment/to-cf.md
```diff
@@ -231,14 +231,20 @@ The previous steps are required _only once_ in a project's lifetime. With that d
 
 ## Build and Deploy
 
+Make sure you are logged in to Cloud Foundry and target the space you want to deploy to:
+```sh
+cf login --sso
```
renejeglinsky, line 236: As there are multiple options I'd like to give a hint which scenario this is.
```suggestion
cf login --sso  // to log on with SAP Universal ID
```

swaldmann: What do you mean by multiple options?

renejeglinsky: You don't have to use the sso option and you would also need the API endpoint if you haven't logged in recently. I think I agree with the sso option as I think this might be widely used by now. But for those wondering what this is, I'd like to give a hint.

swaldmann:
Applied the suggestion, merging as I'd like to have this hint for the testing round next week (at least in the internal deployment).

PR #1836, opened 2025-05-12
chore(cookbook): improve databases guide

patricebender:
TL;DR: Make db guides more homogenous.

- prefer `cds add …` over `npm add …` in all guides
- dont use `new` and `old` as those terms are outdated pretty quick → `cap-js/databases` and `legacy` are more accurate


renejeglinsky:
@patricebender Meanwhile there are many conflicts. How should we proceed with this PR?

patricebender:
lets just close it, bit of a pity that no one cared about this. However, with Daniels recent changes it seems to be obsolete.

PR #1843, opened 2025-05-14
Remove deprecated APIs from CAP Java Messaging docs

beckermarc:
We should no longer mention deprecated APIs in our CAP Java messaging docs, that are already deprecated since CAP Java 3.0 (almost a year ago!). Those APIs will also be finally removed with the next major release. This documentation change is however already valid, before the upcoming major release.

### DIFF
--- a/java/messaging.md
+++ b/java/messaging.md
@@ -811,6 +811,7 @@ Besides these kinds of topic manipulations, additional topic manipulations might
 
 
 ### Messages Representation
+{#enhanced-messages-representation}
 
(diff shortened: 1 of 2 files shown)


t-bonk (approved):

renejeglinsky (approved):

java/messaging.md
```diff
@@ -808,99 +810,19 @@ When using SAP Event Mesh, the placeholder `$namespace` can be used to dynamical
...
-```txt
-hello world
-```
-
-If the service is configured with the structured flag, the message is converted to a map and on the consumer side `TopicMessageEventContext.getData()` returns:
-
-```json
-{"data": {"message": "hello world"}}
-```
+### Messages Representation
```
renejeglinsky, line 813:
```suggestion
### Messages Representation
{#enhanced-messages-representation}
```

PR #1848, opened 2025-05-15
Adding Mapping table for OpenAPI to CDS Conversion( Import)

RoshniNaveenaS:


RoshniNaveenaS:
@renejeglinsky can you comment if this table would be an useful information to have in OpenAPI page?

renejeglinsky:
I do think it's useful. Could you move it to the end where we also have this OData mapping table?

renejeglinsky:
Hi @stewsk ,
do you have an opinion here?

stewsk:
Hi @renejeglinsky , yes definitely helpful, and having it close to the OData table also seems reasonable

PR #1850, opened 2025-05-15
Dedicated chapter about code generation configuration

vmikhailenko:
Introduces chapter to describe switches in Java Code generator. Previously that was documented in the Maven Plugin. This chapter somewhat repeats that documentation and aims to improve the visibility of said switches.


java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
 :::warning SAP Business Application Studio
 If you develop your application in SAP Business Application Studio and Java 21 is not available there, use the Java 17, instead.
 :::
+
+## Configuring Code Generation for Typed Access {#codegen-config}
```
beckermarc, line 67: Maybe rather put this whole paragraph into the "Building Applications" section?

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+### Package for Generated Code
+
+The option [`basePackage`](/java/assets/cds-maven-plugin-site/generate-mojo.html#basePackage) can be used to specify package as the root for generated code. The underlying package structure will reflect namespaces defined in your CDS model.
+
+### Filter for CDS Entities
+
+By default, complete model of your application is generated including all imported or re-used models. 
+You can use options [`includes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#includes) or [`excludes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#excludes) to specify the part of your overall model that is subject to code generation.
+
+For example, the pattern `my.bookshop.*` will exclude all definitions with namespace `my.bookshop` and the pattern `my.bookshop.**` will exclude all definitions with fully qualified name starting with `my.bookshop`.
```
beckermarc, line 97: Maybe mention that `excludes` operate on the sum of all `includes`

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+   Books.create().author(author).title("Wuthering Heights");
+```
+
+The generation mode is configured by the option [`methodStyle`](/java/assets/cds-maven-plugin-site/generate-mojo.html#methodStyle). The selected style affects all entities and event contexts in your services. The default value is `BEAN`, which represents JavaBeans-style interfaces.
+
+Once, when starting a project, decide on the style of the interfaces that is best for your team and project. We recommend the default JavaBeans style.
+
+The way the interfaces are generated determines only how data is accessed by custom code. It does not affect how the data is represented in memory and handled by the CAP Java runtime.
+
+Moreover, it doesn't change the way how event contexts and entities, delivered by CAP, look like. Such interfaces from CAP are always modelled in the default JavaBeans style.
```
beckermarc, line 133: Which entities do you mean here?

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+
+### Code Generation Features
```
beckermarc, line 135: Is this doc already reflecting the 4.0.0 state regarding defaults of these parameters?

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+  ```
+  It does not introduce any additional type checks at runtime, the correctness of the assignment is checked only at the time of compilation.
+
+- [`interfacesForAspects`](/java/assets/cds-maven-plugin-site/generate-mojo.html#interfacesForAspects)
+
+  If your entity is modelled with the [composition of aspects](/cds/cdl#with-named-targets), the generated interfaces always reference original aspect as type for setters and getters.
+  When this switch is enabled, code generator will use type generated by the compiler instead of type of the aspect itself and will include methods to fetch keys, for example.
+
+  :::warning Limitations
+  This is supported only for the named aspects (inline targets are not supported) and does not respect all possible options how such entities might be exposed by services.
```
beckermarc, line 160: What does it mean that inline targets are not supported? What happens if inline targets are used in the CDS? Will this generate invalid Java code?

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+  :::
+
+- [`betterNames`](/java/assets/cds-maven-plugin-site/generate-mojo.html#betterNames)
+  
+  CDS models from external sources might include elements that have some special characters in their names or include elements that clash with Java keywords. Such cases always can be solved with [renaming features](/java/cds-data#renaming-elements-in-java) provided by code generator, but in case of large models, this is tedious.
+  This switch enables following conversions on the names coming from CDS models to reduce the amount of renaming that needs to be done:
+
+  - Names from CDS model that are Java keywords are prefixed with `_`.
```
beckermarc, line 168: I thought they were suffixed.

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+  :::
+
+- [`betterNames`](/java/assets/cds-maven-plugin-site/generate-mojo.html#betterNames)
+  
+  CDS models from external sources might include elements that have some special characters in their names or include elements that clash with Java keywords. Such cases always can be solved with [renaming features](/java/cds-data#renaming-elements-in-java) provided by code generator, but in case of large models, this is tedious.
+  This switch enables following conversions on the names coming from CDS models to reduce the amount of renaming that needs to be done:
+
+  - Names from CDS model that are Java keywords are prefixed with `_`. 
+  - Names from CDS model that use characters that are not valid as Java identifiers, are replaced by `_`. This, however, might lead to a conflicts between names that yield same name in Java.
+  - Characters `/` and `$` behave just like `_`: they used to provide camel-cased name and provide better support for models coming from ABAP systems.
```
beckermarc, line 170: > they used to provide camel-cased name and provide better support for models coming from ABAP systems.

Can you clarify that. I am not sure if I understand that sentence.

EDIT: Ah I think I got it, they are not replaced by "_" but behave like a "_" would. Maybe it would be beneficial to explain how a "_" behaves. This is not explicitly explained here at the moment.

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+  - If the name of the element is prefixed by `_`, this character will remain the name after conversions. This supports convention where association and its foreign key have names like `_assoc` and `assoc`.
```
beckermarc, line 171: I think also this section would benefit from an explanation somewhere how "_" is treated otherwise.

java/developing-applications/configuring.md
vmikhailenko: Order was to do it here. :) I thought that we need introduce new top-level chapter: there is much more stuff that needs to be documented. 

@BraunMatthias @agoerler ??

java/developing-applications/configuring.md
vmikhailenko: Nothing in particular. This is simply to indicate that people should not expect that stuff coming from us will also change style if codegen style is changed. Basically, `setCompleted()` will always be `setCompleted()` and will not magically go away.

java/developing-applications/configuring.md
vmikhailenko: They still yield interface matching the bare aspect. We have BLI to fix this for good.

java/developing-applications/configuring.md
beckermarc: But that's event contexts. I find entities confusing in that section, as we specifically use the different style for entity interfaces.

java/developing-applications/configuring.md
vmikhailenko: I rather use something more generic to avoid deep dives in the implementation details.

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+
+- [`cqnService`](/java/assets/cds-maven-plugin-site/generate-mojo.html#cqnService)
+  
+  This switch enables generation of [typed service interfaces](/java/cqn-services/application-services#trigger-action-or-function). This switch is deprecated and should not be disabled.
```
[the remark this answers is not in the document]
vmikhailenko: Why not? It is still available, let's say somewhere what it does and that people should not use it.

java/developing-applications/configuring.md
beckermarc: Well I think it is helpful that these rules are explained. They are very visible and I wouldn't consider it an implementation detail.

java/developing-applications/configuring.md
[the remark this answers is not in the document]
beckermarc: Because we plan to remove it, so why let people use something that we plan to remove. Usually we don't recommend deprecated stuff, especially if it is discouraged to set it.

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,151 @@ properties:
...
+
+- [`eventContext`](/java/assets/cds-maven-plugin-site/generate-mojo.html#eventContext)
+
+  This switch enables generation of interfaces representing contexts for actions, functions and events that are modelled through CDS. This switch is deprecated and should not be disabled. 
+
+:::warning Check migration guides!
+In major releases of CAP Java, some of these switches can be made new default and some other switches might be removed. This might introduce compile errors
+in your application that needs to be fixed.
+:::
```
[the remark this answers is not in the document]
vmikhailenko: It was intended as a reminder to someone who comes here from search and somehow confused why some switch is no longer mentioned or why they have compile errors now.

java/developing-applications/configuring.md
vmikhailenko: To which extent?

java/developing-applications/configuring.md
```diff
@@ -63,3 +63,143 @@ properties:
...
+  This is supported only for the named aspects (inline targets are not supported) and does not respect all possible options how such entities might be exposed by services.
+  :::
+
+- [`betterNames`](/java/assets/cds-maven-plugin-site/generate-mojo.html#betterNames)
+  
+  CDS models from external sources might include elements that have some special characters in their names or include elements that clash with Java keywords. Such cases always can be solved with [renaming features](/java/cds-data#renaming-elements-in-java) provided by code generator, but in case of large models, this is tedious.
+  This switch enables following conversions on the names coming from CDS models to reduce the amount of renaming that needs to be done:
+
+  - Names from CDS model that are Java keywords are suffixed with `_`. 
+  - Names from CDS model that use characters that are not valid as Java identifiers, are replaced by `_`. This, however, might lead to a conflicts between names that yield the same name in Java.
```
beckermarc, line 169: If they are replaced by `_`, will the `_` be removed again in the end according to the rules of bullet point 3 or remain?

beckermarc: At least it should be understandable by clients how the names are generated. For me it still feels a little unpredictable.
For me this question came from this statement:

> This, however, might lead to a conflicts between names that yield the same name in Java.

However that doesn't seem to be the case, as something like `my_foo` would be handled differently than `my-foo`. The first would yield `myFoo`, while the second would yield `my_foo`. This feels strange IMHO.

vmikhailenko: `my-foo` and `my#foo` can clash.

beckermarc: Sure, from the text my naive first assumption was the above two would clash. Why not handle `-` and `#` also as camel casing separators?

vmikhailenko: Can be. Will not eliminate the clashes, though. I would rather wait until someone complains about that and, more importantly, that our case conversion is broken and inconsistent.

beckermarc: Well I wouldn't wait until someone complains to be honest. Aren't we trying to make things consistent now with the "betterNames" switch? I wouldn't want to do another incompatible change, if we already know about inconsistencies in our handling now.

What do you mean with our case conversion is broken and inconsistent? Why don't we fix this with the new "betterNames" switch?

vmikhailenko: We decided that this is too incompatible, especially with builder interfaces. Only technical changes remain.

java/developing-applications/configuring.md
BraunMatthias: I think it should be clode to https://pages.github.tools.sap/cap/docs/java/developing-applications/building#the-maven-archetype, maybe right after the archetype

vmikhailenko: Done, but I think this might create an impression that once needs to configure code generation explicitly.

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  void setManager(Map<String, ?> manager);
+  ```  
+  you get the following:
+
+  ```java
+  void setManager(Manager manager);
+  ```
+  It does not introduce any additional type checks at runtime, the correctness of the assignment is checked only at the time of compilation.
+
+- [`interfacesForAspects`](/java/assets/cds-maven-plugin-site/generate-mojo.html#interfacesForAspects)
```
beckermarc, line 462: Should we also mention that there are scenarios that don't work if this switch is not set? AFAIK compilation of the generated code breaks, when using the attachments aspect, but not enabling this switch.
I am wondering why we didn't enable this default in the 4.0 major.

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  This is supported only for the named aspects (inline targets are not supported) and does not respect all possible options how such entities might be exposed by services.
+  :::
+
+- [`betterNames`](/java/assets/cds-maven-plugin-site/generate-mojo.html#betterNames)
+
+  CDS models from external sources might include elements that have some special characters in their names or include elements that clash with Java keywords. Such cases always can be solved with [renaming features](/java/cds-data#renaming-elements-in-java) provided by code generator, but in case of large models, this is tedious.
+  This switch enables following conversions on the names coming from CDS models to reduce the amount of renaming that needs to be done:
+
+    - Names from CDS model that are Java keywords are suffixed with `_`.
+    - Names from CDS model that use characters that are not valid as Java identifiers, are replaced by `_`. This, however, might lead to a conflicts between names that yield the same name in Java.
```
beckermarc, line 478: Should we also mention that these two kind of `_` rules are not followed up by the rule below, which replaces `_` with casing rules again? Maybe move these two rules closer to the last rule, as they are somehow similar in nature.

java/developing-applications/building.md
vmikhailenko: Because not all cases that are broken are fixed with this switch. We have BLI to fix this for good.

java/developing-applications/building.md
```diff
@@ -479,9 +479,9 @@ Other options in this goal enable or disable certain features that change the wa
...
 
+    - Characters `/` and `$` behave as a separator for the name during case conversions, similar to `_` and `.`. For example, `GET_MATERIAL` yields `GetMaterial` (or `getMaterial` for attributes and methods). The same now applies for the names with `/`, for example, name `/DMO/GET_MATERIAL` will be converted to `DmoGetMaterial`
     - Names from CDS model that are Java keywords are suffixed with `_`.
     - Names from CDS model that use characters that are not valid as Java identifiers, are replaced by `_`. This, however, might lead to a conflicts between names that yield the same name in Java.
-    - Characters `/` and `$` behave as a separator for the name during case conversions, similar to `_` and `.`. For example, `GET_MATERIAL` yields `GetMaterial` (or `getMaterial` for attributes and methods). The same now applies for the names with `/`, for example, name `/DMO/GET_MATERIAL` will be converted to `DmoGetMaterial`.
     - Leading `_` will remain in the name after conversions. This supports convention where association and its foreign key have names like `_assoc` and `assoc`.
```
beckermarc: Already better I think. However maybe it makes it more understandable if we make the first rule a general statement and then say that there are three rules where `_` can be added to the names, however not with the effect previously explained.

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
 [Learn more about configuration and `cds.env`](../../node.js/cds-env){.learn-more}
 
 
+## Code Generation for Typed Access {#codegen-config}
+
+The [interfaces for typed access](../cds-data#generated-accessor-interfaces) are generated at each build
+by the [goal `cds:generate`](/java/assets/cds-maven-plugin-site/generate-mojo.html) of the [CDS Maven Plugin](/java/assets/cds-maven-plugin-site/plugin-info.html).
```
agoerler:
```suggestion
by the [`cds:generate`](/java/assets/cds-maven-plugin-site/generate-mojo.html) goal of the [CDS Maven Plugin](/java/assets/cds-maven-plugin-site/plugin-info.html).
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+Each time your application is built, these interfaces are regenerated so you should exclude them from your version control system.
+
```
agoerler: Please move below the config of the generate goal. Maybe add a _.gitignore_-snippet.

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+    <configuration>
+        <basePackage>cds.gen</basePackage>
+        ...
+    </configuration>
+</execution>
+```
+
+### Package for Generated Code
+
+The option [`basePackage`](/java/assets/cds-maven-plugin-site/generate-mojo.html#basePackage) can be used to specify package as the root for generated code. The underlying package structure will reflect namespaces defined in your CDS model.
```
agoerler:
```suggestion
The option [`basePackage`](/java/assets/cds-maven-plugin-site/generate-mojo.html#basePackage) can be used to specify a base package prefix for generated code. The suffix package structure will reflect namespaces defined in your CDS model.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+### Filter for CDS Entities
+
+By default, complete model of your application is generated including all imported or re-used models.
```
agoerler:
```suggestion
By default, the complete model of your application is generated including all imported or re-used models.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+You can use options [`includes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#includes) and [`excludes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#excludes) to specify the part of your overall model that is subject to code generation. Both inclusion and exclusion can be used together, inclusion evaluated first, then exclusion filter out included set of entities.
```
agoerler:
```suggestion
Use the options [`includes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#includes) and [`excludes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#excludes) to specify the part of your overall model that is subject to code generation. Both inclusion and exclusion can be used together, inclusion is evaluated first, then exclusion filters from included set of entities.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+:::warning Cross-namespace references are not resolved
+Options `includes` and `excludes` are simple filters. If included parts of your model reference types from the excluded area, the resulting code will not compile.
+:::
+
+### Style of Interfaces
+
+By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification.
+
+Following example uses accessor interfaces that have been generated with the default (JavaBeans) style:
```
agoerler:
```suggestion

```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+These options use patterns that are applied on the fully qualified names of the entities in CDS models. For example, the pattern `my.bookshop.*` will cover all definitions with namespace `my.bookshop` and the pattern `my.bookshop.**` will cover all definitions with fully qualified name starting with `my.bookshop`.
+
+:::warning Cross-namespace references are not resolved
+Options `includes` and `excludes` are simple filters. If included parts of your model reference types from the excluded area, the resulting code will not compile.
+:::
+
+### Style of Interfaces
+
+By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification.
```
agoerler:
```suggestion
By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification. In this style getter and setter method names are prefixed with `get` and `set`, resp.:
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+    author.setName("Emily Brontë");
+
+    Books book = Books.create();
+    book.setAuthor(author);
+    book.setTitle("Wuthering Heights");
+```
+
+Alternatively, you can generate accessor interfaces in _fluent style_. In this mode, the getter methods are named after the property names. To enable fluent chaining, the setter methods return the accessor interface itself.
+
+Following is an example of the fluent style:
```
agoerler:
```suggestion
Alternatively, you can generate accessor interfaces in _fluent style_. In this mode, the getter methods are named after the property names. To enable fluent chaining, the setter methods return the accessor interface itself:

```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+```java
+   Authors author = Authors.create().name("Emily Brontë");
+   Books.create().author(author).title("Wuthering Heights");
+```
+
+The generation mode is configured by the option [`methodStyle`](/java/assets/cds-maven-plugin-site/generate-mojo.html#methodStyle). The selected style affects all entities and event contexts in your services. The default value is `BEAN`, which represents JavaBeans-style interfaces.
+
+Once, when starting a project, decide on the style of the interfaces that is best for your team and project. We recommend the default JavaBeans style.
```
agoerler:
```suggestion
Once, when starting a project, decide on the style of the interfaces that is best for your team and project.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+The way the interfaces are generated determines only how data is accessed by custom code. It does not affect how the data is represented in memory and handled by the CAP Java runtime.
```
agoerler:
```suggestion
The way the interfaces are generated only determines how data is accessed by custom code. It does not affect how the data is represented in memory and handled by the CAP Java runtime.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+Moreover, it doesn't change the way how event contexts, delivered by CAP, look like. Such interfaces from CAP are always modelled in the default JavaBeans style.
+
+### Code Generation Features
+
+Other options in this goal enable or disable certain features that change the way generated code looks in a certain aspect. These changes can be incompatible with the existing code and require manual adaptation.
```
agoerler, line 448: > These changes can be incompatible with the existing code and require manual adaptation.

This also applied to the `methodStyle`. The difference is that `methodStyle` is a question of taste. But `strictSetters` and `interfacesForAspects`, ... are recommended setting. We just don't change the default to avoid breaking existing code.

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  :::warning Limitations
+  This is supported only for the named aspects (inline targets are not supported) and does not respect all possible options how such entities might be exposed by services.
+  :::
+
+- [`betterNames`](/java/assets/cds-maven-plugin-site/generate-mojo.html#betterNames)
+
+  CDS models from external sources might include elements that have some special characters in their names or include elements that clash with Java keywords. Such cases always can be solved with the [renaming features](/java/cds-data#renaming-elements-in-java) provided by code generator, but in case of large models, this is tedious.
+  When this switch is enabled, characters `/` and `$` behave as a separators for the name during case conversions, similar to `_` and `.`. For example, `GET_MATERIAL` yields `GetMaterial` (or `getMaterial` for attributes and methods). The same now applies for the names with `/`, for example, name `/DMO/GET_MATERIAL` will be converted to `DmoGetMaterial`.
+  
+  The following conversions applied after:
```
agoerler:
```suggestion
  The following conversions are applied:  
```

?

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  When this switch is enabled, characters `/` and `$` behave as a separators for the name during case conversions, similar to `_` and `.`. For example, `GET_MATERIAL` yields `GetMaterial` (or `getMaterial` for attributes and methods). The same now applies for the names with `/`, for example, name `/DMO/GET_MATERIAL` will be converted to `DmoGetMaterial`.
+  
+  The following conversions applied after:  
+    - Names from CDS model that are Java keywords are suffixed with `_`.
+    - Names from CDS model that use characters that are not valid as Java identifiers, are replaced by `_`. This, however, might lead to a conflicts between names that yield the same name in Java.
+    - Leading `_` will remain in the name after conversions. This supports convention where association and its foreign key have names like `_assoc` and `assoc`.
+  These conversions no longer influence the splitting.
```
agoerler: This is maybe too technical!

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+:::
+
+See [Maven Plugin Documentation](/java/assets/cds-maven-plugin-site/generate-mojo.html) for actual status of deprecation and switches that are not described here. {.learn-more}
+
+### Annotation Detail Level
+
+Option [`annotationDetailLevel`](/java/assets/cds-maven-plugin-site/generate-mojo.html#annotationDetailLevel) lets you choose the amount of the details for the Java annotation [`@Generated`](https://docs.oracle.com/en/java/javase/21/docs/api/java.compiler/javax/annotation/processing/Generated.html) added to each interface. This annotation has no effect at runtime.
+
+Following levels of the details are available:
+- `MINIMAL` - only annotation is added, no additional information is added.
```
agoerler:
```suggestion
- `MINIMAL` (default) - only the annotation is added, no additional information is added.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+   ```java
+    @CdsName("service.Entity")
+    @Generated("cds-maven-plugin")
+    public interface Entity extends CdsData {  }
+   ```
+
+- `FULL` - annotation contains date and time of the generation.
```
agoerler:
```suggestion
- `FULL` - annotation contains the timestamp of the generation.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  @CdsName("service.Entity")
+  @Generated(
+      value = "cds-maven-plugin",
+      date = "9999-12-31T23:59:59.999999Z",
+      comments = ""
+  )
+  public interface Entity extends CdsData {  }
+  ```
+
+- `NONE` - annotation is not added at all. This is not recommended.
```
agoerler:
```suggestion
- `NONE` - no `@Generated` annotation is added. This is not recommended.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,151 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+:::warning Check migration guides!
+In major releases of CAP Java, some of these switches can be made new default and some other switches might be removed. This might introduce compile errors
+in your application that needs to be fixed.
+:::
+
+See [Maven Plugin Documentation](/java/assets/cds-maven-plugin-site/generate-mojo.html) for actual status of deprecation and switches that are not described here. {.learn-more}
+
+### Annotation Detail Level
+
+Option [`annotationDetailLevel`](/java/assets/cds-maven-plugin-site/generate-mojo.html#annotationDetailLevel) lets you choose the amount of the details for the Java annotation [`@Generated`](https://docs.oracle.com/en/java/javase/21/docs/api/java.compiler/javax/annotation/processing/Generated.html) added to each interface. This annotation has no effect at runtime.
```
agoerler:
```suggestion
The option [`annotationDetailLevel`](/java/assets/cds-maven-plugin-site/generate-mojo.html#annotationDetailLevel) lets you choose the amount of the details for the Java annotation [`@Generated`](https://docs.oracle.com/en/java/javase/21/docs/api/java.compiler/javax/annotation/processing/Generated.html) added to each interface. This annotation has no effect at runtime but is evaluated by static code analysis tools to identify the artifacts as generated.
```

java/developing-applications/building.md
vmikhailenko: Where exactly? 

Exclusion is a part of archetype. I can inverse the statement and write that inclusion of them is a bad idea.

java/developing-applications/building.md
vmikhailenko: @beckermarc ?

java/developing-applications/building.md
vmikhailenko: Better?

java/developing-applications/building.md
```diff
@@ -416,9 +417,8 @@ Options `includes` and `excludes` are simple filters. If included parts of your
 
 ### Style of Interfaces
 
-By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification.
+By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification. In this style getter and setter method names are prefixed with `get` and `set`, resp.:
```
agoerler:
```suggestion
By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification. In this style, getter and setter method names are prefixed with `get` and `set`, resp.:
```

agoerler (approved):

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+Each time your application is built, these interfaces are regenerated. By default, they are excluded from your version control. 
+
+### Package for Generated Code
+
+The option [`basePackage`](/java/assets/cds-maven-plugin-site/generate-mojo.html#basePackage) can be used to specify a base package prefix for generated code. The suffix package structure will reflect namespaces defined in your CDS model.
+
+### Filter for CDS Entities
+
+By default, the complete model of your application is generated including all imported or re-used models.
+You can use options [`includes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#includes) and [`excludes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#excludes) to specify the part of your overall model that is subject to code generation. Both inclusion and exclusion can be used together, inclusion evaluated first, then exclusion filter out included set of entities.
```
renejeglinsky: So exclusion trumps inclusion, right? If I include sth and at the same exclude it, it will be excluded?
```suggestion
You can use options [`includes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#includes) and [`excludes`](/java/assets/cds-maven-plugin-site/generate-mojo.html#excludes) to specify the part of your overall model that is subject to code generation. Both inclusion and exclusion can be used together, inclusion is evaluated first, then exclusion filters out of the included set of entities.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+These options use patterns that are applied on the fully qualified names of the entities in CDS models. For example, the pattern `my.bookshop.*` will cover all definitions with namespace `my.bookshop` and the pattern `my.bookshop.**` will cover all definitions with fully qualified name starting with `my.bookshop`.
+
+:::warning Cross-namespace references are not resolved
+Options `includes` and `excludes` are simple filters. If included parts of your model reference types from the excluded area, the resulting code will not compile.
+:::
+
+### Style of Interfaces
+
+By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification. In this style, getter and setter method names are prefixed with `get` and `set`, resp.:
```
renejeglinsky:
```suggestion
By default, the accessor interfaces provide the setter and getter methods inspired by the JavaBeans specification. In this style, getter and setter method names are prefixed with `get` and `set`:
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  For example, instead of the following:
+
+  ```java
+  void setManager(Map<String, ?> manager);
+  ```  
+  you get the following:
+
+  ```java
+  void setManager(Manager manager);
+  ```
```
renejeglinsky:
```suggestion
  ```java
  void setManager(Map<String, ?> manager); // [!code --]
  void setManager(Manager manager); // [!code ++]
  ```
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+Moreover, it doesn't change the way how event contexts, delivered by CAP, look like. Such interfaces from CAP are always modelled in the default JavaBeans style.
+
+### Code Generation Features
+
+Other options in this goal enable or disable certain features that change the way generated code looks in a certain aspect. These changes can be incompatible with the existing code and require manual adaptation.
+
+- [`strictSetters`](/java/assets/cds-maven-plugin-site/generate-mojo.html#strictSetters)
+
+  This switch changes the signature of the setter methods in typed access interfaces so that they require concrete type instead of generic `Map` interface.
+  For example, instead of the following:
```
renejeglinsky:
```suggestion
  For example:
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+  ```java
+  void setManager(Manager manager);
+  ```
+  It does not introduce any additional type checks at runtime, the correctness of the assignment is checked only at the time of compilation.
+
+- [`interfacesForAspects`](/java/assets/cds-maven-plugin-site/generate-mojo.html#interfacesForAspects)
+
+  If your entity is modelled with the [composition of aspects](/cds/cdl#with-named-targets), the generated interfaces always reference original aspect as type for setters and getters.
+  When this switch is enabled, code generator will use type generated by the compiler instead of type of the aspect itself and will include methods to fetch keys, for example.
```
renejeglinsky:
```suggestion
  When this switch is enabled, the code generator uses the type generated by the compiler instead of the type of the aspect itself and will include methods to fetch keys, for example.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+  CDS models from external sources might include elements that have some special characters in their names or include elements that clash with Java keywords. Such cases always can be solved with the [renaming features](/java/cds-data#renaming-elements-in-java) provided by code generator, but in case of large models, this is tedious.
+  When this switch is enabled, characters `/` and `$` behave as a separators for the name during case conversions, similar to `_` and `.`. For example, `GET_MATERIAL` yields `GetMaterial` (or `getMaterial` for attributes and methods). The same now applies for the names with `/`, for example, name `/DMO/GET_MATERIAL` will be converted to `DmoGetMaterial`.
+  
+  The following conversions are applied:  
+    - Names from CDS model that are Java keywords are suffixed with `_`.
+    - Names from CDS model that use characters that are not valid as Java identifiers, are replaced by `_`. This, however, might lead to a conflicts between names that yield the same name in Java.
+    - Leading `_` will remain in the name after conversions. This supports convention where association and its foreign key have names like `_assoc` and `assoc`.
```
renejeglinsky:
```suggestion
    - Leading `_` will remain in the name after conversions. This supports conventions where an association and its foreign key have names like `_assoc` and `assoc`.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+  These conversions no longer influence the splitting.
+
+- ['cqnServiceGetters'](/java/assets/cds-maven-plugin-site/generate-mojo.html#cqnServiceGetters)
+
+  The method `getService()` in generated [event-specific Event Context interfaces](../event-handlers/#eventcontext) is overridden to return the typed service interface instead of generic `Service` type.
```
renejeglinsky:
```suggestion
  The method `getService()` in generated [event-specific Event Context interfaces](../event-handlers/#eventcontext) is overridden to return the typed service interface instead of the generic `Service` type.
```

java/developing-applications/building.md
```diff
@@ -378,6 +378,150 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+:::warning Check migration guides!
+In major releases of CAP Java, some of these switches can be made new default and some other switches might be removed. This might introduce compile errors
```
renejeglinsky:
```suggestion
In major releases of CAP Java, some of these switches can be made the new default and some other switches might be removed. This might introduce compile errors
```

java/developing-applications/building.md
```diff
@@ -376,7 +376,148 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+
+```xml [pom.xml]
+<execution>
+    <id>cds.generate</id>
+    <goals>
+        <goal>generate</goal>
+    </goals>
+    <configuration>
+        <basePackage>cds.gen</basePackage>
+        ...
```
MattSchur, line 394:
```suggestion
        <linkedInterfaces>true</linkedInterfaces>
        ...
```

java/developing-applications/building.md
```diff
@@ -376,7 +376,148 @@ Use the _.cdsrc.json_ file to add project specific configuration of `@sap/cds-dk
...
+    </configuration>
+</execution>
+```
+
+Each time your application is built, these interfaces are regenerated. By default, they are excluded from your version control. 
+
```
MattSchur, line 400:
```suggestion

Set the `linkedInterfaces: true` option to link the generated query builder interfaces to the data accessor interfaces for queries with typed results (TODO link).

```

java/developing-applications/building.md
vmikhailenko: This switch should be documented below. I would rather add this in the dedicated PR.

vmikhailenko:
@renejeglinsky Can we merge this?

PR #1874, opened 2025-05-23
Include `cds add approuter` in MT guide again

tim-sh:
Adding App Router is mentioned explicitly, whereas the Cloud Portal facet that would include it is considered optional.

cf. [this comment](https://github.com/cap-js/docs/commit/eab3b5d50c26882c615ea07fe4ac9401164c4dc3#commitcomment-157837891) on the original commit.

These changes concern the Multitenancy Guide, featuring an excerpt from the Deployment Guide.

The [Deployment Guide](https://cap.cloud.sap/docs/guides/deployment/to-cf#add-app-router), in turn, also mentions adding App Router as a separate item, followed by Cloud Portal as only one of two options.


swaldmann (commented):
Unfortunately the Cloud Portal service is considered deprecated and Work Zone doesn't use its own approuter but a managed one. Therefore I don't want to promote hosting your own approuter too much as you shouldn't need it in the MT-only case.

renejeglinsky:
Closing this due to inactivity. Please reopen if it's still needed.

PR #1876, opened 2025-05-26
fix: switch code automatically where possible by using impl class

tim-sh:


renejeglinsky:
Hi Tim,

Thanks a lot for your effort. Actually we're in the process of reducing the usage of toggles in capire. So we won't add any new ones. Sorry.

PR #1891, opened 2025-06-02
Release microservices guide

Akatuoro:


DanSchlachter (approved):

guides/deployment/microservices.md
```diff
@@ -906,7 +907,6 @@ resources:
 
 #### Subsequent updates
 
-- TODO...
```
renejeglinsky, line 909: What is missing here before publication?

Akatuoro: Nothing really. The important part is to ensure `shared-db` is deployed before the others, so that the db model is compatible once the apps are deployed.

I've tidied up the section a bit.

renejeglinsky (approved):

PR #1911, opened 2025-06-10
microservices - add missing directory

vl-leon:


Akatuoro (approved):

guides/deployment/microservices.md
```diff
@@ -155,10 +155,10 @@ These are the (not so beneficial) side effects you when using a shared persisten
...
 ```sh
-cds db -2 sql
+cds shared-db/db -2 sql
 ```
 ```sh
-cds db -2 hana
+cds shared-db/db -2 hana
 ```
 
 ```sh
```
renejeglinsky: Does `cds deploy` and `cds build` work w/o being in the folder?
Or would it make sense to include a `cd shared-db` step and not mention the folder with every command?

Akatuoro: Good point, for the context of the 'Try it out' box a `cd shared-db` step could make sense.
`cds build shared-db --for hana` works, but `cds deploy` only works in the folder.

vl-leon: Changing directories can be confusing, a subsequent  `cd ..` is also required, in addition the suggested change shows that the cds command in that case consumes a model and the `db` parameter is not a command but a directory with a cds model. Afterwards the guide continues in the root directory so better make sure we stay there.

renejeglinsky: In line 165 what does `cds deploy` deploy? This would then also need the "shared-db" as part of the command, or?
image

Same question for line 168.

vl-leon: We can use the cds **compile** command to make it more clear what happens,. It seems to be the default command, thus skipped. 
```
cds compile shared-db/db --to hana
cds compile shared-db/db --to sql
````

Comment (vl-leon) [reply guides/deployment/microservices.md]: > In line 165 what does `cds deploy` deploy? This would then also need the "shared-db" as part of the command, or? image
> 
> Same question for line 168.

You are right, we need the models here also:

```
npx cds deploy -2 sqlite
No .cds models found, please specify one as an argument

```

vl-leon: just uploaded the "cd shared-db" variant for the whole block - commands work now

renejeglinsky (approved):

PR #1913, opened 2025-06-10
nodejs: document disabling of fuzzy search

johannes-vogel:


guides/providing-services.md
```diff
@@ -441,8 +441,9 @@ Fuzzy search is a fault-tolerant search feature of SAP HANA Cloud, which returns
 You can configure the fuzziness in the range [0.0, 1.0]. The value 1.0 enforces exact search.
 
 - Java: <Config java keyOnly>cds.sql.hana.search.fuzzinessThreshold = 0.8</Config>
-- Node.js:<Config keyOnly>cds.hana.fuzzy = 0.7</Config>
+- Node.js:<Config keyOnly>cds.hana.fuzzy = 0.7</Config><sup>(1)</sup> 
 
+<sup>(1)</sup> If set to `false`, fuzzy search is disabled and falls back to a case insensitive substring search.
```
renejeglinsky, line 446: Where is the difference between 0.0 and false?

renejeglinsky: Seems as if false isn't possible for Java, correct?

johannes-vogel: > Seems as if false isn't possible for Java, correct?

Java has mentioned an additional config above. only in hex mode, fuzzy search is possible. Node always uses hex mode.

> Where is the difference between 0.0 and false?

0.0 will still use the hana fuzzy search (even though I don't know if it makes sense). we explicitly check for value `false` to not use hana fuzzy search but a self implemented case insensitive substring search using the like operator.

renejeglinsky: Thanks for the explanation!

PR #1916, opened 2025-06-11
Java: Views and Projections

MattSchur:
Introduces "Views and Projections" section with
- Runtime Views
- Write through Views

and documents some rules for write via path expressions in views.

Also shortens some headers to improve the readability in the navigation bar.

### DIFF
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -385,7 +385,7 @@ That means:
 -> [Draft-enabled](../fiori-drafts#reading-drafts) CDS views must fulfill all requirements of [updatable views](#updatable-views).
 {.indent}
 
-Remember to run draft specific queries through the [Draft Service](../fiori-drafts#draft-service) or [Application Service](../cqn-services/application-services#application-services). The [Persistence Service](../cqn-services/persistence-services) only works for non-draft specific queries.
+Remember to run draft specific queries through the [Draft Service](../fiori-drafts#draftservices) or [Application Service](../cqn-services/application-services#application-services). The [Persistence Service](../cqn-services/persistence-services) only works for non-draft specific queries.
 
 
 ::: tip Draft Queries on Runtime Views
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -227,20 +227,22 @@ long deleteCount = service.run(delete).rowCount();
 
 With CDS [views](../../cds/cdl#views-projections) you can derive new entities from existing ones, for example to rename or exclude certain elements, or to add [virtual elements](../../cds/cdl#virtual-elements-in-views) for specific use cases.
 
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
+
 The CDS compiler generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files from your CDS model, including SQL views for the CDS views. These views are deployed to the [database](../cqn-services/persistence-services#database-support) and used by the CAP runtime to read data.
 
 For *read-only* views, you can use the full feature set of [selects](../../cds/cdl#as-select-from), including *joins*, *unions*, and *aggregations*. However, such complex views are not writable and require a schema redeployment if the view definition is changed.
 
-To [write data](#updatable-views) or [delete](#delete-via-view) through views, only use simple [projections](../../cds/cdl#as-projection-on). The CAP Java runtime attempts to resolve the CDS views to their underlying persistence entities, rewriting the statement and data accordingly, which is not supported for complex views.
-
-For simple projections the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
-
 ::: tip Indicate read-only
 Use the `@readonly` annotation to indicate that a view or a view element is not writable.
 :::
-::: tip Prefer simple views
-Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
-:::
+
+To [write data](#updatable-views) or [delete](#delete-via-view) through views, only use simple [projections](../../cds/cdl#as-projection-on). The CAP Java runtime attempts to resolve the CDS views to their underlying persistence entities, rewriting the statement and data accordingly, which is not supported for complex views.
+
+For simple projections, the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
+
 ::: warning Avoid selecting to-many Associations
 Do not use [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
 :::
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -117,12 +117,12 @@ var params = Map.of("minStock", 100);
 Result result = service.run(query, params);
 ```
 
-#### Adding Query Hints for SAP HANA { #hana-hints}
+### Query Hints { #hana-hints}
 
 To add a hint clause to a statement, use the `hints` method and prefix the [SAP HANA hints](https://help.sap.com/docs/HANA_CLOUD_DATABASE/c1d3f60099654ecfb3fe36ac93c121bb/4ba9edce1f2347a0b9fcda99879c17a1.htmlS) with `hdb.`:
 
 ```java
-CqnSelect query = Select.from(BOOKS).hints("hdb.USE_HEX_PLAN", "hdb.ESTIMATION_SAMPLES(0)");
+Select.from(BOOKS).hints("hdb.USE_HEX_PLAN", "hdb.ESTIMATION_SAMPLES(0)");
 ```
 ::: warning
 Hints prefixed with `hdb.` are directly rendered into SQL for SAP HANA and therefore **must not** contain external input!
@@ -160,7 +160,7 @@ If no rows are touched the execution is successful but the row count is 0.
 The setters of an [update with expressions](../working-with-cql/query-api#update-expressions) are evaluated on the database. The result of these expressions is not contained in the update result.
 :::
 
-### Working with Structured Documents
+### Structured Documents
 
 It's possible to work with structured data as the insert, update, and delete operations cascade along *compositions*.
 
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
 long deleteCount = service.run(delete).rowCount();
 ```
 
-### Resolvable Views and Projections { #updatable-views}
+## Views and Projections { #views }
 
-The CAP Java SDK aims to resolve statements on non-complex views and projections to their underlying entity. When delegating queries between Application Services and Remote Services, statements are resolved to the entity definitions of the targeted service. Using the Persistence Service, only modifying statements are resolved before executing database queries. This allows to execute [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) operations on database views. For [Select](./query-api#select) statements database views are always leveraged, if available.
+The CDS compiler automatically generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files from your CDS model, including database views for all CDS [views and projections](../../cds/cdl#views-projections).
 
-Views and projections can be resolved if the following conditions are met:
+These views are used to read data from the database, while write operations are resolved to the underlying entities, if possible.
 
-- The view definition does not use any other clause than `columns` and `excluding`.
-- The projection includes all key elements; with the exception of insert operations with generated UUID keys.
-- The projection includes all elements with a `not null` constraint, unless they have a default value.
-- The projection must not include calculated fields when running queries against a remote OData service.
-- The projection must not include [path expressions](../../cds/cql#path-expressions) using to-many associations.
+::: warning
+Avoid using to-many associations on the select list of CDS views, as this prevents write operations and can lead to performance issues due to duplicating records on the DB.
+:::
 
-For [Insert](./query-api#insert) or [Update](./query-api#update) operations, if the projection contains functions or expressions, these values are ignored. Path expressions navigating *to-one* associations, can be used in projections as shown by the `Header` view in the following example. The `Header` view includes the element `country` from the associated entity `Address`.
+### Runtime Views { #runtimeviews }
+
+To add or update CDS views without redeploying the database schema, annotate your views with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This tells the CDS compiler to skip generating database views for these entities. Instead, the CAP Java runtime dynamically resolves these views at runtime.
+
+::: warning Limitations
+Runtime views do not support complex views that use aggregations, unions, joins, or subqueries in the `FROM` clause.
+
+To read [draft-enabled](../fiori-drafts#reading-drafts) entities, set `cds.drafts.persistence` to `split`.
+:::
+
+For example, consider the following CDS model and query:
 
 ```cds
-// Supported
-entity Order as projection on bookshop.Order;
-entity Order as projection on bookshop.Order { ID, status as state };
-entity Order as projection on bookshop.Order excluding { status };
-entity Header as projection on bookshop.OrderHeader { key ID, address.country as country };
+entity Books {
+  key id     : UUID;
+      title  : String;
+      stock  : Integer;
+      author : Association to one Authors;
+}
+@cds.persistence.skip
+entity BooksWithLowStock as projection on Books {
+    id, title, author.name as author
+} where stock < 10;
+```
+```sql
+Select BooksWithLowStock where author = 'Kafka'
 ```
 
-If a view is too complex to be resolved by the CDS runtime, the statement remains unmodified. Views that cannot be resolved by the CDS runtime include the use of `join`, `union` and the `where` clause.
+CAP Java provides two modes for resolving runtime views:
+
+**`cte` mode**: The runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database.
+
+```sql
+WITH BOOKSWITHLOWSTOCK_CTE AS (
+    SELECT B.ID,
+           B.TITLE,
+           A.NAME AS "AUTHOR"
+      FROM BOOKS B
+      LEFT OUTER JOIN AUTHOR A ON B.AUTHOR_ID = A.ID
+     WHERE B.STOCK < 10
+)
+SELECT ID, TITLE, AUTHOR AS "author"
+  FROM BOOKSWITHLOWSTOCK_CTE
+ WHERE A.NAME = ?
+```
+
+::: tip
+In CAP Java 3.10, enable `cte` mode with **cds.sql.runtimeView.mode: cte**
+:::
+
+**`resolve` mode**: The runtime _resolves_ the view definition to the underlying persistence entities and executes the query directly against them.
+
+```sql
+SELECT B.ID, B.TITLE, A.NAME AS "author"
+  FROM BOOKS AS B
+  LEFT OUTER JOIN AUTHORS AS A ON B.AUTHOR_ID = A.ID
+ WHERE B.STOCK < 10 AND A.NAME = ?
+```
+
+::: warning Limitations of `resolve` mode
+Expands to other runtime views and complex draft queries are not supported.
+:::
+
+
+### Write through Views { #updatable-views}
+
+To run [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) statements on CDS views, the CAP Java runtime resolves them to the underlying entity definitions, if possible - similar to the runtime view *resolve* mode for queries.
+
+When delegating queries between Application Services and Remote Services, statements are also resolved to the entity definitions of the targeted service.
+
+Views and projections can be resolved for write operations if the following conditions are met:
+
+- The view definition does not use any other clause than `columns` and `excluding` - `join`, `union` and `where` are not supported.
+- The projection includes all key elements; with the exception of insert operations with generated UUID keys.
+- The projection includes all elements with a `not null` constraint, unless they have a default value.
+- The projection must not include [path expressions](../../cds/cql#path-expressions) using to-many associations.
+- Projections targeting remote OData services must not include calculated elements.
+
+For [Insert](./query-api#insert) or [Update](./query-api#update) operations, if the projection contains functions or expressions, these values are ignored. 
+
+Path expressions navigating *to-one* compositions, can be used in projections as shown by `headerStatus` in the following example, which includes the element `status` of the `OrderHeader` entity.
+
+```cds
+entity Order as projection on bookshop.Order;
+entity Order as projection on bookshop.Order { ID, status as state };
+
+entity Order as projection on bookshop.Order {
+  ID,
+  header.status        as headerStatus,
+  header.customer.name as customerName @readonly
+} excluding { status };
+```
+
+::: warning Path Expressions using Associations are readonly
+Path expressions navigating associations are by default [not writable](#cascading-over-associations) and need to be annotated with [@readonly](../../guides/providing-services#readonly).
+:::
+
+Write operations on views that cannot be resolved by the CAP Java runtime remain unmodified
 - For the Persistence Service, this means the runtime _attempts_ to execute the write operation on the database view. Whether this execution is possible is [database dependent](../cqn-services/persistence-services#database-support).
 - For Application Services and Remote Services, the targeted service will reject the statement.
 
-Example of a view that can't be resolved:
-
-```cds
-// Unsupported
-entity DeliveredOrders as select from bookshop.Order where status = 'delivered';
-entity Orders as select from bookshop.Order inner join bookshop.OrderHeader on Order.header.ID = OrderHeader.ID { Order.ID, Order.items, OrderHeader.status };
-```
 
 ## Concurrency Control
 
 Concurrency control allows protecting your data against unexpected concurrent changes.
 
-### Optimistic Concurrency Control {#optimistic}
+### Optimistic Locking {#optimistic}
 
 Use _optimistic_ concurrency control to detect concurrent modification of data _across requests_. The implementation relies on an _ETag_, which changes whenever an entity instance is updated. Typically, the ETag value is stored in an element of the entity.
 
@@ -400,64 +478,6 @@ The `lock()` method has an optional parameter `timeout` that indicates the maxim
 
 The parameter `mode` allows to specify whether an `EXCLUSIVE` or a `SHARED` lock should be set.
 
-## Runtime Views { #runtimeviews }
-
-The CDS compiler generates [SQL DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) statements from your CDS model, including SQL views for all CDS [views and projections](../../cds/cdl#views-projections). As a result, adding or modifying CDS views typically requires redeploying the database schema.
-
-To avoid schema redeployments when you add or update CDS views, annotate them with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This annotation tells the CDS compiler to skip generating database views for these entities. Instead, the CAP Java runtime dynamically resolves such views at runtime.
-
-::: warning Limitations
-Runtime views support only simple [CDS projections](../../cds/cdl#as-projection-on). They do not support complex views that use aggregations, unions, joins, or subqueries in the `FROM` clause. To read [draft-enabled](../fiori-drafts#reading-drafts) entities, set `cds.drafts.persistence` to `split`. [Calculated elements](../../cds/cdl#calculated-elements) are not yet supported in runtime views.
-:::
-
-For example, consider the following CDS model and query:
-
-```cds
-entity Books {
-  key id     : UUID;
-      title  : String;
-      stock  : Integer;
-      author : Association to one Authors;
-}
-@cds.persistence.skip
-entity BooksWithLowStock as projection on Books {
-    id, title, author.name as author
-} where stock < 10;
-```
-```sql
-Select BooksWithLowStock where author = 'Kafka'
-```
-
-CAP Java provides two modes for resolving runtime views:
-
-**`cte` mode**: The runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database.
-
-```sql
-WITH BOOKSWITHLOWSTOCK_CTE AS (
-    SELECT B.ID,
-           B.TITLE,
-           A.NAME AS "AUTHOR"
-      FROM BOOKS B
-      LEFT OUTER JOIN AUTHOR A ON B.AUTHOR_ID = A.ID
-     WHERE B.STOCK < 10
-)
-SELECT ID, TITLE, AUTHOR AS "author"
-  FROM BOOKSWITHLOWSTOCK_CTE
- WHERE A.NAME = ?
-```
-
-::: tip
-CAP Java 4.x uses `cte` mode by default. In 3.10, enable it with **cds.sql.runtimeView.mode: cte**.
-:::
-
-**`resolve` mode**: The runtime _resolves_ the view definition to the underlying persistence entities and executes the query directly against them.
-
-```sql
-SELECT B.ID, B.TITLE, A.NAME AS "author"
-  FROM BOOKS AS B
-  LEFT OUTER JOIN AUTHORS AS A ON B.AUTHOR_ID = A.ID
- WHERE B.STOCK < 10 AND A.NAME = ?
-```
 
 ## Using I/O Streams in Queries
 
@@ -602,7 +622,7 @@ Map<String, String> titleToDescription =
 For the entities defined in the data model, CAP Java SDK can generate interfaces for you through [a Maven plugin](../cqn-services/persistence-services#staticmodel).
 
 
-### Using Entity References from Result Rows in CDS QL Statements {#entity-refs}
+### Entity References {#entity-refs}
 
 For result rows that contain all key values of an entity, you get an [entity reference](./query-api#entity-refs) via the `ref()` method. This reference addresses the entity via the key values from the result row.
 
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -229,6 +229,9 @@ The CDS compiler generates [DDL](../../guides/databases?impl-variant=java#genera
 
 Entity tables and views are deployed to the database. Read statements on CDS views target the database view, while write operations are resolved to the underlying entity table by the CAP Java runtime, when possible.
 
+::: tip Avoid mixin compositions
+Avoid introducing new compositions in CDS views and prefer associations instead, as [deep write](#updatable-views) and [cascading delete](#delete-via-view) are only supported for compositions in persistence entities.
+:::
 ::: warning Avoid to-many associations
 Do not use to-many associations in the select clause of CDS views. This blocks write operations and can cause performance issues due to record duplication on read.
 :::
@@ -345,11 +348,29 @@ If the CAP Java runtime cannot resolve a view, write operations are either rejec
 
 ### Delete through Views { #delete-via-view }
 
-[Delete](./query-api#delete) operations on CDS views are resolved to the underlying entity definitions by the CAP Java runtime, if possible. Delete via CDS views using *join*, *union*, or *where* is not supported.
+The CAP Java runtime attempts to resolve [Delete](./query-api#delete) operations to the underlying entity definitions—similar to the [runtime view](#runtimeviews) *resolve* mode for queries.
 
 ::: warning Cascading Delete is applied on persistence entity level
 Compositions that are added, changed or removed in CDS views are not considered by [cascading delete](./query-execution#cascading-delete).
 :::
+
+For example, the following CDS model defines `Order` with header and items, and `OrderView` which excludes header and exposes items as `lineItems`:
+
+```cds
+entity Order : cuid, managed {
+      header : Composition of one OrderHeader;
+      items  : Composition of many OrderItem on items.parent = $self;
+}
+entity OrderView as projection on db.Order {
+  ID,
+  items as lineItems,
+  delivery : Composition of one Delivery on delivery.order = $self
+}
+```
+```sql
+DELETE from OrderView where ID = 42
+```
+The delete operation is resolved to the underlying `Order` entity with ID *42* and cascades over the `header` and `items` compositions. The `delivery` composition, which is only defined in the view, is ignored and does not cascade the delete operation to `Delivery`.
 
 
 ## Concurrency Control
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -229,7 +229,7 @@ With CDS [views](../../cds/cdl#views-projections) you can derive new entities fr
 
 From the CDS model the CDS compiler generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files, which include SQL views for the CDS views. These views are deployed to the [database](../cqn-services/persistence-services#database-support) and used by the CAP runtime to read data.
 
-For *read-only* views, you can use the full feature set of [selects](../../cds/cdl#as-select-from), including *joins*, *unions*, and *aggregations*. However, such complex views are not writable and require a schema redeployment if the view definition is changed.
+For *read-only* views, you can use the full feature set of [selects](../../cds/cdl#as-select-from), including *aggregations* to summarize, as well as *joins* and *unions* to combine data from multiple entities. However, such complex views are *not writable* and require a schema redeployment if the view definition is changed.
 
 ::: warning Annotate read-only views and elements
 Use the `@readonly` annotation to indicate that a view or a view element is not writable, as this is not automatically detected by the CDS compiler.
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -264,7 +264,9 @@ SELECT from BooksWithLowStock where author = 'Kafka'
 
 CAP Java supports two modes for resolving runtime views:
 
-**`cte` mode**: The runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database.
+#### CTE mode
+
+This is the default mode in CAP Java `4.x`. The runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database.
 
 ```sql
 WITH BOOKSWITHLOWSTOCK_CTE AS (
@@ -280,11 +282,13 @@ SELECT ID, TITLE, AUTHOR AS "author"
  WHERE A.NAME = ?
 ```
 
-::: tip
-In CAP Java 3.10, enable **cte** mode with *cds.sql.runtimeView.mode: cte*
+::: tip CAP Java 3.10
+Enable *cte* mode with *cds.sql.runtimeView.mode: cte*
 :::
 
-**`resolve` mode**: The runtime _resolves_ the view definition to the underlying persistence entities and executes the query directly against the corresponding tables.
+#### Resolve mode
+
+The runtime _resolves_ the view definition to the underlying persistence entities and executes the query directly against the corresponding tables.
 
 ```sql
 SELECT B.ID, B.TITLE, A.NAME AS "author"
@@ -309,7 +313,7 @@ If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) e
 
 ### Write through Views { #updatable-views }
 
-You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert), and [Update](./query-api#update) statements on CDS views. The CAP Java runtime attempts to resolve these to the underlying entity definitions—similar to the [runtime view](#runtimeviews) *resolve* mode for queries.
+You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert), and [Update](./query-api#update) statements on CDS views. The CAP Java runtime attempts to resolve these to the underlying entity definitions—similar to the runtime view [resolve mode](query-execution#resolve-mode) for queries.
 
 When delegating queries between Application Services and Remote Services, statements are also resolved to the targeted service's entity definitions.
 
@@ -348,7 +352,7 @@ If the CAP Java runtime cannot resolve a view, write operations are either rejec
 
 ### Delete through Views { #delete-via-view }
 
-The CAP Java runtime attempts to resolve [Delete](./query-api#delete) operations to the underlying entity definitions—similar to the [runtime view](#runtimeviews) *resolve* mode for queries.
+The CAP Java runtime attempts to resolve [Delete](./query-api#delete) operations to the underlying entity definitions—similar to the runtime view [resolve mode](query-execution#resolve-mode) for queries.
 
 ::: warning Cascading Delete is applied on persistence entity level
 Compositions that are added, changed or removed in CDS views are not considered by [cascading delete](./query-execution#cascading-delete).
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -325,7 +325,7 @@ Runtime views must be simple [projections](../../cds/cdl#as-projection-on), not 
 
 CAP Java provides two modes for resolving runtime views during read operations: [cte](#rtview-cte) and [resolve](#rtview-resolve). 
 
-::: details Set the runtime view mode
+::: details Changing the runtime view mode
 To globally set the runtime view mode, use the property `cds.sql.runtimeView.mode` with value `cte` or `resolve` in the *application.yml*. To set the mode for a specific runtime view, annotate it with `@cds.java.runtimeView.mode: cte|resolve`.
 
 To set the mode for a specific query, use a [hint](#hana-hints):
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -332,7 +332,7 @@ entity Order as projection on bookshop.Order {
 };
 ```
 
-[Deep write](./query-execution#deep-insert-upsert) via (aliased) compositions is supported if there are corresponding compositions in the underlying entity definition. Deep write via compositions only defined in the view (mixins) is not supported and the data for these composition targets is ignored. Values for elements corresponding to *expressions* and *functions* is ignored as well.
+[Deep write](./query-execution#deep-insert-upsert) via (aliased) compositions is supported if there are corresponding compositions in the underlying entity definition. Deep write via compositions that are only defined in the view (mixins) is not supported and the data for such mixin compositions is ignored. Data for elements corresponding to *expressions* and *functions* is ignored as well.
 
 ::: warning Path Expressions
 Path expressions navigating *associations* (e.g. *header.customer.name*) are [not writable](#cascading-over-associations) by default. To avoid issues on write, annotate them with [@readonly](../../guides/providing-services#readonly).
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -301,39 +301,45 @@ Expands to other runtime views and complex draft queries are not supported.
 
 ### Write through Views { #updatable-views }
 
-You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) statements on CDS views. If possible, the CAP Java runtime resolves these to the underlying entity definitions—similar to the [runtime view](#runtimeviews) *resolve* mode for queries.
+You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert), and [Update](./query-api#update) statements on CDS views. If possible, the CAP Java runtime resolves these to the underlying entity definitions—similar to the [runtime view](#runtimeviews) *resolve* mode for queries.
 
 When delegating queries between Application Services and Remote Services, statements are also resolved to the targeted service's entity definitions.
 
 Write operations on CDS views are supported if:
 
-- The view definition uses only `columns` and `excluding` (no `join`, `union`, or `where`).
-- The projection includes all key elements (except for insert operations with generated UUID keys).
-- The projection includes all `not null` elements, unless they have a default value.
+- The view definition uses only *columns* and *excluding* (no *join*, *union*, or *where*).
+- The projection includes all *not null* elements (incl. keys), unless they have a default or generated value.
 - The projection does not include [path expressions](../../cds/cql#path-expressions) using *to-many* associations.
-- Projections targeting remote OData services do not include calculated elements.
+- Projections targeting *remote OData* services must not include calculated elements.
 
-For [Insert](./query-api#insert) or [Update](./query-api#update), values for functions and expressions in the projection are ignored.
-
-You can use path expressions navigating *to-one* compositions, as shown by `headerStatus` in this example:
+You can use path expressions navigating *to-one* compositions, as shown by `headerStatus`:
 
 ```cds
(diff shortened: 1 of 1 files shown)


java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
 
-The CAP Java SDK aims to resolve statements on non-complex views and projections to their underlying entity. When delegating queries between Application Services and Remote Services, statements are resolved to the entity definitions of the targeted service. Using the Persistence Service, only modifying statements are resolved before executing database queries. This allows to execute [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) operations on database views. For [Select](./query-api#select) statements database views are always leveraged, if available.
+The CDS compiler automatically generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files from your CDS model, including database views for all CDS [views and projections](../../cds/cdl#views-projections).
 
-Views and projections can be resolved if the following conditions are met:
+These views are used to read data from the database, while write operations are resolved to the underlying entities, if possible.
 
-- The view definition does not use any other clause than `columns` and `excluding`.
+::: warning
```
MattSchur:
```suggestion
::: warning Avoid to-many associations
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+entity Order as projection on bookshop.Order {
+  ID,
+  header.status        as headerStatus,
+  header.customer.name as customerName @readonly
+} excluding { status };
 ```
 
-If a view is too complex to be resolved by the CDS runtime, the statement remains unmodified. Views that cannot be resolved by the CDS runtime include the use of `join`, `union` and the `where` clause.
+::: warning Path Expressions using Associations are readonly
```
MattSchur:
```suggestion
::: warning Path Expressions using associations are readonly
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+:::
+
+### Runtime Views { #runtimeviews }
+
+To add or update CDS views without redeploying the database schema, annotate your views with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This tells the CDS compiler to skip generating database views for these entities. Instead, the CAP Java runtime dynamically resolves these views at runtime.
+
+::: warning Limitations
+Runtime views do not support complex views that use aggregations, unions, joins, or subqueries in the `FROM` clause.
+
+To read [draft-enabled](../fiori-drafts#reading-drafts) entities, set `cds.drafts.persistence` to `split`.
```
vmikhailenko: Is this rather the precondition to use them?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+@cds.persistence.skip
+entity BooksWithLowStock as projection on Books {
+    id, title, author.name as author
+} where stock < 10;
+```
+```sql
+Select BooksWithLowStock where author = 'Kafka'
+```
+
+CAP Java provides two modes for resolving runtime views:
```
vmikhailenko: How developers should decide which mode to use?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+When delegating queries between Application Services and Remote Services, statements are also resolved to the entity definitions of the targeted service.
+
+Views and projections can be resolved for write operations if the following conditions are met:
+
+- The view definition does not use any other clause than `columns` and `excluding` - `join`, `union` and `where` are not supported.
 - The projection includes all key elements; with the exception of insert operations with generated UUID keys.
 - The projection includes all elements with a `not null` constraint, unless they have a default value.
-- The projection must not include calculated fields when running queries against a remote OData service.
 - The projection must not include [path expressions](../../cds/cql#path-expressions) using to-many associations.
+- Projections targeting remote OData services must not include calculated elements.
```
vmikhailenko: Can view include mixin associations or compositions? What will be writable then?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+### Write through Views { #updatable-views}
+
+To run [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) statements on CDS views, the CAP Java runtime resolves them to the underlying entity definitions, if possible - similar to the runtime view *resolve* mode for queries.
+
+When delegating queries between Application Services and Remote Services, statements are also resolved to the entity definitions of the targeted service.
+
+Views and projections can be resolved for write operations if the following conditions are met:
+
+- The view definition does not use any other clause than `columns` and `excluding` - `join`, `union` and `where` are not supported.
```
vmikhailenko: In the example above `where` is used.

java/working-with-cql/query-execution.md
vmikhailenko: Would be nice to explain that drafts are not runtime views and still remain the plain tables. This confused some stakeholders: e.g. that value from path expression is writable through draft and not writable through view.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,123 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+  header.customer.name as customerName @readonly
+} excluding { status };
 ```
 
-If a view is too complex to be resolved by the CDS runtime, the statement remains unmodified. Views that cannot be resolved by the CDS runtime include the use of `join`, `union` and the `where` clause.
+::: warning Path Expressions using Associations are readonly
+Path expressions navigating associations are by default [not writable](#cascading-over-associations) and need to be annotated with [@readonly](../../guides/providing-services#readonly).
+:::
+
+Write operations on views that cannot be resolved by the CAP Java runtime remain unmodified
```
vmikhailenko: Maybe also cross-reference that drafts must be handled through an application service. Just to make people aware of this one more time.

java/working-with-cql/query-execution.md
vmikhailenko: Also, these restrictions at least partially apply to the views in general, right? Are they described somewhere?

java/working-with-cql/query-execution.md
MattSchur: I move this into a separate "tip" box

java/working-with-cql/query-execution.md
MattSchur: We can add recommendations later with experience from app development. Currently `cte` mode is the new default in `4.x` with less limitations.

java/working-with-cql/query-execution.md
MattSchur: `where` is supported in runtime views (on read), but currently not on write. The "Write through Views" section is not specific to runtime views but to views in general.

java/working-with-cql/query-execution.md
MattSchur: I added a section on deep write via views

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,139 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+```
+
+::: tip
+In CAP Java 3.10, enable `cte` mode with **cds.sql.runtimeView.mode: cte**
+:::
+
+::: warning Draft-enabling runtime views requires schema update
+When [draft-enabling](../fiori-drafts#reading-drafts) a runtime view, a corresponding draft persistence table is created for the CDS view and a database schema update is required when changing the runtime view. 
+
+[Draft activate](../fiori-drafts#editing-drafts) triggers a [write through view](#updatable-views) and the restrictions apply.
```
vmikhailenko: Restrictions of write-through? 

I would rather gather draft-related part in the separate subsection so that later it can be referenced.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,139 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+::: warning Avoid to-many associations
+Do not use to-many associations in the select clause of CDS views. This blocks write operations and can cause performance issues due to record duplication.
+:::
+
+### Runtime Views { #runtimeviews }
+
+To add or update CDS views without redeploying the database schema, annotate your views with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This tells the CDS compiler to skip generating database views for these entities, and the CAP Java runtime resolves them dynamically at runtime.
+
+::: warning Limitations
+Runtime views do not support aggregations, unions, joins, or subqueries in the FROM clause.
```
vmikhailenko: Not sure if we need that many warning everywhere.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,139 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+entity Order as projection on bookshop.Order {
+  key ID,
+      header.status        as headerStatus,
+      header.customer.name as customerName @readonly,
+      items                as lineItems // aliased composition
+};
 ```
 
+[Deep write](./query-execution#deep-insert-upsert) via (aliased) compositions is supported if there are corresponding compositions in the underlying entity definition. Deep write via compositions that are only defined in the view (mixins) is not supported and the data for such mixin compositions is ignored. Data for elements corresponding to *expressions* and *functions* is ignored as well.
```
vmikhailenko: >Data for elements corresponding to *expressions* and *functions* is ignored as well.

Sample? 

Also, why not merge this in the list above? It can be nicer to have them listed as a reference.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,139 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+SELECT B.ID, B.TITLE, A.NAME AS "author"
+  FROM BOOKS AS B
+  LEFT OUTER JOIN AUTHORS AS A ON B.AUTHOR_ID = A.ID
+ WHERE B.STOCK < 10 AND A.NAME = ?
+```
+
+::: warning Limitations of `resolve` mode
+Using associations introduced by the view (mixins), as well as complex draft queries, and [draft-enabling](../fiori-drafts#reading-drafts) runtime views are not supported in *resolve* mode.
+:::
+
```
vmikhailenko, line 394: Maybe also add that runtime views are not lockable. See https://cap.cloud.sap/docs/java/working-with-cql/query-api#write-lock

java/working-with-cql/query-execution.md
MattSchur: Restrictions of write-through -> Draft activate deletes the draft via the draft table and writes/updates the active entity via the view, which is a write through view and only works for views that support write operations.

java/working-with-cql/query-execution.md
MattSchur: Pessimistic locking is possible if the runtime view can be resolved or the view (generated CTE) fulfills the conditions mentioned in [locking](https://cap.cloud.sap/docs/java/working-with-cql/query-api#write-lock), which should be the case since we don't support union, join and aggregations in runtime views. It locks the underlying table rows.

java/working-with-cql/query-execution.md
MattSchur: I'll change this to an info box

java/working-with-cql/query-execution.md
MattSchur: The list above are the requirements for views to be updatable. Here we describe that some data is ignored on write.

java/working-with-cql/query-execution.md
MattSchur: Added a separate section

java/working-with-cql/query-execution.md
vmikhailenko: On HANA? Wow.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
 
-Views and projections can be resolved if the following conditions are met:
+Entity tables and views are deployed to the database. Read statements on CDS views target the database view, while write operations are resolved to the underlying entity table by the CAP Java runtime, when possible.
 
-- The view definition does not use any other clause than `columns` and `excluding`.
-- The projection includes all key elements; with the exception of insert operations with generated UUID keys.
-- The projection includes all elements with a `not null` constraint, unless they have a default value.
-- The projection must not include calculated fields when running queries against a remote OData service.
-- The projection must not include [path expressions](../../cds/cql#path-expressions) using to-many associations.
+::: tip Avoid mixin compositions
```
vmikhailenko: Maybe a red box? If this leads to a data loss or garbage being left, then it is nasty.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+      FROM BOOKS B
+      LEFT OUTER JOIN AUTHOR A ON B.AUTHOR_ID = A.ID
+     WHERE B.STOCK < 10
+)
+SELECT ID, TITLE, AUTHOR AS "author"
+  FROM BOOKSWITHLOWSTOCK_CTE
+ WHERE A.NAME = ?
+```
+
+::: tip CAP Java 3.10
```
vmikhailenko, line 373: Should we really document outdated stuff?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+::: info Limitations of `resolve` mode
+Using associations introduced by the view (mixins), as well as complex draft queries, and [draft-enabling](../fiori-drafts#reading-drafts) runtime views are not supported in *resolve* mode.
+:::
+
+### Draft with Runtime Views { #draft-views }
+
+If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) entities and want to run draft specific queries on these views, set the *cds.drafts.persistence* configuration to `split` and run the queries through the [Draft Service](../fiori-drafts#draft-service) or [Application Service](../cqn-services/application-services#application-services). The [Persistence Service](../cqn-services/persistence-services) only works for non-draft specific queries.
+
+::: warning Avoid draft-enabling runtime views
+[Draft-enabling](../fiori-drafts#reading-drafts) runtime views is only supported in *cte* mode. The cds-compiler creates a corresponding draft persistence table for the view and a database schema update is required each time the runtime view is changed.
```
vmikhailenko: Maybe this whole chapter should sound like: if you draft-enable runtime views make sure that `CTE` mode is used and draft persistence is in `split` mode, you should use draft service for queries to them (which is general requirement) and your view is updatable. Like this whole chapter is a warning without the yellow box.

>cds-compiler

CDS Compiler?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+You can use path expressions navigating *to-one* compositions, as shown by `headerStatus`:
 
 ```cds
-// Supported
-entity Order as projection on bookshop.Order;
-entity Order as projection on bookshop.Order { ID, status as state };
+// CDS views supporting write operations
 entity Order as projection on bookshop.Order excluding { status };
-entity Header as projection on bookshop.OrderHeader { key ID, address.country as country };
```
vmikhailenko, line 245: This model should have reference to its base model. It is not clear that `header` is a composition.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
-- For the Persistence Service, this means the runtime _attempts_ to execute the write operation on the database view. Whether this execution is possible is [database dependent](../cqn-services/persistence-services#database-support).
-- For Application Services and Remote Services, the targeted service will reject the statement.
+- Data for elements corresponding to *expressions* and *functions* (*country*) is ignored.
+- [Deep write](./query-execution#deep-insert-upsert) via (aliased) compositions (*items*) is supported if there are corresponding compositions in the underlying entity definition. Deep write via compositions that are only defined in the view (mixins) is not supported and the data for such mixin compositions is ignored. 
+
+::: warning Path Expressions in Views
+Path expressions navigating *associations* (e.g. *header.customer.name*) are [not writable](#cascading-over-associations) by default. To avoid issues on write, annotate them with [@readonly](../../guides/providing-services#readonly).
```
vmikhailenko: I think that warning is not necessary here. We just list instructions: the one about mixins (maybe also add a reference to some definition of this concept) is as important as the paths inside associations not being writable. 

Except the keys, right?

java/working-with-cql/query-execution.md
vmikhailenko: This seems to be sufficiently explain later. Why mention it here?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+::: tip CAP Java 3.10
+Enable *cte* mode with *cds.sql.runtimeView.mode: cte*
+:::
+
+#### Resolve mode
+
+The runtime _resolves_ the view definition to the underlying persistence entities and executes the query directly against the corresponding tables.
+
+```sql
```
vmikhailenko, line 381: The more I read this (especially these references later in chapters on modifications and deletions), the more I feel that we are to vague about what this `resolve` mode actually does.

vmikhailenko: Maybe we should not reference this mode later for modifications. This is actually very hard to match the statements: 
- this mode does not support draft-enabled entities (for read, but this never clearly stated)
- the updates write through view using similar technique. 
- but drafts somehow function? What exactly will work?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,165 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+::: warning Avoid draft-enabling runtime views
+[Draft-enabling](../fiori-drafts#reading-drafts) runtime views is only supported in *cte* mode. The cds-compiler creates a corresponding draft persistence table for the view and a database schema update is required each time the runtime view is changed.
+
+[Draft activate](../fiori-drafts#editing-drafts) updates the active entity via the runtime view, therefore the view must fulfill all requirements of an [updatable view](#updatable-views).
+:::
+
+### Write through Views { #updatable-views }
+
+You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert), and [Update](./query-api#update) statements on CDS views. The CAP Java runtime attempts to resolve these to the underlying entity definitions—similar to the runtime view [resolve mode](query-execution#resolve-mode) for queries.
```
vmikhailenko: https://github.com/cap-js/docs/pull/1916/files#r2150306039

java/working-with-cql/query-execution.md
MattSchur: You're right, HANA seems to not support locks via CTEs, so only the resolve mode works. I'll add a note.

java/working-with-cql/query-execution.md
MattSchur: I think we should mention this, since `3.10` is the LTS release.

java/working-with-cql/query-execution.md
vmikhailenko: Maybe we should fix this? E.g. use different mode for queries with locks? They are usually very simple and should resolve to the baseline entity 1:1.

MattSchur: Yes, I already created a BLI to introduce query hints to control the mode and for lock we can automatically set the hint to resolve mode.

java/working-with-cql/query-execution.md
MattSchur: Restructured the section and removed the refs

java/working-with-cql/query-execution.md
MattSchur: These are independent of the type of view and should go into a general section

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,191 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+```
+- Data for elements corresponding to *expressions* and *functions* (*country*) is ignored.
+- [Deep write](./query-execution#deep-insert-upsert) via (aliased) compositions (*lineItems*) is supported if there are corresponding compositions (*items*) in the underlying entity definition. Deep write via compositions that are only defined in the view (for example via [mixins](../../cds/cql#association-definitions)) is not supported and the data is ignored.
+- [Path expressions](../../cds/cql#path-expressions) over compositions (*header.status*) are writable. For [Inserts](./query-api#insert), the view must expose all *not null* elements of the target entity and the data must include values for all of them. In the example above, the order header must have a generated key to support inserting new orders with a value for *headerStatus*.
+
+    ::: warning Path Expressions over Associations
+    Path expressions navigating *associations* (*header.customer.name*) are [not writable](#cascading-over-associations) by default. To avoid issues on write, annotate them with [@readonly](../../guides/providing-services#readonly).
```
vmikhailenko, line 286: This is explained earlier 

>For example, the following CDS model defines two resolvable views on the `Order` entity, which has a to-many *items* and a to-one *header* composition to a Header entity with a to-one customer *association*. In the projection, you can use path expressions as shown by *headerStatus* (writable) and *customerName* (read-only):

Why repeat it again?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,191 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+For [write](#updatable-views) and [delete](#delete-via-view) operations, the CAP Java runtime attempts to resolve the CDS views to their underlying persistence entities, which is only supported for simple projections.
 
-For [Insert](./query-api#insert) or [Update](./query-api#update) operations, if the projection contains functions or expressions, these values are ignored. Path expressions navigating *to-one* associations, can be used in projections as shown by the `Header` view in the following example. The `Header` view includes the element `country` from the associated entity `Address`.
+::: warning Avoid selecting to-many Associations
+Do not use [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
+:::
+::: warning Avoid Composition Definitions in Views
+Avoid [defining](../../cds/cql#association-definitions) new *compositions* in CDS views and prefer *associations* instead, as [deep write](#updatable-views) and [cascading delete](#delete-via-view) are only supported for compositions in persistence entities.
```
vmikhailenko, line 248: Isn't this true only for runtime views? I seen stakeholders who used such models and that surprisingly worked.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,191 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
 
+::: info Limitations of `resolve` mode
+Using associations that are only [defined](../../cds/cql#association-definitions) in the view, as well as complex draft queries are not supported in *resolve* mode.
+:::
+
+### Draft Queries on Views { #draft-views }
+
+When draft-enabling a CDS view, the CDS Compiler creates a corresponding draft persistence table for this view. [Draft activate](../fiori-drafts#editing-drafts) updates the active entity via the view, therefore
+
+- [Draft-enabled](../fiori-drafts#reading-drafts) CDS views must fulfill all requirements of [updatable views](#updatable-views).
```
vmikhailenko: Weird list from one item.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,191 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+Remember to run draft specific queries through the [Draft Service](../fiori-drafts#draft-service) or [Application Service](../cqn-services/application-services#application-services). The [Persistence Service](../cqn-services/persistence-services) only works for non-draft specific queries.
+
+
+::: tip Draft Queries on Runtime Views
+If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) entities and want to run draft specific queries on these views, set the *cds.drafts.persistence* configuration to `split`.
```
vmikhailenko: This should be "if you draft-enable runtime view", no? When I read this I always think about introducing projection on already draft-enabled projection.

I my projection is a runtime view and I want to draft-enable it, I have to use `split` persistence.

vmikhailenko: Maybe also cross-reference http://localhost:5173/docs/java/fiori-drafts#reading-drafts

vmikhailenko:
I like new structure a lot more! Apart from minors, I am OK with this.

java/working-with-cql/query-execution.md
MattSchur: Above it's only a side note and since `@readonly` should be added, a box makes it clearer

java/working-with-cql/query-execution.md
MattSchur: This recommendation is for all views. On write there is not much of a difference between standard views and runtime views, as we need to resolve in both cases. Therefore the "write|delete through view" sections. Deep write/delete for such compositions needs to be implemented via custom logic.

java/working-with-cql/query-execution.md
MattSchur: Yes, I want to highlight this without a box. Need to look up other formatting options.

Maybe @renejeglinsky can help with the formatting

java/working-with-cql/query-execution.md
MattSchur: This describes the case of defining a runtime view on an already draft-enabled entity.

From my understanding `split` mode is required for both cases and as of now, draft-enabling runtime views does not make much sense due to the draft table enforcing schema updates.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,191 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+To add or update CDS views without redeploying the database schema, annotate them with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This tells the CDS compiler to skip generating database views for these CDS views, and the CAP Java runtime resolves them dynamically on each request.
+
 
-If a view is too complex to be resolved by the CDS runtime, the statement remains unmodified. Views that cannot be resolved by the CDS runtime include the use of `join`, `union` and the `where` clause.
-- For the Persistence Service, this means the runtime _attempts_ to execute the write operation on the database view. Whether this execution is possible is [database dependent](../cqn-services/persistence-services#database-support).
-- For Application Services and Remote Services, the targeted service will reject the statement.
+Runtime views must be simple [projections](../../cds/cdl#as-projection-on), not using *aggregations*, *join*, *union* or *subqueries* in the *from* clause, but may have a *where* condition if they are only used to read. On write, the restrictions for [write through views](#updatable-views) apply in the same way as for standard CDS views.
```
rjayasinghe: It might be a bit obvious but you could mention that the fallback on the database view does not work as there is none existing. Also, what would happen in case one runs into those restrictions? Does the request fail?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,191 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+```
+```sql
+SELECT from BooksWithLowStock where author = 'Kafka'
+```
+
+CAP Java supports two modes for resolving runtime views on read:
+
+#### Read in `CTE` mode { #rtview-cte }
+
+This is the default mode in CAP Java `4.x`. The runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database.
```
rjayasinghe: Love the new default! 😍

java/working-with-cql/query-execution.md
MattSchur: added

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,189 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
-- For Application Services and Remote Services, the targeted service will reject the statement.
+To add or update CDS views without redeploying the database schema, annotate them with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This tells the CDS compiler to skip generating database views for these CDS views, and the CAP Java runtime resolves them dynamically on each request.
 
-Example of a view that can't be resolved:
+Runtime views must be simple [projections](../../cds/cdl#as-projection-on), not using *aggregations*, *join*, *union* or *subqueries* in the *from* clause, but may have a *where* condition if they are only used to read. On write, the restrictions for [write through views](#updatable-views) apply in the same way as for standard CDS views. However, if a runtime view cannot be resolved, a fallback to database views is not possible, and the statement fails with an error.
+
+**Read via Runtime Views** - consider the following CDS model and query:
```
vmikhailenko: Typo? Looks weird in preview.

java/working-with-cql/query-execution.md
renejeglinsky: Here's a suggestion:
image

renejeglinsky: We could also omit the "That means" or get rid of the ->

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,189 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+    ID, title, author.name as author
+} where stock < 10; // makes the view read only
+```
+```sql
+SELECT from BooksWithLowStock where author = 'Kafka'
+```
+
+CAP Java supports two modes for resolving runtime views on read:
+
+#### Read in `CTE` mode { #rtview-cte }
```
renejeglinsky: is it CTE or cte?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,189 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+::: info Limitations of `cte` mode
+[Pessimistic locking](#pessimistic-locking) queries on runtime views in *cte* mode are not supported on SAP HANA.
+:::
+
+::: tip CAP Java 3.10
+Enable *cte* mode with *cds.sql.runtimeView.mode: cte*
+:::
+
+#### Read in `Resolve` mode { #rtview-resolve }
```
renejeglinsky: Is lowercase fine?
```suggestion
#### Read in `resolve` mode { #rtview-resolve }
```

java/working-with-cql/query-execution.md
MattSchur: The formatting?

vmikhailenko: Yes.

java/working-with-cql/query-execution.md
MattSchur: I'll change this to lower case

MattSchur:
```suggestion
#### Read in `cte` mode { #rtview-cte }
```

java/working-with-cql/query-execution.md
MattSchur: @renejeglinsky Thanks! I'm fine with both options (list/arrow), whatever you think is better.

java/working-with-cql/query-execution.md
MattSchur: @renejeglinsky please feel free to change the formatting

renejeglinsky: done

MattSchur: Thanks :-)

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,196 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+::: warning Avoid selecting to-many Associations
+Do not use [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
+:::
+::: warning Avoid Composition Definitions in Views
+Avoid [defining](../../cds/cql#association-definitions) new *compositions* in CDS views and prefer *associations* instead, as [deep write](#updatable-views) and [cascading delete](#delete-via-view) are only supported for compositions in persistence entities.
+:::
+
+### Write through Views { #updatable-views }
+
+You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert), and [Update](./query-api#update) statements on CDS views that are writable on the database or can be resolved to a single persistence entity by the CAP Java runtime.
```
BraunMatthias:
```suggestion
You can run [Insert](./query-api#insert), [Upsert](./query-api#upsert) and [Update](./query-api#update) statements on CDS views that are writable on the database or can be resolved to a single persistence entity by the CAP Java runtime.
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,202 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+With CDS [views](../../cds/cdl#views-projections) you can derive new entities from existing ones, for example to rename or exclude certain elements, or to add [virtual elements](../../cds/cdl#virtual-elements-in-views) for specific use cases.
 
-Views and projections can be resolved if the following conditions are met:
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
+
+The CDS compiler generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files from your CDS model, including SQL views for the CDS views. These views are deployed to the [database](../cqn-services/persistence-services#database-support) and used by the CAP runtime to read data.
+
+For *read-only* views, you can use the full feature set of [selects](../../cds/cdl#as-select-from), including *joins*, *unions*, and *aggregations*. However, such complex views are not writable and require a schema redeployment if the view definition is changed.
```
BraunMatthias:
```suggestion
In addition, views that are used only for read access offer complex features to recombine data including *joins*, *unions*, and *aggregations*. However, be aware that such views enhanced with complex features **do not allow write queries and hence are restricted to read-only access**. Also such views always require schema redeployment if the view definition is changed. Read-only views support the full feature set of  [selects](../../cds/cdl#as-select-from) to query data. 

::: tip Prefer simple views
Prefer creating simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
:::
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,202 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
 ```
 
-### Resolvable Views and Projections { #updatable-views}
+## Views and Projections { #views }
 
-The CAP Java SDK aims to resolve statements on non-complex views and projections to their underlying entity. When delegating queries between Application Services and Remote Services, statements are resolved to the entity definitions of the targeted service. Using the Persistence Service, only modifying statements are resolved before executing database queries. This allows to execute [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) operations on database views. For [Select](./query-api#select) statements database views are always leveraged, if available.
+With CDS [views](../../cds/cdl#views-projections) you can derive new entities from existing ones, for example to rename or exclude certain elements, or to add [virtual elements](../../cds/cdl#virtual-elements-in-views) for specific use cases.
 
-Views and projections can be resolved if the following conditions are met:
+::: tip Prefer simple views
```
BraunMatthias: s.u.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
+
+The CDS compiler generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files from your CDS model, including SQL views for the CDS views. These views are deployed to the [database](../cqn-services/persistence-services#database-support) and used by the CAP runtime to read data.
```
agoerler:
```suggestion
From the CDS model the CDS compiler generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files, which include SQL views for the CDS views. These views are deployed to the [database](../cqn-services/persistence-services#database-support) and used by the CAP runtime to read data.
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+## Views and Projections { #views }
 
-The CAP Java SDK aims to resolve statements on non-complex views and projections to their underlying entity. When delegating queries between Application Services and Remote Services, statements are resolved to the entity definitions of the targeted service. Using the Persistence Service, only modifying statements are resolved before executing database queries. This allows to execute [Insert](./query-api#insert), [Upsert](./query-api#upsert), [Update](./query-api#update), and [Delete](./query-api#delete) operations on database views. For [Select](./query-api#select) statements database views are always leveraged, if available.
+With CDS [views](../../cds/cdl#views-projections) you can derive new entities from existing ones, for example to rename or exclude certain elements, or to add [virtual elements](../../cds/cdl#virtual-elements-in-views) for specific use cases.
 
-Views and projections can be resolved if the following conditions are met:
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
```
agoerler: I think this rather belongs to the best practices

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+The CDS compiler generates [DDL](../../guides/databases?impl-variant=java#generating-sql-ddl) files from your CDS model, including SQL views for the CDS views. These views are deployed to the [database](../cqn-services/persistence-services#database-support) and used by the CAP runtime to read data.
+
+For *read-only* views, you can use the full feature set of [selects](../../cds/cdl#as-select-from), including *joins*, *unions*, and *aggregations*. However, such complex views are not writable and require a schema redeployment if the view definition is changed.
+
+::: tip Indicate read-only
+Use the `@readonly` annotation to indicate that a view or a view element is not writable.
+:::
```
agoerler:
```suggestion
::: warning Indicate read-only
The CDS Compiler does not automatically detect wether a view is not writable. Instead the model should indicate using the `@readonly` annotation if a view or a view element is not writable.
:::
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+To [write data](#updatable-views) or [delete](#delete-via-view) through views, only use simple [projections](../../cds/cdl#as-projection-on). The CAP Java runtime attempts to resolve the CDS views to their underlying persistence entities, rewriting the statement and data accordingly, which is not supported for complex views.
+
+For simple projections, the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
```
agoerler:
```suggestion
The generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
```

We have different limitations for CTE-based runtime view. E.g. we don't support unions or joins.

On the other hand, I think, you should be able to used expands (nested projections) in runtime views. We have not tried this, I think.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
-entity Order as projection on bookshop.Order excluding { status };
-entity Header as projection on bookshop.OrderHeader { key ID, address.country as country };
+// CDS views supporting write operations
+entity OrderView1 as projection on Order excluding { createdBy };
+
+entity OrderView2 as projection on OrderView1 {
+  key ID,
+      header.status          as headerStatus,  // writable composition path
+      header.customer.name   as customerName @readonly,
+      items                  as lineItems,        // writable composition
```
agoerler, line 273: This contradicts [line 247](https://github.com/cap-js/docs/pull/1916/files?diff=split&w=0#diff-463b0e370d7a65bc761af955dfa1a50e4a9c50df0a2fefc2ca2770d758e095aeR247)

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+::: tip Indicate read-only
+Use the `@readonly` annotation to indicate that a view or a view element is not writable.
+:::
+
+To [write data](#updatable-views) or [delete](#delete-via-view) through views, only use simple [projections](../../cds/cdl#as-projection-on). The CAP Java runtime attempts to resolve the CDS views to their underlying persistence entities, rewriting the statement and data accordingly, which is not supported for complex views.
+
+For simple projections, the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
+
+::: warning Avoid selecting to-many Associations
+Do not use [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
```
agoerler: I think only paths across to-many associations are evil:

entity Order {
  items : association to many Items 
}

view MyOrder {
  items as myItems // OK
  items.amount // evil
}

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+      toUpper(shipToCountry) as country : String  // ignored on write
+};
+```
+```sql
+UPDATE entity OrderView2
+  { ID: 42, headerStatus: 'Processing', lineItems: [{ID: 1, book:251}] }
 ```
+- Data for elements corresponding to *expressions* and *functions* (*country*) is ignored.
+- [Deep write](./query-execution#deep-insert-upsert) via (aliased) compositions (*lineItems*) is supported if there are corresponding compositions (*items*) in the underlying entity definition. Deep write via compositions that are only defined in the view (for example via [mixins](../../cds/cql#association-definitions)) is not supported and the data is ignored.
+- [Path expressions](../../cds/cql#path-expressions) over compositions (*header.status*) are writable. For [Inserts](./query-api#insert), the view must expose all *not null* elements of the target entity and the data must include values for all of them. In the example above, the order header must have a generated key to support inserting new orders with a value for *headerStatus*.
```
agoerler:
```suggestion
- [Path expressions](../../cds/cql#path-expressions) over compositions *of one* (*header.status*) are writable. For [Inserts](./query-api#insert), the view must expose all *not null* elements of the target entity and the data must include values for all of them. In the example above, the order header must have a generated key to support inserting new orders with a value for *headerStatus*.
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
 
-If a view is too complex to be resolved by the CDS runtime, the statement remains unmodified. Views that cannot be resolved by the CDS runtime include the use of `join`, `union` and the `where` clause.
-- For the Persistence Service, this means the runtime _attempts_ to execute the write operation on the database view. Whether this execution is possible is [database dependent](../cqn-services/persistence-services#database-support).
-- For Application Services and Remote Services, the targeted service will reject the statement.
+### Delete through Views { #delete-via-view }
 
-Example of a view that can't be resolved:
+The CAP Java runtime attempts to resolve [Deletes](./query-api#delete) on CDS views to the underlying persistence entity definitions and the [write through views](#updatable-views) restrictions apply accordingly.
+
+If a view cannot be resolved, the delete operation is executed directly on the database view and the execution depends on the [database](../cqn-services/persistence-services#database-support) support.
```
agoerler, line 293: Has this ever been sucessful? Or do we just get worse error messages?

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+}
+```
+```sql
+DELETE from OrderView where ID = 42
 ```
+The delete operation is resolved to the underlying `Order` entity with ID *42* and cascades over the `header` and `items` compositions. The `delivery` composition, which is only defined in the view, is ignored and does not cascade the delete operation to `Delivery`.
+
+### Runtime Views { #runtimeviews }
+
+To add or update CDS views without redeploying the database schema, annotate them with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This tells the CDS compiler to skip generating database views for these CDS views, and the CAP Java runtime resolves them dynamically on each request.
```
agoerler:
```suggestion
To add or update CDS views without redeploying the database schema, annotate them with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This advises the CDS compiler to skip generating *persistent* database views for these CDS views. Instead, CAP Java resolves them *at runtime* on each request.
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+Runtime views must be simple [projections](../../cds/cdl#as-projection-on), not using *aggregations*, *join*, *union* or *subqueries* in the *from* clause, but may have a *where* condition if they are only used to read. On write, the restrictions for [write through views](#updatable-views) apply in the same way as for standard CDS views. However, if a runtime view cannot be resolved, a fallback to database views is not possible, and the statement fails with an error.
```
agoerler, line 322: Runtime views may use only CDS QL, which is supported by the CAP Java runtime. Joins and unions are not supported. We have more restrictions in the "resolve" mode. The "cte" mode is more powerful.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+CAP Java provides two modes for resolving runtime views during read operations: [cte](#rtview-cte) and [resolve](#rtview-resolve). 
+
+::: details Changing the runtime view mode
+To globally set the runtime view mode, use the property `cds.sql.runtimeView.mode` with value `cte` or `resolve` in the *application.yml*. To set the mode for a specific runtime view, annotate it with `@cds.java.runtimeView.mode: cte|resolve`.
```
agoerler:
```suggestion
To globally set the runtime view mode, use the property `cds.sql.runtimeView.mode` with value `cte` (the default) or `resolve` in the *application.yml*. To set the mode for a specific runtime view, annotate it with `@cds.java.runtimeView.mode: cte|resolve`.
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+To set the mode for a specific query, use a [hint](#hana-hints):
+
+```Java
+Select.from(BooksWithLowStock).hint("cds.sql.runtimeView.mode", "cte");
```
agoerler:
```suggestion
Select.from(BooksWithLowStock).hint("cds.sql.runtimeView.mode", "resolve");
```

(as `cte` is the default)

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,214 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+} where stock < 10; // makes the view read only
+```
+```sql
+SELECT from BooksWithLowStock where author = 'Kafka'
+```
+
+
+#### Read in `cte` mode { #rtview-cte }
+
+This is the default mode in CAP Java `4.x`. The runtime translates the [view definition](#runtimeviews) into a _Common Table Expression_ (CTE) and sends it with the query to the database.
```
agoerler:
```suggestion
This is the default mode since CAP Java `4.x`. The runtime translates the [view definition](#runtimeviews) into a _Common Table Expression_ (CTE) and sends it with the query to the database.
```

agoerler:
Maybe an overview table would be helpful to list which "features" can be used with

* read-only 
  * DB views
  * RT views CTE mode
  * RT views resolve mode
* read/write
  * DB views
  * RT views CTE mode
  * RT views resolve mode

java/working-with-cql/query-execution.md
MattSchur: *persistent* database views -> not sure about adding *persistent* here, it might be confused with materialized views

java/working-with-cql/query-execution.md
MattSchur: Probably rather the latter - we could skip the fallback

java/working-with-cql/query-execution.md
MattSchur: 247 rather means to-many associations in paths. I will make this clear.

java/working-with-cql/query-execution.md
MattSchur: Yes, this was meant in the context of paths (see example in 275). I will make this clear

java/working-with-cql/query-execution.md
MattSchur: "we don't support unions or joins." - yes, therefor the "For simple projections", which don't allow unions and joins.

Nested projections can be used in cte mode but only with ucsn

java/working-with-cql/query-execution.md
```diff
@@ -227,24 +227,23 @@ long deleteCount = service.run(delete).rowCount();
...
 
-For simple projections, the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
+For simple [projections](../../cds/cdl#as-projection-on), the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
 
-::: warning Avoid selecting to-many Associations
-Do not use [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
```
agoerler, line 242: -> https://en.wikipedia.org/wiki/Interface_segregation_principle

MattSchur:
> @MattSchur Are all comments addressed and can we review and merge?

Yes, please review and merge

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,213 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+For simple [projections](../../cds/cdl#as-projection-on), the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
+
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
+::: warning Avoid selecting paths over to-many Associations
+Do not use [path expressions](../../cds/cql#path-expressions-in-all-other-clauses) over [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
+:::
+::: warning Avoid Composition Definitions in Views
+Avoid [defining](../../cds/cql#association-definitions) new *compositions* in CDS views and prefer *associations* instead, as [deep write](#updatable-views) and [cascading delete](#delete-via-view) are only supported for compositions in persistence entities.
```
renejeglinsky:
```suggestion
::: warning Avoid composition definitions in views
Do not [define](../../cds/cql#association-definitions) *compositions* in CDS views and prefer *associations* instead, to avoid [cascading over associations](#cascading-over-associations).
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,213 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+For simple [projections](../../cds/cdl#as-projection-on), the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
+
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
+::: warning Avoid selecting paths over to-many Associations
+Do not use [path expressions](../../cds/cql#path-expressions-in-all-other-clauses) over [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
+:::
+::: warning Avoid Composition Definitions in Views
```
MattSchur:
```suggestion
::: Avoid composition definitions in views
```

java/working-with-cql/query-execution.md
MattSchur: It's not about avoiding cascading over associations but deep write and cascade delete are not supported by the runtime for such compositions and would need to be handled by custom code.

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,213 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
+
+For simple [projections](../../cds/cdl#as-projection-on), the generation of SQL views can be avoided by using [runtime views](#runtimeviews). This allows you to change the view definition without redeploying the database schema and is the prerequisite for lightweight extensibility via predefined extension fields.
+
+::: tip Prefer simple views
+Prefer creating multiple simple views, each tailored to a specific use case, rather than a single complex view that tries to address multiple use cases simultaneously.
+:::
+::: warning Avoid selecting paths over to-many Associations
+Do not use [path expressions](../../cds/cql#path-expressions-in-all-other-clauses) over [*to-many associations*](../../cds/cdl#to-many-associations) in the select clause of CDS views. This blocks write operations and may cause performance issues due to record duplication on read.
+:::
+::: Avoid composition definitions in views
```
MattSchur:
```suggestion
::: warning Avoid composition definitions in views
```

java/working-with-cql/query-execution.md
```diff
@@ -223,45 +223,213 @@ CqnDelete delete = Delete.from("bookshop.Orders").matching(singletonMap("OrderNo
...
 
-Example of a view that can't be resolved:
+To set the mode for a specific query, use a [hint](#hana-hints):
+
+```Java
+Select.from(BooksWithLowStock).hint("cds.sql.runtimeView.mode", "resolve");
+```
+:::
+
+The following section introduces both modes using the CDS model and query below:
```
renejeglinsky: "modes" is not about cte or resolve here, right? It's about the characteristics of a runtime view, or?

MattSchur: It is about cte and resolve, the modes mentioned in line 324.
```suggestion
The following two sections introduce both modes using the CDS model and query below:
```

renejeglinsky:
```suggestion
The next two sections introduce both modes using the following CDS model and query:
```

java/working-with-cql/query-execution.md
MattSchur: -> https://github.com/cap-js/docs/pull/1957

PR #1925, opened 2025-06-18
Microservices Guide: use cds up in deploy section

renejeglinsky:


renejeglinsky:
@Akatuoro Can you please review? Thanks :)

Akatuoro:
> How about adding this script then to our samples? I never heard a voice complaining about missing undeploy docs, but I do see the need to undeploy things when you're playing around with samples. So why not put it into the samples? :)

👍

It's already in the capire/samples [package.json](https://github.com/capire/samples/blob/377573c8bb86da48d03f230279bb9dd115856f7f/package.json#L28) if that's what you mean.

The question may come up more once people are calling `cds up` instead of `cf deploy` ;) but we can also keep it out of the docs for now.

PR #1927, opened 2025-06-20
install cds-dk instead of add

AdrianRygiol:
While setting up my new notebook, I've noticed that we shall `npm install` the cds-dk (not `npm add`) in getting started.


AdrianRygiol:
@chgeo: Hi Christian, alright then I close this PR, no change needed.

PR #1929, opened 2025-06-23
Fix very minor issue in authorization.md

qmacro:
Singular 'entity' vs plural 'entities'; it seems that the plural is used predominantly here, so fixed in a way that leans in that direction.


renejeglinsky (approved):
Thanks a lot!

PR #1930, opened 2025-06-23
Improve description of srv.send

vkozyura:


guides/providing-services.md
```diff
@@ -1123,18 +1123,45 @@ POST .../sue/Foo(2)/Sue.order {"x":1} // bound action
...
 
 
-**Programmatic** usage via **generic APIs** would look like this for Node.js:
+**Programmatic** usage via **generic APIs** for Node.js:
+
+For unbound actions and functions:
+
+```ts
+async function srv.send (
+  event   : string | { event, data?, headers? },
```
etimr: please add
`headers?: object`

guides/providing-services.md
```diff
@@ -1123,18 +1123,45 @@ POST .../sue/Foo(2)/Sue.order {"x":1} // bound action
...
 
 ```js
   const srv = await cds.connect.to('Sue')
   // unbound actions/functions
   await srv.send('sum',{x:1,y:2})
   await srv.send('add',{x:11,to:2})
   await srv.send('stock',{id:2})
-  // bound actions/functions
+  // actions/functions bound to collection
   await srv.send('getStock','Foo',{id:2})
```
etimr, line 1162: did you check that it works the same way for functions? Usually for actions we are sending parameters as payload, for functions they are sent in the URL, for example: `/my/func(name='Hubert')` or `/my/func?name='Peter'`, will in this case `name` be sent programmatically as a `data` as well?

vkozyura: @etimr: just double-checked - yes, it works also for functions

guides/providing-services.md
vkozyura: done

node.js/core-services.md
```diff
@@ -796,9 +796,11 @@ return : result of this.dispatch(req)
 
 Use this method to send synchronous requests to a service for execution.
 
--  `method` can be an HTTP method, or a name of a custom action or function
+-  `method` is an HTTP method
 -  `path` can be an arbitrary URL, starting with a leading `'/'`, it is passed to a service without any modification as a string
 
+To call bound / unbound actions and functions from the service further variants of `srv.send` are additionally supported, as described in the Section [Calling Actions / Functions](../guides/providing-services#calling-actions-functions). Basically, use the action or function name instead of the HTTP method.
```
renejeglinsky:
```suggestion
To call bound / unbound actions and functions from the service, further variants of `srv.send` are additionally supported, as described in the section [Calling Actions / Functions](../guides/providing-services#calling-actions-functions). Basically, use the action or function name instead of the HTTP method.
```

renejeglinsky (approved):

PR #1934, opened 2025-06-25
fix (?) the after phase handler sample in messaging topic

qmacro:
I may have misunderstood something, but ...

The `req` parameter is second, not first, in `after` phase handlers (see https://cap.cloud.sap/docs/node.js/core-services#srv-after-request). Different values, depending on the actual event (create, update, delete) will be provided for the first argument, none of which is the request.


sjvans:
good catch!

renejeglinsky (approved):

PR #1937, opened 2025-06-26
Fix (?) to shared-db testing in microservices.md

qmacro:
The `cds db -2 sql` and `cds db -2 hana` commands are a bit suspect, as the `db` command doesn't exist. From the paragraph just before these commands, I suspect this should be `compile`, but even then, we need also to specify the `--all` option, otherwise we get the message "You must specify a model to compile.".

### DIFF
--- a/guides/deployment/microservices.md
+++ b/guides/deployment/microservices.md
@@ -162,7 +162,7 @@ cd shared-db
 cds compile db -2 sql
 ```
 ```sh
-cds compile -2 hana --all
+cds compile db -2 hana
 ```
 
 ```sh
--- a/guides/deployment/microservices.md
+++ b/guides/deployment/microservices.md
@@ -159,7 +159,7 @@ cd shared-db
 ```
 
 ```sh
-cds compile -2 sql --all
+cds compile db -2 sql
 ```
 ```sh
 cds compile -2 hana --all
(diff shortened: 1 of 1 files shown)


vl-leon:
I do not see the **--all** option documented in cds 9, see `cds help compile`
Documented are **-s all** and **-l all**

renejeglinsky:
Why not use cds compile, so the full command? Is there a need to use the shortcut/default? If not, we should not make our users guess/research what we mean.

guides/deployment/microservices.md
```diff
@@ -159,10 +159,10 @@ cd shared-db
 ```
 
 ```sh
-cds db -2 sql
+cds compile -2 sql --all
```
renejeglinsky:
```suggestion
cds compile db -2 sql
```

guides/deployment/microservices.md
```diff
@@ -159,10 +159,10 @@ cd shared-db
...
 ```
 ```sh
-cds db -2 hana
+cds compile -2 hana --all
```
renejeglinsky:
```suggestion
cds compile db -2 hana
```

vl-leon (approved):

renejeglinsky (approved):

PR #1939, opened 2025-06-27
Returning media data in actions and functions

davidhunglam:
Added subsection about returning media data streams in custom actions and functions.


stewsk:
Hi @davidhunglam ,
is this relevant for all kinds of services or is it specific for OData?
If OData specific, this section should better be placed in https://pages.github.tools.sap/cap/docs/advanced/odata.

davidhunglam:
@stewsk on second thought, in my opinion, annotating operation returns with `@Core.MediaType` is not really adapter specific. It's just that there's currently an implementation only for OData services, but in theory this could be supported also for HCQL, etc.

renejeglinsky:
Is this shown in a sample?

davidhunglam:
> Is this shown in a sample?

No, not yet.

cds/cdl.md
```diff
@@ -2101,7 +2101,21 @@ service CatalogService {
 
 Explicitly modelled binding parameters are ignored for OData V2.
 
+#### Returning Media Data Streams { #actions-returning-media}
 
+Actions and functions can also be modelled to return streamed media data like images, CSVs, etc. In order to do so, the action or function's return type has to refer to a [predefined type](#types) annotated with [media data annotations](/guides/providing-services#annotating-media-elements) that is defined in the same service. The minimum set of annotations required is `@Core.MediaType`.
```
renejeglinsky:
```suggestion
Actions and functions can also be modeled to return streamed media data such as images and CSV files. To achieve this, the return type of the actions or functions must refer to a [predefined type](#types), annotated with [media data annotations](/guides/providing-services#annotating-media-elements), that is defined in the same service. The minimum set of annotations required is `@Core.MediaType`.
```

PR #1940, opened 2025-06-27
Provided a better solution for reading dead outbox entries

t-bonk:


java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
 
 :::
 
-### Filter for Dead Entries
+### Reading Dead Entries
 
-This filtering can't be done on the database since the maximum number of attempts is only available from the CDS properties.
-
-To ensure that only dead outbox entries are returned when reading `DeadOutboxMessages`, the following code provides the handler for the `DeadLetterQueueService` and the `@After-READ` handler that filters for the dead outbox entries:
+This filtering of dead entries is done on the database by adding `where` clauses for each outbox and their maximum number of retries. The following code provides the handler for the `DeadLetterQueueService` that modifies the where clause by adding
```
BraunMatthias:
```suggestion
Filtering the dead entries is done by adding by adding an appropriate `where`-clause to `READ`-queries on the outbox message entries which reached maximum number of retries. The following code provides an example handler implementation defining this behaviour for the `DeadLetterQueueService`:
```

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
+additional conditions for filtering the outbox entries:
```
BraunMatthias:
```suggestion
```

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
-		List<DeadOutboxMessages> deadEntries = context
-				.getResult()
-				.listOf(DeadOutboxMessages.class)
-				.stream()
-				.filter(entry -> entry.getAttempts() >= outboxConfigs.getService(entry.getTarget()).getMaxAttempts())
-				.toList();
-
-		context.setResult(deadEntries);
-	}
+	private final PersistenceService db;
```
BraunMatthias: identation

java/outbox.md
```diff
@@ -419,8 +455,7 @@ The injected `PersistenceService` instance is used to perform the operations on
 [Learn more about CQL statement inspection.](./working-with-cql/query-introspection#cqnanalyzer){.learn-more}
 
 ::: tip Use paging logic
-Avoid to read all entries of the `cds.outbox.Messages` or `OutboxDeadLetterQueueService.DeadOutboxMessages` table at once, as the size of an entry is unpredictable
-and depends on the size of the payload. Prefer paging logic instead.
+Avoid to read all entries of the `cds.outbox.Messages` or `OutboxDeadLetterQueueService.DeadOutboxMessages` table at once, as the size of an entry is unpredictable and depends on the size of the payload. Prefer paging logic instead.
```
BraunMatthias:
```suggestion
Avoid to read all outbox entries at once as a single entry can have significant size reflecting the request's payload. Prefer `READ`-queries with paging instead.
```

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
-		context.setResult(deadEntries);
-	}
+	private final PersistenceService db;
+
+    public DeadOutboxMessagesHandler(@Qualifier(PersistenceService.DEFAULT_NAME) PersistenceService db) {
+        this.db = db;
+    }
+
+    @Before(entity = DeadOutboxMessages_.CDS_NAME)
+    public void modifyWhereClause(CdsReadEventContext context) {
```
BraunMatthias:
```suggestion
    public void addDeadEntryFilter(CdsReadEventContext context) {
```

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
+        CqnSelect cqn = context.getCqn();
+        Optional<Predicate> outboxFilters = this.createOutboxFilters(context.getCdsRuntime());
+        CqnSelect modifiedCqn = copy(
+          cqn,
```
BraunMatthias:
```suggestion
          context.getCqn(),
```

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
-	}
+	private final PersistenceService db;
+
+    public DeadOutboxMessagesHandler(@Qualifier(PersistenceService.DEFAULT_NAME) PersistenceService db) {
+        this.db = db;
+    }
+
+    @Before(entity = DeadOutboxMessages_.CDS_NAME)
+    public void modifyWhereClause(CdsReadEventContext context) {
+        CqnSelect cqn = context.getCqn();
```
BraunMatthias:
```suggestion
```

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
+        Optional<Predicate> outboxFilters = this.createOutboxFilters(context.getCdsRuntime());
+        CqnSelect modifiedCqn = copy(
+          cqn,
+          new Modifier() {
+              @Override
+              public CqnPredicate where(Predicate where) {
```
BraunMatthias: much simpler logic:
- do the copy only in case of outboxFilters.isPresent()
- just return outboxFilters.get().and(where) in the modifier. Anding null should be fine.

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
+                      return null;
+                  }
+              }
+          });
+
+        context.setCqn(modifiedCqn);
+    }
+
+    private Optional<Predicate> createOutboxFilters(CdsRuntime runtime) {
+        List<OutboxService> outboxServices = runtime.getServiceCatalog().getServices(OutboxService.class)
```
BraunMatthias: what about iterating the persistent outboxes from application configuration (outboxConfigs)+ the known defaults? INMEMORY_NAME would not be required.

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
+                .filter(s -> !s.getName().equals(OutboxService.INMEMORY_NAME)).toList();
+        CdsProperties.Outbox outboxConfigs = runtime.getEnvironment().getCdsProperties().getOutbox();
+
+        Predicate where = null;
+        for(OutboxService service : outboxServices) {
+            OutboxServiceConfig config = outboxConfigs.getService(service.getName());
```
BraunMatthias: why is config required at all?

java/outbox.md
```diff
@@ -355,29 +355,65 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
...
+            Predicate targetPredicate = CQL.get(Messages.TARGET).eq(service.getName()).and(CQL.get(Messages.ATTEMPTS).ge(config.getMaxAttempts()));
```
BraunMatthias: why do we need to filter by individual outbox and then OR the condition meeting all outboxes?

BraunMatthias (changes requested):

java/outbox.md
t-bonk: @BraunMatthias Every outbox can have its own configuration for `maxAttempts`. To read only the dead entries per outbox, this filter is required.

java/outbox.md
t-bonk: @BraunMatthias If an outbox has the default configuration, iterating over the configurations doesn't return the specific config for an outbox. I observed this while testing the code; in this case no outbox can be found and hence no dead entries are read at all.

java/outbox.md
t-bonk: @BraunMatthias To get the configuration for the `maxAttempts`.

java/outbox.md
BraunMatthias: same reason - I just overlooked this aspect

java/outbox.md
```diff
@@ -355,29 +355,58 @@ It is crucial to make the service `OutboxDeadLetterQueueService` accessible for
 
 :::
 
-### Filter for Dead Entries
+### Reading Dead Entries
 
-This filtering can't be done on the database since the maximum number of attempts is only available from the CDS properties.
-
-To ensure that only dead outbox entries are returned when reading `DeadOutboxMessages`, the following code provides the handler for the `DeadLetterQueueService` and the `@After-READ` handler that filters for the dead outbox entries:
+Filtering the dead entries is done by adding by adding an appropriate `where`-clause to `READ`-queries on the outbox message entries which reached maximum number of retries. The following code provides an example handler implementation defining this behaviour for the `DeadLetterQueueService`:
```
BraunMatthias:
```suggestion
Filtering the dead entries is done by adding an appropriate `where`-clause to all `READ`-queries which covers all outbox message entries with maximum number of retries. The following code provides an example handler implementation defining this behaviour for the `DeadLetterQueueService`:
```

java/outbox.md
BraunMatthias: This might be only true for the two default outboxes which needed to be explicitly named here. But what happens in 398 for those?

PR #1944, opened 2025-06-30
Java: Docs for improved event handler signatures

beckermarc:
- Arbitrary return types
- Service arguments
- Entity reference arguments


java/event-handlers/index.md
```diff
@@ -416,12 +466,27 @@ public List<Books> readBooks(CdsReadEventContext context) {
...
 ::: tip
 To learn how to build your own Result objects, have a look at the [Result Builder API](../cqn-services/application-services#result-builder)
 :::
 
+For custom actions or functions you can directly return the return value of the corresponding action or function in your event handler:
+
+Given the following CDS model:
+```cds
+service World {
+  function hello() returns String;
```
vmikhailenko, line 478: I wonder if we can also inject arguments for operations. 

We have the generator feature for template event handlers. I wonder if it now should use this feature as well?

java/event-handlers/index.md
```diff
@@ -391,6 +389,58 @@ During the `Before` and `On` phase changes affect the data carried by the CQN st
...
+:::
+
+Entity data arguments work on all events that operate with a `CqnStatement`. This is the case for all CRUD events and custom bound actions or functions.
+
+You can directly use these reference to build further queries in your event handlers:
+
+```java
+@After(event = CqnService.EVENT_UPDATE)
+public void changedBook(Books_ ref) {
+  CqnSelect select = Select.from(ref).columns(b -> b.title());
```
BraunMatthias:
```suggestion
  var select = Select.from(ref).columns(b -> b.title());
```
I'd better avoid `CqnSelect` in favour of upcoming improvements.

java/event-handlers/index.md
```diff
@@ -391,6 +389,58 @@ During the `Before` and `On` phase changes affect the data carried by the CQN st
...
+You can get a entity reference reflecting the reference of the currently processed CQN statement, by declaring a corresponding argument in your method signature.
+
+```java
+@After(event = CqnService.EVENT_UPDATE, entity = Books_.CDS_NAME)
+public void changedBook(CqnStructuredTypeRef ref) { }
+```
+
+The CAP Java SDK Maven Plugin can generate query builder interfaces for entities defined in the CDS model. These interfaces allow for a [type-safe query building](../working-with-cql/query-api#concepts) and can be used in arguments as well:
+
+```java
```
BraunMatthias, line 403: I'd omit this option as the entity argument is redundant

vmikhailenko:
Maybe we should also add some kind of "when-to-use-what" tips or guidelines: e.g. prefer tight and typed signatures for the handlers first, switch to the more generic versions if you really want to write a generic handler. 

E.g. `Books_ ref, CatalogService service` is better because [reason], the `Books book, Books_ ref, CatalogService service` also possible if I want to have payload and the ref together.

java/event-handlers/index.md
beckermarc: > I wonder if we can also inject arguments for operations.

We don't support that at the moment. It is more complex as arguments would need to be matched by some qualifier as they can't solely be matched by type (an action might have multiple String input parameters). Also I discussed with @BraunMatthias already that I prefer using the event contexts to obtain the arguments. This ensures that you don't need to repeat the action/function signature completely in your event handler classes and thus makes it easier to extend an action/function with additional parameters

> We have the generator feature for template event handlers. I wonder if it now should use this feature as well?

Yes we can look into that. Good point.

java/event-handlers/index.md
beckermarc: This matches the documentation structure that we also when describing the same concept for events based on contexts and entities based on data accessor interfaces. I'd keep it symmetrical for now.

beckermarc:
> Whole statements are left for generic handlers?

There is actually not really a great use case for these so far, so I'd keep that a hidden feature for now. The reference has proven much more useful.

beckermarc:
> Maybe we should also add some kind of "when-to-use-what" tips or guidelines: e.g. prefer tight and typed signatures for the handlers first, switch to the more generic versions if you really want to write a generic handler.

That's actually not so easy and highly depends on the logic that is within the event handler method I think. I guess we still need to learn from some good examples for these new possibilities to be able to document something here.

java/event-handlers/index.md
vmikhailenko: >Also I discussed with @BraunMatthias already that I prefer using the event contexts to obtain the arguments

Looks kinda off when I need `Book_ ref, BookCreateBookContext context` to enjoy the convenience.

vmikhailenko:
Maybe also mention that event handler returning a collection of entities will lead to broken result in case of queries with `$count`. I remember more than one issue about this. This is very hard to figure that out without documentation. Or we should fix it for good.

java/event-handlers/index.md
beckermarc: You mean from naming perspective?
1. A bound action on books should not be named `createBook()`, but simply `create()` I think => `Book_ ref, BookCreateContext context`. That already looks better.
2. I stumbled already multiple times from naming perspective over the prefixing with the entity name that we did to avoid conflicts. We didn't find a better solution at that point in time, so I guess we have to live with it now.

beckermarc:
> Maybe also mention that event handler returning a collection of entities will lead to broken result in case of queries with `$count`. I remember more than one issue about this. This is very hard to figure that out without documentation. Or we should fix it for good.

That's something that is mentioned here: https://pages.github.tools.sap/cap/docs/java/cqn-services/application-services#read-result. Which is also somehow linked from this page. I will improve the link however to link to the upper section instead.

java/event-handlers/index.md
vmikhailenko: No, I mean that we use different "styles" together.

- "inject only what you need". Take payload or a ref as an argument, when you only need that. Take the service or autowire different service if you need that as well. 
- "inject event-specific generic context". E.g. one for action or `CdsWhateverEventContext`. It has all you will ever need while being type-safe enough.
- "generic handler". EventContext + appropriate branching/casting etc.

This is kinda in between. 

Maybe we should include couple of examples for actions so that handler generator uses these to suggest same? E.g. `Result create(Book_ ref, BookCreateContext context)`?

beckermarc: I am not sure what exactly you would want to improve about the docs now? 
I think like this is rather into the area of best practice, while this PR just wants to describe the technical possibilities first.
Can we maybe follow up on these topics separately to unblock merging this PR for the release?

vmikhailenko: Go ahead. I just think that this makes event handlers even more confusing unless we provide advice.

vmikhailenko:
> > Maybe also mention that event handler returning a collection of entities will lead to broken result in case of queries with `$count`. I remember more than one issue about this. This is very hard to figure that out without documentation. Or we should fix it for good.
> 
> That's something that is mentioned here: https://pages.github.tools.sap/cap/docs/java/cqn-services/application-services#read-result. Which is also somehow linked from this page. I will improve the link however to link to the upper section instead.

Maybe a warning then? It is _very_ unexpected behaviour, especially in `@After` handler. 
Non-blocker, again, just an ask.

beckermarc:
> Maybe a warning then? It is very unexpected behaviour, especially in @After handler.
Non-blocker, again, just an ask.

We can make the text in the application service section more explicit and add a warning to the count docs there I think.

renejeglinsky (approved):

PR #1946, opened 2025-07-02
Add description for js rules

daogrady:
We introduced a new set of lint rules targeting JS implementation files, instead of just .cds files.
This PR introduces automatic documentation extraction for these new rules, to be found in list form under `/tools/cds-lint/rules/`

Screenshot 2025-07-02 at 12 39 12

We do not yet introduce extended documentation per rule for these new rules, as I am not clear yet whether they are automatically generated or maintained manually.


chgeo (commented):
LGTM.

The detail pages are written manually, with teh good/bad examples taken from the rules' unit test.  We can add them later, so that we can publish this with the June release today.

chgeo:
@renejeglinsky this should be merged for the June release.

PR #1958, opened 2025-07-07
Java: Document `@Capabilities.ExpandRestrictions.*`

vmikhailenko:


guides/security/aspects.md
```diff
@@ -528,17 +528,27 @@ The total number of request of OData batches can be limited by application confi
...
-Settings <Config java>cds.odataV4.batch.maxRequests</Config> resp. <Config java>cds.odataV2.batch.maxRequests</Config> specify the corresponding limits.
+Use settings <Config java>cds.odataV4.batch.maxRequests</Config> resp. <Config java>cds.odataV2.batch.maxRequests</Config> to limit the amount of queries per OData `$batch`.
+
+To prevent clients from requesting too much data, you can define restrictions on `$expands` for your entities: 
+
+- Use `@Capabilities.ExpandRestrictions.Expandable: false` to prevent any expands from the entity. 
+- Use `@Capabilities.ExpandRestrictions.NonExpandableProperties: [...]` to restrict expands for certain properties.
+- Use `@Capabilities.ExpandRestrictions.MaxLevels: ...` to set maximum allowed depth of an `$expand` from this entity. You can set an application-wide limit with <Config>cds.query.restrictions.expand.maxLevels = \<max depth\></Config> that applies to all entities. Value `-1` indicates absence of limit.
+
+Good candidates for expand restrictions are the associations to the same type (for example, when your entity represents tree or a hierarchy) or backlink associations of the compositions or many-to-many associations.
```
agoerler:
```suggestion
Good candidates for expand restrictions are associations to the same type (for example, when your entity represents tree or a hierarchy<sup>1></sup>), backlink associations of compositions, or many-to-many associations.

<sup>1></sup>Hierarchical requests from the UI5 tree table do not use expand and are not affected by expand restriction.
```

guides/security/aspects.md
```diff
@@ -528,17 +528,29 @@ The total number of request of OData batches can be limited by application confi
 
 <div class="impl java">
 
-Settings <Config java>cds.odataV4.batch.maxRequests</Config> resp. <Config java>cds.odataV2.batch.maxRequests</Config> specify the corresponding limits.
+Use settings <Config java>cds.odataV4.batch.maxRequests</Config> resp. <Config java>cds.odataV2.batch.maxRequests</Config> to limit the amount of queries per OData `$batch`.
```
beckermarc: The way we have implemented this, it is not really a HTTP Server / Protocol Adapter specific thing, as it also applies on `ApplicationService`-level. Should we somehow make that clear?

beckermarc (commented):
Looks good!

guides/security/aspects.md
vmikhailenko: >I wonder if that is more about the runtime than about protocol adapter.

Next chapter?

beckermarc: The annotations are from OData, so Protocol Adapter section from that perspective makes sense. Maybe we can just mention it in the text?

guides/security/aspects.md
```diff
@@ -536,6 +536,8 @@ To prevent clients from requesting too much data, you can define restrictions on
 - Use `@Capabilities.ExpandRestrictions.NonExpandableProperties: [...]` to restrict expands for certain properties.
 - Use `@Capabilities.ExpandRestrictions.MaxLevels: ...` to set maximum allowed depth of an `$expand` from this entity. You can set an application-wide limit with <Config java>cds.query.restrictions.expand.maxLevels = \<max depth\></Config> that applies to all entities. Value `-1` indicates absence of limit.
 
+These restrictions also apply on CQN queries sent to the [application services](/java/cqn-services/#application-services).
```
beckermarc: Maybe put this into a tip. And I wouldn't say also, as they just apply there.

vmikhailenko: Not an advice, rather just a fact.

guides/security/aspects.md
```diff
@@ -536,7 +536,7 @@ To prevent clients from requesting too much data, you can define restrictions on
 - Use `@Capabilities.ExpandRestrictions.NonExpandableProperties: [...]` to restrict expands for certain properties.
 - Use `@Capabilities.ExpandRestrictions.MaxLevels: ...` to set maximum allowed depth of an `$expand` from this entity. You can set an application-wide limit with <Config java>cds.query.restrictions.expand.maxLevels = \<max depth\></Config> that applies to all entities. Value `-1` indicates absence of limit.
 
-These restrictions also apply on CQN queries sent to the [application services](/java/cqn-services/#application-services).  
+These restrictions apply on CQN queries sent to the [application services](/java/cqn-services/#application-services).
```
beckermarc:
```suggestion
::: tip
These restrictions are enforced for `READ` events on [Application Services](/java/cqn-services/#application-services).  
:::
```

guides/security/aspects.md
beckermarc: I think a tip also has the effect that this sentence is not so easily overlooked.

vmikhailenko: Warning then? We warn people that their queries from custom code also subjects to these restrictions.

beckermarc (approved):

agoerler (approved):

guides/security/aspects.md
```diff
@@ -528,17 +528,40 @@ The total number of request of OData batches can be limited by application confi
...
+Use settings <Config java>cds.odataV4.batch.maxRequests</Config> resp. <Config java>cds.odataV2.batch.maxRequests</Config> to limit the amount of queries per OData `$batch`.
+
+To prevent clients from requesting too much data, you can define restrictions on `$expands` for your entities: 
+
+- Use `@Capabilities.ExpandRestrictions.Expandable: false` to prevent any expands from the entity. 
+- Use `@Capabilities.ExpandRestrictions.NonExpandableProperties: [...]` to restrict expands for certain properties.
+- Use `@Capabilities.ExpandRestrictions.MaxLevels: ...` to set maximum allowed depth of an `$expand` from this entity. You can set an application-wide limit with <Config java>cds.query.restrictions.expand.maxLevels = \<max depth\></Config> that applies to all entities. Value `-1` indicates absence of limit.
+
+:::warning
+These restrictions are enforced on 'READ' events on [Application services](/java/cqn-services/#application-services).
```
renejeglinsky, line 544: What would I need to do if I skip the application service in my application? It is possible to go directly to db services, right?

vmikhailenko: Yes, but we do not recommend that. You can always introduce special projection or re-route the request to the different service.

renejeglinsky: So we're just stating this applies to READ on application services, which, by the way, is the way you should do it. Right?
Isn't it then more like a TIP than a warning? Or let's say, shouldn't we phrase it as a TIP rather than a warning?

vmikhailenko: >So we're just stating this applies to READ on application services, which, by the way, is the way you should do it. Right?

Yes. If expand or filter are forbidden, there is the reason for this.

We warn people that this also applies to accesses via APIs, because the annotation is OData-specific, but also applies to the API access. Kinda like "be aware of this when you later use the service e.g. in an action or from another service".

renejeglinsky: Understood, thanks!
Thinking out loud: If I, as a developer, am not aware of that when using the service, for example, in an action, I would get an error message, right? How about adding a troubleshooting for that error so that developers facing this error get some help?

vmikhailenko: I hope we will provide the clear way forward when domain services will be done. 

These restrictions are only one case of more generic problem: what to do when service itself or other services want "premium" access and the restrictions should be skipped. And which ones.

renejeglinsky: Alright, then I'd say we're done for now. Thanks a lot! :)

PR #1963, opened 2025-07-09
Java: Explain deep updates supported by change tracking with examples

vmikhailenko:
Attempts to illustrate supported/non-supported statements in change tracking.

### DIFF
--- a/java/change-tracking.md
+++ b/java/change-tracking.md
@@ -283,14 +283,65 @@ values of the entity: data that weren't present in the old values are considered
 the new values are considered as deleted. Elements that are present in both old and new values but have different values
 are considered as modified. Each change detected by the change tracking feature is stored in the change log as a separate entry.
 
+As a general rule, modify change tracked entities only with statements that specify primary keys and avoid [searched updates](/java/working-with-cql/query-api#searched-update).
+
+### Changes in Deeply Structured Documents
+
 In the case of the deeply structured documents, for example, entities with the compositions, the change tracking feature detects
 the changes across the complete document and stores them in the change log with the metadata reflecting the structure of the change.
 
-Take the order and item model used previously in this guide as an example. If you change values for the tracked elements with the deep update, for example, the customer name in the order and the quantity of the item, the change log contains two entries: one for the order and one for the item. The change log entry for the item will also reflect that the root of the change is an order. Both changes will be reachable through the association `changes` of the order entity.
+Take the order and item model used previously in this guide as an example.
 
-:::warning Prefer deep updates for change-tracked entities
-If you change the values of the `OrderItems` entity directly via an OData request or a CQL statement, the change log contains only one entry for the item. The change won't be associated with an order and will not be reachable through the `changes` association. While the updates of composition targets directly are possible, the change tracking feature does not attempt to resolve the parent entity by itself, it requires that either OData request or a CQL statement provide the reference to the parent, for example, `Orders`.   
+For deep updates, use `Update` statement with [delta representation](/java/working-with-cql/query-api#deep-update-delta) for its items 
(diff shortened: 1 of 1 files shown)


java/change-tracking.md
```diff
@@ -283,15 +283,65 @@ values of the entity: data that weren't present in the old values are considered
...
+### Changes in Deeply Structured Documents
+
 In the case of the deeply structured documents, for example, entities with the compositions, the change tracking feature detects
 the changes across the complete document and stores them in the change log with the metadata reflecting the structure of the change.
 
-Take the order and item model used previously in this guide as an example. If you change values for the tracked elements with the deep update, for example, the customer name in the order and the quantity of the item, the change log contains two entries: one for the order and one for the item. The change log entry for the item will also reflect that the root of the change is an order. Both changes will be reachable through the association `changes` of the order entity.
+Take the order and item model used previously in this guide as an example.
+
+For deep updates, use `Update` statement with [delta representation](/java/working-with-cql/query-api#deep-update-delta) for its items 
+or [full set representation](/java/working-with-cql/query-api#deep-update-full-set) if you overwrite complete document. Rest of
```
beckermarc: Rest of ... ?

java/change-tracking.md
```diff
@@ -283,15 +283,65 @@ values of the entity: data that weren't present in the old values are considered
...
+Similar `Delete` statement can be used to remove an item from an order:
+
+```java
+Delete.from(Orders_.class, o -> o.filter(f -> f.ID().eq("...")).items().filter(i -> i.ID().eq("...")));
+```
+
+Such statements supported only if: 
+- path expression starts in the root of the document (order, in this case)
+- path navigates only through the compositions 
+- all its segments specify keys so that the complete path targets single target instance (item, in this case)
```
beckermarc: single instance is not really required, right? A bulk update to the items() collection or a bulk delete of the items() collection would also work, right?

java/change-tracking.md
vmikhailenko: Mistake.

java/change-tracking.md
vmikhailenko: Yes.

beckermarc: Should we correct the docs then, as they seem to state that single instance is required?

vmikhailenko: This is exactly what I wanted to demonstrate. 

I can add an example for bulk update separately.

java/change-tracking.md
```diff
@@ -283,15 +283,73 @@ values of the entity: data that weren't present in the old values are considered
...
+```java
+Delete.from(Orders_.class, o -> o.filter(f -> f.ID().eq("...")).items());
+```
+
+Same path expression can be used in the `Insert` statement to directly add an item to a composition.
+
+Changes are correctly referenced to the root if the following conditions are true:
+- path expression starts at the root of the document (order, in this case)
+- path navigates only through the compositions within the same document tree
+- segments of the path, except the last one, include keys
```
beckermarc:
```suggestion
- All segments of the path, except the last one, must include keys
```

java/change-tracking.md
```diff
@@ -283,15 +283,73 @@ values of the entity: data that weren't present in the old values are considered
...
+Last segment can omit keys to indicate bulk deletion:
+
+```java
+Delete.from(Orders_.class, o -> o.filter(f -> f.ID().eq("...")).items());
+```
+
+Same path expression can be used in the `Insert` statement to directly add an item to a composition.
+
+Changes are correctly referenced to the root if the following conditions are true:
+- path expression starts at the root of the document (order, in this case)
```
beckermarc:
```suggestion
- path expression starts at the root of the document (`Orders`, in this case)
```

beckermarc (approved):

smahati (approved):

PR #1982, opened 2025-07-22
Replace 'entity' identifier example from cdl.md

qmacro:
In the "Keywords & Identifiers" section there's an example of the keyword `entity` being used as an identifier, which seems confusing and "asking for trouble" :-)

Moreover, a little further on, in a "Warning" section, it states "Delimited identifiers in general, but in particular non-ASCII characters, _or keywords as identifiers_ should be avoided as much as possible, for reasons of interoperability." (emphasis mine).

In addition, the definition of the `entity` aspect also had an `Integer` key called `ID`, which is perhaps not to be universally encouraged (better to use `cuid` with its `UUID` type).

So I reworked this section.


danjoa:
Oha ... well, I'm afraid to agree with all you wrote and fixed 😬
mea culpa :)

danjoa (approved):

renejeglinsky (approved):

PR #1988, opened 2025-07-25
Document `@Capabilities.FilterRestrictions`

eugene-andreev:


vmikhailenko:
LGTM.

If we will implement more of these, we need a chapter for these.

PR #1996, opened 2025-07-30
Minor fix in common.md

qmacro:
There were likely only three `CodeList` based entities but there are now four. So I've removed the (now inaccurate) number leaving it flexible and (now) correct :-)


smahati (approved):
Thank you :)

PR #2002, opened 2025-08-01
assert.constraint remove duplicate section

DanSchlachter:
`<div id="assertconstraints" />` can be found in line 924 and 931, so the fragment is visible twice in internal capire page


renejeglinsky:
Duplicate of https://github.com/cap-js/docs/pull/1972

PR #2010, opened 2025-08-06
add SELECT.forUpdate(ignoreLocked)

kulliax:
add the forUpdate(ignoreLocked) property on CQL Select, see https://github.com/cap-js/cds-dbs/issues/917

official sql docs: 
https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20fcf24075191014a89e9dc7b8408b26.html?locale=en-US#loio20fcf24075191014a89e9dc7b8408b26__for_update_clause
https://www.postgresql.org/docs/17/sql-select.html


johannes-vogel:
Hi @kulliax,

thanks for opening the PR.
Before merging, we first have to discuss whether we want to make that an official API. So far, the tendency was not make it a public API.

renejeglinsky:
Hi @johannes-vogel , would be nice if you gave an update here. Thanks.

PR #2011, opened 2025-08-06
More tables for `cds build`

swaldmann:
Just stumbled over our documentation for `cds build` and found some parts difficult to read because of some large walls of prose for configuration options, e.g. here:
https://cap.cloud.sap/docs/guides/deployment/custom-builds#build-task-properties

Let's make these MD tables, which I think is much more pleasant to read.

### DIFF
--- a/guides/deployment/custom-builds.md
+++ b/guides/deployment/custom-builds.md
@@ -19,8 +19,8 @@ status: released
 
 Build tasks are derived from the CDS configuration and project context. By default, CDS models are resolved from these sources:
 
-- _db_, _srv_, _app_ — [default root folders](../../get-started/#project-structure)
-- _fts_ and its subfolders when using [feature toggles](../extensibility/feature-toggles#enable-feature-toggles)
+- _db/_, _srv/_, _app/_ — [default root folders](../../get-started/#project-structure)
+- _fts/_ and its subfolders when using [feature toggles](../extensibility/feature-toggles#enable-feature-toggles)
 - CDS model folders and files defined by [required services](../../node.js/cds-env#services), including built-in ones
   - Examples: [persistent queue](../../node.js/queue#persistent-queue) or [MTX-related services](../multitenancy/mtxs#mtx-services-reference)
 - Explicit `src` folder configured in the build task


guides/deployment/custom-builds.md
```diff
@@ -15,18 +15,19 @@ status: released
 
 ## Build Configurations {#build-config}
 
-`cds build` executes _build tasks_ on your project folders to prepare them for deployment. Build tasks compile source files (typically CDS sources) and create the required artifacts, for example, EDMX files, SAP HANA design-time artifacts, and so on. By default, `cds build` dynamically determines the build tasks from the CDS configuration and from the project context. See [build task properties](#build-task-properties) for a concrete list of the different build task types.
+`cds build` runs _build tasks_ on your project folders to prepare them for deployment. Build tasks compile _source files_ (typically CDS sources) and create required artifacts, for example, EDMX files or SAP HANA design-time artifacts.
 
-The CDS model folders and files used by `cds build` are determined as follows:
+Build tasks are derived from the CDS configuration and project context. By default, CDS models are resolved from these sources:
 
-- Known root folders, by [default](../../get-started/#project-structure) the folders _db/, srv/, app/_,  can also be configured by [`folders.db`, `folders.srv`, `folders.app`](../../get-started/#project-structure).
```
swaldmann, line 22: let's not promote overriding these so much…

guides/deployment/custom-builds.md
```diff
@@ -15,18 +15,19 @@ status: released
...
-- The _src_ folder configured for the individual build task.
-- If [feature toggles](../extensibility/feature-toggles#enable-feature-toggles) are enabled: subfolders of the folder _fts_.
-- CDS Model folders and files defined by the [required services](../../node.js/cds-env#services) of your project. This also includes models used by required built-in CDS services, like [persistent queue](../../node.js/queue#persistent-queue) or [MTX related services](../multitenancy/mtxs#mtx-services-reference).
+- _db_, _srv_, _app_ — [default root folders](../../get-started/#project-structure)
+- _fts_ and its subfolders when using [feature toggless](../extensibility/feature-toggles#enable-feature-toggles)
+- CDS model folders and files defined by [required services](../../node.js/cds-env#services), including built-in ones
+  - Examples: [persistent queue](../../node.js/queue#persistent-queue) or [MTX-related services](../multitenancy/mtxs#mtx-services-reference)
+- Explicit `src` folder configured in the build task
```
swaldmann, line 26: Using `src` instead of _src_ here as that's a JSON config property, not a file path. Was a bit confused by that – details matter :)

guides/deployment/custom-builds.md
```diff
@@ -76,40 +77,40 @@ The following build tasks represent the default configuration dynamically determ
...
+| `nodejs`         | Node.js applications |
+| `java`           | Java applications |
+| `mtx-sidecar`    | [MTX](../multitenancy/mtxs)-enabled projects _with_ sidecar architecture.<br><br>[Learn more about **Multitenant Saas Application Deployment**](./to-cf){.learn-more} |
+| `mtx`            | MTX-enabled projects _without_ sidecar architecture (Node.js only). Required services are served by the Node.js application itself. |
+| `mtx-extension`         | MTX extension project (_extension.tgz_), which is required for extension activation using `cds push`. Extension point restrictions defined by the SaaS app provider are validated by default. If any restriction is violated the build aborts and the errors are logged.<br><br>The build task is created by default for projects that have `"cds": { "extends": "\<SaaS app name\>" }` configured in their _package.json_.<br><br>[Learn more about **Extending and Customizing SaaS Solutions**](../extensibility/customization){.learn-more} |
 
-  [Learn more about **Running Build Plugins**](#run-the-plugin){.learn-more}
```
swaldmann, line 100: Let's not link too much within a single guide, it's a short one and there's a ToC. Overly using cross-references increases cognitive load for our users → removed this

guides/deployment/custom-builds.md
```diff
@@ -15,18 +15,19 @@ status: released
 
 ## Build Configurations {#build-config}
 
-`cds build` executes _build tasks_ on your project folders to prepare them for deployment. Build tasks compile source files (typically CDS sources) and create the required artifacts, for example, EDMX files, SAP HANA design-time artifacts, and so on. By default, `cds build` dynamically determines the build tasks from the CDS configuration and from the project context. See [build task properties](#build-task-properties) for a concrete list of the different build task types.
```
swaldmann, line 18: >  See [build task properties](#build-task-properties) for a concrete list of the different build task types.

This is the first section in our guide anyway, no need to link to it in our introduction.

guides/deployment/custom-builds.md
daogrady: why not? Will this configuration be phased out?

guides/deployment/custom-builds.md
daogrady: that is a good point and a frequent source of confusion for me as well. I have seen all styles and a mixture thereof spread across all kinds of guides and sub pages. Do we have a style guide for this kind of things?

daogrady: tbh, I am not even sure the current formatting reflects it entirely. Personally, I like to add a slash to the end of paths to hint at it being a file path, so 

```md
- _fts/_ and its subfolders when using [feature toggless](../extensibility/feature-toggles#enable-feature-toggles)
```
but I guess I am a bit biased, too.

guides/deployment/custom-builds.md
```diff
@@ -15,18 +15,19 @@ status: released
...
 
-The CDS model folders and files used by `cds build` are determined as follows:
+Build tasks are derived from the CDS configuration and project context. By default, CDS models are resolved from these sources:
 
-- Known root folders, by [default](../../get-started/#project-structure) the folders _db/, srv/, app/_,  can also be configured by [`folders.db`, `folders.srv`, `folders.app`](../../get-started/#project-structure).
-- The _src_ folder configured for the individual build task.
-- If [feature toggles](../extensibility/feature-toggles#enable-feature-toggles) are enabled: subfolders of the folder _fts_.
-- CDS Model folders and files defined by the [required services](../../node.js/cds-env#services) of your project. This also includes models used by required built-in CDS services, like [persistent queue](../../node.js/queue#persistent-queue) or [MTX related services](../multitenancy/mtxs#mtx-services-reference).
+- _db_, _srv_, _app_ — [default root folders](../../get-started/#project-structure)
+- _fts_ and its subfolders when using [feature toggless](../extensibility/feature-toggles#enable-feature-toggles)
```
daogrady:
```suggestion
- _fts_ and its subfolders when using [feature toggles](../extensibility/feature-toggles#enable-feature-toggles)
```
(typo, unrelated to my suggestion of adding a trailing slash)

guides/deployment/custom-builds.md
```diff
@@ -15,18 +15,19 @@ status: released
...
+- CDS model folders and files defined by [required services](../../node.js/cds-env#services), including built-in ones
+  - Examples: [persistent queue](../../node.js/queue#persistent-queue) or [MTX-related services](../multitenancy/mtxs#mtx-services-reference)
+- Explicit `src` folder configured in the build task
 
-Feature toggle folders and required built-in service models will also be added if user defined models have already been configured as [_model_ option](#build-task-properties) in your build tasks.
+Feature toggle folders and required built-in service models will also be added if user-defined models have been configured as a [_model_ option](#build-task-properties) in your build tasks.
```
daogrady: I guess we would want `model` instead of _model_? 🙃 Also see below in the "tip"-box,

guides/deployment/custom-builds.md
```diff
@@ -49,7 +50,7 @@ For example, you want to configure the _src_ folder and add the default models.
 ```
 
 
- That way, the model paths will still be dynamically determined, but the _src_ folder is taken from the build task configuration. So you benefit from the automatic determination of models, for example, when adding a new external services, or when CAP is changing any built-in service configuration values.
+ This way, the model paths will still be dynamically determined, but the _src_ folder is taken from the build task configuration. You still benefit from the automatic determination of models – for example when adding a new external services or when CAP is changing any built-in service defaults.
```
daogrady, line 53: 🍪 using Gedankenstrich always earns brownie points from me! 🙂

guides/deployment/custom-builds.md
```diff
@@ -261,7 +258,7 @@ The `cds.build.Plugin` class provides methods for copying or writing contents to
 ```js [postgres/lib/build.js]
 await this.copy(path.join(this.task.src, 'package.json')).to('package.json');
 await this.write({
-  dependencies: { '@sap/cds': '^7', '@cap-js/postgres': '^1' },
+  dependencies: { '@sap/cds': '^9', '@cap-js/postgres': '^2' },
```
daogrady, line 261: 🙁 I guess this is a point of frequent maintenance... we probably don't have any mechanism to auto-upgrade these versions, do we?

guides/deployment/custom-builds.md
swaldmann: In the [linked section](https://cap.cloud.sap/docs/get-started/#project-structure), the explicit configuration is hidden behind a `<details>` block and there's a tip block below advising you should stick with the defaults. I think linking to  `get-started/#project-structure` is enough here, as that information isn't relevant for almost all projects.

daogrady: If we keep the link to `get-started/#project-structure` then I am absolutely okay with this change. 👍

guides/deployment/custom-builds.md
swaldmann: Yes, see [here](https://github.tools.sap/cap/docs/blob/main/resources/authoring.md#typographical-convention). Our UA colleagues have aligned almost all of it by now I think.

guides/deployment/custom-builds.md
daogrady: > If we keep the link 


....and so you do. I should have looked into the changed code, rather than just this comment on the conversation page. 🙂

guides/deployment/custom-builds.md
swaldmann: Yep, good catch!

guides/deployment/custom-builds.md
daogrady: Great to know, thanks! 🔖

guides/deployment/custom-builds.md
swaldmann: We do! But this snippet is difficult, as it's in JavaScript. If it was in a JSON our renovate bot would probably catch it. Can follow up on that.

guides/deployment/custom-builds.md
swaldmann: Fixed in https://github.com/cap-js/docs/pull/2011/commits/8c109905fbda4cf357b77b3d303d42a25e9b0cd1

guides/deployment/custom-builds.md
swaldmann: Good idea with the trailing /, did that in https://github.com/cap-js/docs/pull/2011/commits/730cb95a7143ac1b5087cff332d646f809c585cf

PR #2016, opened 2025-08-11
docs: hana cloud maintenance

johannes-vogel:


guides/databases-hana.md
```diff
@@ -53,7 +53,11 @@ The datasource for SAP HANA is then auto-configured based on available service b
 
 :::
 
+## Maintenance
 
+The database services are validated against the maintained versions of SAP HANA Cloud. It's not guaranteed that outdated versions are fully functional with the latest database services.
```
agoerler:
```suggestion
The database services are validated against the latest maintained QRC version of SAP HANA Cloud. It's not guaranteed that outdated versions are fully functional with the latest database services.
```

(to avoid the impression that we would test with older versions)

agoerler (approved):

guides/databases-hana.md
```diff
@@ -53,7 +53,11 @@ The datasource for SAP HANA is then auto-configured based on available service b
 
 :::
 
+## Maintenance
 
+The database services are validated against the latest maintained QRC version of SAP HANA Cloud. It's not guaranteed that outdated versions are fully functional with the latest database services.
```
renejeglinsky: How about adding this to the warning starting in line 12?

renejeglinsky: If that should not be an option, the title should make clear that it's about the db service. Right now my expectation was to read about the maintenance (strategy) of SAP HANA or about the maintenance of the guide maybe.

johannes-vogel: i think we can combine. we don't validate against anything else but hana cloud latest version

guides/databases-hana.md
```diff
@@ -9,10 +9,14 @@ status: released
...
 
-::: warning
+::: warning Validation strategy
 
 CAP isn't validated with other variants of SAP HANA, like "SAP HANA Database as a Service" or "SAP HANA (on premise)".
 
+The database services are validated against the latest maintained QRC version of SAP HANA Cloud. It's not guaranteed that outdated versions are fully functional with the latest database services.
+
+[See the official SAP HANA Cloud documentation for their maintenance strategy.](https://help.sap.com/docs/HANA_CLOUD_CN/1f64fe39189f4176bf659e737d62222a/6ced4d164e234b74aa9bea82435ce9a8.html){.learn-more}
+
```
renejeglinsky, line 19: @johannes-vogel Could you please check if that matches what you want to achieve?

renejeglinsky (approved):

PR #2025, opened 2025-08-13
Add steps to running built (TS) projects locally

daogrady:
Following email conversation with Timo B. and Fabian M.

### DIFF
--- a/guides/deployment/custom-builds.md
+++ b/guides/deployment/custom-builds.md
@@ -323,7 +323,7 @@ cds build --for postgres
 
 <div class="impl node">
 
-If you want to test-run the application as it would run in the cloud, start it from the build result folder `gen/srv/`. This directory will be the app root on cloud platforms and therefore needs to be the root for local serving too:
+The artifacts deployed to the various cloud platforms are generated in the `gen/srv/` folder. So, to test the application as it runs on the cloud start your application from the `gen/srv/` folder:
 
 ```sh
 cds build       # to create the build results, followed by either:
--- a/node.js/typescript.md
+++ b/node.js/typescript.md
@@ -149,7 +149,8 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
 
 ## Running Built Projects Locally
 
-If you want to test-run the application as it would run in the cloud, start it from the build result folder `gen/srv/`. This directory will be the app root on cloud platforms and therefore needs to be the root for local serving too:
+The artifacts deployed to the various cloud platforms are generated in the `gen/srv/` folder. So, to test the application as it runs on the cloud start your application from the `gen/srv/` folder:
+
 
 ```sh
 cds build     # to create the js files


node.js/typescript.md
```diff
@@ -103,6 +103,22 @@ In general, `tsx` is the better choice, as `tsx` is considerably faster than `ts
 See a closer [comparison](https://tsx.is/faq#how-does-tsx-compare-to-ts-node) between the two of them.
 :::
 
+## Running Built Projects Locally
+
+When building a TypeScript project and running it locally, you need to serve the project from the build folder `gen/srv/`. This directory will be the app root on cloud platforms and therefore needs to be the root for local serving too:
```
chgeo:
```suggestion
If you want to test-run the application as it would run in the cloud, start it from the build result folder `gen/srv/`. This directory will be the app root on cloud platforms and therefore needs to be the root for local serving too:
```

chgeo:
@daogrady I have move the section to the build page where it logically belongs (+added a Java variant there).  In the TS page, there is a smaller snippet + a link to the build page).

guides/deployment/custom-builds.md
```diff
@@ -318,3 +318,31 @@ cds build --for postgres
...
 > See also the command line help for further details using `cds build --help`.
+
+## Test-Run Built Projects Locally {#test-run}
+
+<div class="impl node">
+
+If you want to test-run the application as it would run in the cloud, start it from the build result folder `gen/srv/`. This directory will be the app root on cloud platforms and therefore needs to be the root for local serving too:
+
+```sh
+cds build       # to create the js files, followed by either:
```
daogrady: not sure if "the js files" makes sense in a context where js files would be the input

node.js/typescript.md
```diff
@@ -147,6 +147,17 @@ A dedicated build task for `cds build` is provided as part of the `cds-typer` pa
 
 [Learn more about integrating it into your build process.](../tools/cds-typer#integrate-into-your-build-process){.learn-more}
 
+## Running Built Projects Locally
+
+If you want to test-run the application as it would run in the cloud, start it from the build result folder `gen/srv/`. This directory will be the app root on cloud platforms and therefore needs to be the root for local serving too:
```
daogrady: should we just link to the section in the build doc? This seems to be a duplicate now

guides/deployment/custom-builds.md
chgeo: Good point, will fix.

node.js/typescript.md
chgeo: That was my first version, but a single link looked too sparse which is a recipe for being ignored by readers.  A bit of redundancy is good I guess.

daogrady: Makes sense. I guess we do not have a mechanism to include a section from another page locally, like an `<iframe>`-light?

chgeo: There is an [include mechanism](https://vitepress.dev/guide/markdown#markdown-file-inclusion).  It looks over the top here though.  There is a danger of fragmentation and illegibility when overused.

daogrady: True, good to know that mechanism anyway.

smahati (approved):

PR #2037, opened 2025-08-19
feat: Status-Transition Flows

sjvans:


guides/flows.md
```diff
@@ -0,0 +1,308 @@
...
+[[toc]]
+
+
+
+## Enabling Flows
+
+Status-transition flows are part of the CAP Node.js core (`@sap/cds`).
+Hence, there are no steps required to enable the feature.
+
+To get started with the flow feature in CAP Java, simply add the [cds-feature-flow](https://central.sonatype.com/artifact/com.sap.cds/cds-feature-flow) dependency to your `srv/pom.xml` file:
```
agoerler:
```suggestion
Status-transition flows are supported by both CAP runtimes. 

In CAP Node.js support for flows is part of the CAP Node.js core (`@sap/cds`). 

For CAP Java, support for flows is provided by the feature [cds-feature-flow](https://central.sonatype.com/artifact/com.sap.cds/cds-feature-flow). Enable it by adding this dependency to your _srv/pom.xml_ file:
```

guides/assets/flows/xtravels-flow-simple.svg
agoerler, line 1: * The transition open -> reject -> canceled is surprising. I would expect a transition to "rejected".
* I think the appearance of this diagram is a bit sloppy. I'd like to suggest that we use a clearer representation. Maybe we want to adhere to the style of a TAM state diagram:
Screenshot 2025-11-24 at 14 29 15

guides/flows.md
```diff
@@ -0,0 +1,308 @@
...
+  // Define actions
+  actions {
+    action acceptTravel();
+    action rejectTravel();
+    action deductDiscount( percent: Percentage not null ) returns Travels;
+  };
+
+  // Define flow (+ status check for "deductDiscount")
+  annotate Travels with @flow.status: Status actions {
+    acceptTravel    @from: #Open  @to: #Accepted;
```
agoerler: We firstly should declare the enum that has the enumerals # Open, #Accepted and #Canceled.

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+The following is an extract of the relevant parts of the domain model:
+
+::: details `db/schema.cds`
+```cds [db/schema.cds]
+// db/schema.cds
+namespace sap.capire.travels;
+
+entity Travels : managed {
+  // [...]
+  Status       : Association to TravelStatus default 'O';
```
agoerler:
```suggestion

  @flow.status
  status       : Association to TravelStatus default 'O';
```

agoerler: Although possible, I think modeling with a code list + enum is pretty advanced. From a didactical perspective maybe rather use a plain enum. You could show the codelist variant later.

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Ingredients
+
+Flows are made up of a _status element_ and a set of _flow actions_ that define the transitions between different states. 
+
+#### `@flow.status`
+
+To model a flow, one of the entity fields needs to be annotated with `@flow.status`. This field must be one of the following:
+
+- A String enum consisting of keys and values
```
agoerler: a numeric enum should also be fine, not sure

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+- A String enum with only symbols
+- A Codelist entity with the key `code` if localization is needed (`code` must be one of the two above)
+
+::: tip The status field should be `@readonly` and have a default value.
+We recommend to always use `@flow.status` in combination with `@readonly`.
```
agoerler: Then please make it readonly in the example here as well. Also please provide a default value.

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+This ensures that the status element is immutable from the client side, giving the service provider full control over all state transitions.
+As no initial state can be provided on `CREATE`, there should be a default value.
```
agoerler: If a flow starts with always in the same state a default is appropriate. Do we need to have a means to model which start states are permitted? Do we need "create" actions in this case?

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+:::
+
+If you annotate `@flow.status: <element name>` on entity level (as in the example above), the annotation get's propagated to the respective element and the element is also annotated with `@readonly`.
+
+Notes:
+- This annotation is mandatory.
+- The annotated element must be either an enum or an association to a code list.
+- Only one status element per entity is supported.
+- Draft-enabled entities are supported.
```
agoerler: ... but the flows are only applied to the active version

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+:::
+
+After declaring a `@flow.status`, use the following annotations on bound actions to model the transitions of the flow:
+
+#### `@from`
+
+- Defines valid entry states for the action
+- Validates whether the entity is in a valid entry state before executing the action (the current state of the entity must be included in the states defined here)
+- Can be a single value or an array of values (each element must be a value from the status enum)
+- UI annotations to allow/disallow buttons and to refresh the page are automatically generated
```
agoerler:
```suggestion
- UI annotations to allow/disallow buttons and to refresh the page are automatically generated for UI5
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### `@to`
+
+- Defines the desired target state of the entity after executing the action
+- Changes the state of the entity to the value defined in this annotation after executing the action
+- Must be a single value from the status enum
+
+Both annotations are optional, but at least one is required to mark an action as a flow action. You can use either one or both, depending on the desired behavior. If you use both, you don't need to implement any handlers for the actions, as all generic handlers are registered automatically.
```
agoerler: Not sure, but it might be the case that somewhere we mention that you have to provide an on-handler for actions and functions. This would then be an exception worth to mention.

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+
+### Generic Handlers
+
+No handler implementations are required for the actions, as the generic handlers are registered automatically.
```
agoerler:
```suggestion
No handler implementations are required for the flow actions, as generic handlers are registered automatically.
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### `before`
+
+From the `@from` annotation, a handler is registered that performs the check described above.
+The default `before` handler will validate that your entity instance is in a valid entry state, meaning the current state of the entity must match one of the states specified in the `@from` annotation.
+If the entity is not in a valid state, the request fails with a suitable error.
+A validation error will return a `409` HTTP Status Code.
```
agoerler:
```suggestion
A validation error will return a `409 Conflict` HTTP Status Code.
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### `on`
+
+An empty handler is registered if none is provided.
+That is, The default `On` handler will complete the action for void return types.
```
agoerler: > `On`

inconsitent casing: on [149](https://github.com/capire/docs/pull/2037/files#diff-015deabd759831433a87957f07c0b0842714fb64b436543e38210e675feb43b4R149) we use `before` in lower case.

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+#### `after`
+
+From the `@to` annotation, a handler is registered that changes the state of the entity to the desired target state.
+The default `after` handler will process the transition by automatically updating the status element of the entity to the target state defined in the `@to` annotation.
+For example, if the current state is `Open` and the target state is `Accepted`, the `after` handler will update the status to `Accepted` after the action is executed.
+This ensures that the state transition is consistently applied without requiring custom logic.
+
+::: tip Generic handlers are not executed for drafts.
+For example, calling `acceptTravel()` on a `Travels` entity that is currently being _edited_ has no effect.
```
agoerler:
```suggestion
For example, calling `acceptTravel()` on a `Travels` entity that is currently being _edited_, i.e. is in _inactive_ state, has no effect.
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+No action handler implementation is needed for simple transitions, as the flow feature's default action handler takes care of the logic.
+It validates that the entry state is `Open` and then transitions the status to `Accepted` or `Canceled` respectively.
+For more complex logic, custom handlers may still be required, as explained later.
+
+
+### Ingredients
+
+Flows are made up of a _status element_ and a set of _flow actions_ that define the transitions between different states. 
+
+#### `@flow.status`
```
agoerler:
```suggestion
#### The Status Element
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+- `null` is **not** a valid state. Model your empty state explicitly.
+
+::: warning Only simple projections are supported.
+In other words, the entity must be _writable_.
+Further, renaming the status element is currently not supported.
+:::
+
+After declaring a `@flow.status`, use the following annotations on bound actions to model the transitions of the flow:
+
+#### `@from`
```
agoerler: This is a bit brief as a heading. Also, I think we should rather describe `@from` and `@to` in a single chapter
 "Modeling State Transitions". 
 
 Firstly, we could describe that `@from` and `@to` specify valid state transitions.
 
 Then we could detail out that we generically handle flows so that we check `@from` before the flow action and change the state according to `@to` after the flow action.

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+The default `after` handler will process the transition by automatically updating the status element of the entity to the target state defined in the `@to` annotation.
+For example, if the current state is `Open` and the target state is `Accepted`, the `after` handler will update the status to `Accepted` after the action is executed.
+This ensures that the state transition is consistently applied without requiring custom logic.
+
+::: tip Generic handlers are not executed for drafts.
+For example, calling `acceptTravel()` on a `Travels` entity that is currently being _edited_ has no effect.
+:::
+
+
+### `$flow.previous`
```
agoerler:
```suggestion
### Reverting to Previous State
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+    acceptTravel    @from: #InReview           @to: #Accepted;
+    rejectTravel    @from: #InReview           @to: #Canceled;
+    deductDiscount  @from: #Open;
+  };
+
+}
+```
+
+To transition to the previous state, each entity with a flow that includes at least one transition to `$flow.previous` is automatically augmented with the necessary data structure to record the transitions.
+Specifically, the entities are appended with the aspect `sap.common.FlowHistory`.
```
agoerler:
```suggestion
Specifically, the entities are extended with the aspect `sap.common.FlowHistory`, which includes `transitions_` composition that captures the history of state transitions.
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+::: tip Transitions are excluded from projections.
+The `transitions_` composition that is automatically appended to the base entity, is also automatically excluded from all projections.
+:::
+
+
+
+## Extending Flows
+
+The flow annotations can be used for all basic flows that should be modelled, when there is the need for more complex cases, they can be implemented with custom event handlers.
```
agoerler:
```suggestion
The flow annotations can be used for all basic flows that should be modelled. When there is the need for more complex cases, they can be implemented with custom event handlers.
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+Let's look at some use-cases for extending flows with custom event handlers:
+- If the entry state validation depends on additional conditions, you could implement the logic in a custom `before` handler. 
+- If the action's return type is anything other than `void`, you must implement a custom `on` handler.
+- If you have multiple target states depending on certain conditions, you must implement a custom `on` handler, and you must **not** use the `@flow.to` annotation.
```
agoerler:
```suggestion
- If you have multiple target states depending on certain conditions, you must implement a custom `on` handler, and you must **not** use the `@to` annotation.
```

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+- If you want to run logic like contacting an external system on a state transition, consider implementing a custom `on` handler for the respective action.
```
agoerler:
```suggestion
```

(You can always do this in a custom handler)

guides/flows.md
```diff
@@ -0,0 +1,349 @@
...
+
+  public WithdrawTravelHandler(PersistenceService persistenceService) {
+    this.persistenceService = persistenceService;
+  }
+
+  @Before(entity = Travel_.CDS_NAME)
+  public void check24HoursBeforeTravel(final TravelWithdrawTravelContext context, CqnStructuredTypeRef travelRef) {
+    Travel travel = ((ApplicationService) context.getService()).run(
+        Select.from(travelRef).columns(Travel_.BEGIN_DATE)).first(Travel.class)
+      .orElseThrow(() -> new ServiceException(ErrorStatuses.BAD_REQUEST, "TRAVEL_NOT_FOUND"));
```
agoerler, line 80:
```suggestion
  public void check24HoursBeforeTravel(final TravelWithdrawTravelContext context, Travel_ travelRef) {
    Travel travel = context.getService().run(Select.from(travelRef).columns(t -> t.beginDate())).first();
```

guides/assets/flows/xtravels-flow-simple.svg
sjvans: > * The transition open -> reject -> canceled is surprising. I would expect a transition to "rejected".

that is existing xtravels content

> * I think the appearance of this diagram is a bit sloppy. I'd like to suggest that we use a clearer representation. Maybe we want to adhere to the style of a TAM state diagram:

agree to disagree :) TAM is super busy, imho. but i defer to @renejeglinsky

guides/flows.md
sjvans: it's in the collapsed `db/schema.cds` above. maybe it wasn't pushed yet...

guides/flows.md
sjvans: you mean `status : String enum {...};`? i get your point, but i also see the value in just copying from xtravels to stay as close to the sample as possible...

btw, `@flow.status` is not needed here

guides/flows.md
sjvans: `@readonly` is added automatically in the sample (see sentence below) and `sap.capire.travels.Travels.Status` already has `default : 'O'`.

guides/flows.md
sjvans: create/ crud actions are currently in discovery

guides/flows.md
sjvans: outdated

guides/flows.md
sjvans: outdated

guides/flows.md
sjvans: 👍 couldn't find it but i'll mention it here

guides/flows.md
sjvans: outdated

guides/flows.md
sjvans: outdated

guides/flows.md
sjvans: i excluded the example for now. i think we need to review that whole scenario ;)

guides/flows.md
sjvans: Screenshot 2025-11-25 at 01 07 14

i think i kind of have it this way already. or i'm misunderstanding... my idea was "modeling flows" with the sample and then details in "annotations"/ "generic handlers"

guides/providing-services.md
```diff
@@ -1239,8 +1239,217 @@ Programmatic usage would look like this for Node.js:
...
+    acceptTravel    @from: #InReview           @to: #Accepted;
+    rejectTravel    @from: #InReview           @to: #Canceled;
+    deductDiscount  @from: #Open;
+  };
+
+}
+```
+
+Entities with flows that include at least one transition to `$flow.previous` are automatically extended with the `sap.common.FlowHistory` aspect, which includes `transitions_` composition that captures the history of state transitions.
+This automatic entity extending can be deactivated via <Config>cds.features.history_for_flows: false</Config>.
```
renejeglinsky, line 1436: Is this only configurable for Node.js?

agoerler: How does `$flow.previous` work in CAP Node.js if `ds.features.history_for_flows` is set to `false`?

guides/assets/flows/xtravels-flow-simple.svg
agoerler: > TAM is super busy

TAM is the "lingua franca" at SAP and hence it would be helpful for internal readers to ahere to TAM. But I am not sure wether it's also used customer-facing.

I think the main difference is the usage of round-edge rectangles instead of elipses. And that there are start-of-/end-of-activity markers.

guides/assets/providing-services/simple-flow-example-sflight.drawio.svg
agoerler, line 1: "new" 
* "new" is no action -> "create" ?
* maybe use a TAM-style solid circle to mark the start of activity.

guides/providing-services.md
sjvans: > Is this only configurable for Node.js?

cds adds it for both runtimes → only a cds config

sjvans: > How does `$flow.previous` work in CAP Node.js if `ds.features.history_for_flows` is set to `false`?

it doesn't if you don't add aspect FlowHistory manually. i'll add a hint.

guides/assets/providing-services/simple-flow-example-sflight.drawio.svg
sjvans: this diagram isn't used (in my docs)

guides/providing-services.md
renejeglinsky: > > Is this only configurable for Node.js?
> 
> cds adds it for both runtimes → only a cds config

Also then, even if it looks the same, I'd propose to show the `<Config java>...` variant as well.

sjvans: glad to, but i don't know how to set cds configs via java's ymls 😅

beckermarc: I don't think that's possible :) Some build-related configuration for Java also needs to be put in `.cdsrc.json` and others, as it is picked up by the `cds` CLI.

renejeglinsky: Alright, understood. Thanks!

guides/providing-services.md
```diff
@@ -1239,8 +1239,217 @@ Programmatic usage would look like this for Node.js:
...
+The following is an extract of the relevant parts of the domain model:
+
+::: details `db/schema.cds`
+```cds [db/schema.cds]
+// db/schema.cds
+namespace sap.capire.travels;
+
+entity Travels : managed {
+  // [...]
+  Status       : Association to TravelStatus default 'O';
```
renejeglinsky, line 1281: @sjvans The `@readonly` annotation is missing here that is contained in the sample. Also it is mentioned later (line 1334) as needed. So should we add it?

renejeglinsky: added to https://github.com/capire/docs/pull/2245
Can be discusses there.

PR #2041, opened 2025-08-20
Searchable @Common.Text Annotation

etimr:


johannes-vogel (approved):
LGTM

@MattSchur does that also hold true for Java?

guides/providing-services.md
```diff
@@ -426,6 +426,23 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
 You can explicitly annotate calculated elements to make them searchable, even though they aren't searchable by default. The virtual elements won't be searchable even if they're explicitly annotated.
 :::
 
+#### The `@Common.Text` Annotation 
+
+If an entity has an element annotated with the `@Common.Text` annotation, then the property that holds the display text is now by default added to the searchable elements. Example:
```
sjvans:
```suggestion
If an entity has an element annotated with the `@Common.Text` annotation, then the property that holds the display text is by default added to the searchable elements. Example:
```

guides/providing-services.md
```diff
@@ -426,6 +426,23 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
 You can explicitly annotate calculated elements to make them searchable, even though they aren't searchable by default. The virtual elements won't be searchable even if they're explicitly annotated.
 :::
 
+#### The `@Common.Text` Annotation 
+
+If an entity has an element annotated with the `@Common.Text` annotation, then the property that holds the display text is by default added to the searchable elements. Example:
```
MattSchur:
```suggestion
The `@Common.Text` annotation allows you to specify a property that holds a text to be displayed on the UI instead of the value of the annotated property. To enhance the user experience, the property that holds the display text is searched by default:
```

guides/providing-services.md
```diff
@@ -426,6 +426,23 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+
+```cds
+entity Books : cuid {
+  title  : String;
+  @Common.Text : author.name
+  author : Association to Author;
+}
+```
+The default searchable elements of `Books` are `title` and `author.name`.
```
MattSchur:
```suggestion
Here the value of the name property of the Authors entity is displayed on the UI instead of the author association and is also searched automatically, without requiring additional configuration.
```

guides/providing-services.md
```diff
@@ -426,6 +426,23 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+#### The `@Common.Text` Annotation 
+
+If an entity has an element annotated with the `@Common.Text` annotation, then the property that holds the display text is by default added to the searchable elements. Example:
+
+```cds
+entity Books : cuid {
+  title  : String;
+  @Common.Text : author.name
+  author : Association to Author;
+}
```
MattSchur, line 440:
```suggestion
}
entity Author : cuid {
  name : String;
}
```

guides/providing-services.md
```diff
@@ -426,6 +426,23 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+```
+The default searchable elements of `Books` are `title` and `author.name`.
+
+::: warning
+Do not use both `@cds.search` and `@Common.Text` annotations on the same entity, unless `@cds.search` is used only to exclude fields. If both are present and `@cds.search` includes fields, the `@Common.Text` annotation will be ignored.
```
MattSchur: Is this specific to `@Common.Text`? AFAIK `@cds.search` with include fields overrides all default search elements

guides/providing-services.md
etimr: @MattSchur the subsection `The @Common.Text Annotation` belongs to the `Searching Data` section, right after `@cds.search`.  I think `@Common.Text` should be described here only from the search perspective, not as UI annotation. 

If we need to describe what `@Common.Text` means as UI annotation, we can add it here https://cap.cloud.sap/docs/advanced/odata, where it mentioned several times. 

Also the exact same text is used in the Java release notes and will be used in the Node release notes as well with the link to this documentation. I think it should be rephrased with the focus on the search.

guides/providing-services.md
etimr: > Is this specific to `@Common.Text`? AFAIK `@cds.search` with include fields overrides all default search elements

`@cds.search: { author }` will add all searchable elements of the associated `Authors` entity to all searchable elements of the `Books` entity, not overwrite them. Mentioning explicitly, that `@cds.search` with included fields has a priority over `@Common.Text`, allows to avoid confusion in the case both annotations are used.

guides/providing-services.md
MattSchur: OK, I didn't notice that this is a subsection of Search

guides/providing-services.md
MattSchur: If I add `@cds.search: { description }`, is `title` still searched?

sjvans: no, then only `description` is searched.
there are three modes:
- including: as soon as one own property is listed, only the listed properties are searched → @common.text is ignored
- extending: adds the searchable properties of the target to the list → own properties of the source remain, but @common.text is ignored nevertheless
- excluding: removes single (own) properties from the effective list → @common.text is respected

guides/providing-services.md
```diff
@@ -426,6 +426,26 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+  author : Association to Author;
+}
+entity Author : cuid {
+  name : String;
+}
+```
+The default searchable elements of `Books` are `title` and `author.name`.
+
+::: warning
+Do not use both `@cds.search` and `@Common.Text` annotations on the same entity, unless `@cds.search` is used only to exclude fields. If both are present and `@cds.search` includes fields, the `@Common.Text` annotation will be ignored.
```
sjvans:
```suggestion
`@cds.search` takes precedence over `@Common.Text`. As a result, `@Common.Text` is ignored as soon as `@cds.search` defines anything in an including mode (that is, not exclusively for excluding properties).
```
just a suggestion for rephrasing, content is fine

guides/providing-services.md
```diff
@@ -426,6 +426,26 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
 You can explicitly annotate calculated elements to make them searchable, even though they aren't searchable by default. The virtual elements won't be searchable even if they're explicitly annotated.
 :::
 
+#### The `@Common.Text` Annotation 
+
+If an entity has an element annotated with the `@Common.Text` annotation, then the property that holds the display text is by default added to the searchable elements. Example:
```
sjvans:
```suggestion
If an entity has an element annotated with the `@Common.Text` annotation, then the property that holds the display text is added to the list of searchable elements (see exception below).

For example, with the following modeling, the list of searchable elements of `Books` are `title` and `author.name`:
```

guides/providing-services.md
```diff
@@ -426,6 +426,26 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+entity Books : cuid {
+  title  : String;
+  @Common.Text : author.name
+  author : Association to Author;
+}
+entity Author : cuid {
+  name : String;
+}
+```
+The default searchable elements of `Books` are `title` and `author.name`.
```
sjvans:
```suggestion
```

guides/providing-services.md
```diff
@@ -426,6 +426,26 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+
+::: warning
+`@cds.search` takes precedence over `@Common.Text`. As a result, `@Common.Text` is ignored as soon as `@cds.search` defines anything in an including mode (that is, not exclusively for excluding properties).
+:::
+
```
sjvans, line 449:
```suggestion

To illustrate the above:
- `@cds.search: { title: false }` on `Books` would only exclude properties such that `author.name` would still be searched
- `@cds.search: { title }` on `Books` defines an include list such that `author.name` is not searched as, in this mode, `@cds.search` is expected to include all that shall be searched → `@cds.search: { title, author.name }`

```

sjvans (approved):

guides/providing-services.md
```diff
@@ -426,6 +426,31 @@ Searches all elements of type `String` excluding the element `isbn`, which leave
...
+  @Common.Text : author.name
+  author : Association to Author;
+}
+entity Author : cuid {
+  name : String;
+}
+```
+
+::: warning
+`@cds.search` takes precedence over `@Common.Text`. As a result, `@Common.Text` is ignored as soon as `@cds.search` defines anything in including mode (that is, not exclusively for excluding properties).
```
renejeglinsky: I'm not sure what the removed part wanted to say. Was it about mixed (include + exclude) definitions in @cds.search?
```suggestion
::: warning `@cds.search` takes precedence over `@Common.Text`
As a result, `@Common.Text` is ignored as soon as `@cds.search` defines anything in including mode.
```

renejeglinsky: @etimr Could you check my suggestion?

sjvans: @renejeglinsky see https://github.com/capire/docs/pull/2041#discussion_r2300068857

there is include, extend, and exclude. as soon as not all are excludes, @common.search is ignored.

renejeglinsky:
```suggestion
::: warning `@cds.search` takes precedence over `@Common.Text`
As a result, `@Common.Text` is ignored as soon as `@cds.search` defines anything in including mode. Only if you exclusively exclude properties using `@cds-search`, the `@Common.Text` is kept.
```

PR #2052, opened 2025-08-22
Details page for cql-select rule

chgeo:


chgeo:
@daogrady maybe you can take it from here.

For local testing, make sure to install the plugin into capire:
```sh
npm i .../eslint-plugin-cds --no-save
```

daogrady:
@smahati I think this should now be ready for review. I can no longer give a qualified review, as I have contributed content myself. Can you please give it a check for wording, formatting, and legibility? Thanks!

swaldmann (approved):

tools/cds-lint/rules/use-cql-select-template-strings.md
```diff
@@ -0,0 +1,46 @@
+---
+status: released
+---
+
+<script setup>
+  import PlaygroundBadge from '../components/PlaygroundBadge.vue'
+</script>
+
+# use-cql-select-template-strings
```
renejeglinsky, line 9: This too long for the ToC. How about only using `cql-select-template-strings`? For other rules we also do not have a prefix `use-` established. So everything without prefix `no-` can be considered a `use-` rule. Or?
image

tools/cds-lint/rules/case-sensitive-well-known-events.md
```diff
@@ -0,0 +1,43 @@
+---
+status: released
+---
+
+<script setup>
+  import PlaygroundBadge from '../components/PlaygroundBadge.vue'
+</script>
+
+# case-sensitive-well-known-events
```
renejeglinsky, line 9: This is too long for the ToC. Would `case-sensitive-events` be sufficient?

renejeglinsky:
The overview page doesn't reflect the JS rules yet:
image

tools/cds-lint/rules/use-cql-select-template-strings.md
daogrady: That's the actual, technical [name of the rule](https://github.tools.sap/cap/eslint-plugin-cds/blob/main/lib/conf/js/all.js#L4-L8) as users have to enable them in their config, so that would not reflect reality. 😕

renejeglinsky: Could we rename the rule? Not because of capire but to be consistent with naming of rules (and not using `use-` as prefix until now)?

daogrady:
> What are project-based rules? Do we have any mentioned here? This is not part of the tables and only found in the details note. BTW: Should this details note be in the general part of the overview page and not inside the Model rules section?
> 
> image

I believe project-based rules are the same as environment rules: checks, that do not target specific files (.cds or .js), but inspect the project's environment for integrity.

tools/cds-lint/rules/use-cql-select-template-strings.md
daogrady: 🤔 we could probably rename it to `cql-template-strings`. But that would be kind of a breaking change in `eslint-plugin-cds`, as it would invalidate the configuration of users referring to the rule as `use-cql-select-template-strings` already. (We presented this rule set at recap, so people were aware of them even before this piece of documentation)

I can rename the rule and add the old name as an alias, but this will then only be included in the next release of the plugin.

tools/cds-lint/rules/case-sensitive-well-known-events.md
daogrady: (same here as above btw)

daogrady: `case-sensitive-events` would not be correct. Events are always case sensitive and users are free to mix cases however they like. The rule is specifically targeting "our" events, which have to be written in all caps.

tools/cds-lint/rules/use-cql-select-template-strings.md
renejeglinsky: This is just my take. Let's not block the PRs by that. If you make it part of your discussions then every result is fine for me.

renejeglinsky (approved):

tools/cds-lint/rules/use-cql-select-template-strings.md
daogrady: Thanks! I have opened a [PR](https://github.tools.sap/cap/eslint-plugin-cds/pull/639) on the plugin to implement the suggested name change. If the `case-sensitive-well-known` rule poses a serious issue for the ToC, do let me know; we can probably find an alternative name.

PR #2060, opened 2025-08-27
fix(`@cds.search`): wrong inclusion logic

patricebender:
this did not reflect what is actually happening.
In reality, providing elements to the `@cds.search` annotation, effectively removes all default searchable elements from being searched. To avoid confusion and future tickets, a promiment warning clarifying this behavior is added with this change.

@t-bonk @beckermarc is this in line with how `@cds.search` works in cap-java?


sjvans (changes requested):
afaik, `@cds.search: { author }` extends and doesn't replace

patricebender:
> afaik, `@cds.search: { author }` extends and doesn't replace

I agree this should be the behaviour. Back in the day when I implemented it, everybody agreed on [the specification I came up with](https://github.com/cap-js/cds-dbs/pull/590/files#diff-6fa5855d53f77dbce5e078b8d8f0970e04f4ecc7860e0cec9f07a6c948374992R249).

Also, _only_ associations have an additive character, while including any other field (also via path expression) overrides the default searchable elements entirely. Not really consistent. 

@johannes-vogel @sjvans What about adding a calculated element to the search? Should this also have an additive character or should it rather overwrite every other searchable element?

> You are working on the fix, don't you?

yes, but we still need to update the documentation to point out those details.

etimr:
Just adding some info from Adrian:
```
@cds.search: { author } entity Books { ... }
Bedeutet: durchsuche alle default-search-Elemente von Books und durchsuche zusätzlich Author. (Was von Author durchsucht wird hängt davon ab, ob Author wiederum eine @cds.search Annotation hat).

Wenn @cds.search nur Assoziationen enthält wird addiert. Wenn irgendwelche nicht-Assoziationen dabei sind wird überschrieben.
```
But I think we are replacing, not adding in `cap-js/cds-dbs`.

sjvans:
two cents from a non-expert 😅

> 1. What about adding a calculated element to the search? Should this also have an additive character or should it rather overwrite every other searchable element?

non-additive, as a calculated element is not an association (= something that can be expanded).

> 2. From [this comment](https://github.com/capire/docs/pull/2041#discussion_r2300068857): "as soon as one own property is listed, only the listed properties are searched". Including `@cds.search { books.title }` does then also _extend_ the search?

no. `books.title` is also an "own property", just not of Authors. may "scalar property" would better describe it. but i'd defer to your compiler expertise.

patricebender:
Thanks @MattSchur & @sjvans - node.js behavior should be in line with doc and Matthias' comment once https://github.com/cap-js/cds-dbs/pull/1355 is merged

closing this

PR #2064, opened 2025-08-28
Production profile in security guide

BraunMatthias:
As a reaction to a security incident. Several customers activated the index page in production.

### DIFF
--- a/guides/security/aspects.md
+++ b/guides/security/aspects.md
@@ -693,6 +693,23 @@ For instance, this is true for [authorizations](#secure-authorization) or applic
 It's recommended to ensure security settings by automated integration tests.
 :::
 
+CAP provides some features that are suitable for development only such as 
+- Index Page
+- Mock Users
+- Developer Dashboard (Java only)
+
+These features are deactivated in the production profile by default.
+
+::: warning 
+**Do not manually enable features for production that are disabled by the production profile**, as this could introduce serious security vulnerabilities.
+:::
+
+[Learn more about production profiles in Java](../../java/developing-applications/configuring#production-profile){.learn-more}
+[Learn more about production profiles in Node.js](../../node.js/cds-env#profiles){.learn-more}
+
+
+
+
 
 ### Fail Securely { #fail-securely }
 <!-- #SEC-239 -->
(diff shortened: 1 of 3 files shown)


BraunMatthias:
@sjvans : I modified the statement in Node.js/"Toggle Generic Index Page". Please check!

node.js/cds-server.md
```diff
@@ -261,7 +261,7 @@ The built-in CORS middleware can be enabled explicitly with <Config>cds.server.c
 
 ### Toggle Generic Index Page
 
-The default generic _index.html_ page is not served if `NODE_ENV` is set to `production`. Set <Config>cds.server.index: true</Config> to restore the generic index page in production.
+The default generic _index.html_ page is not served if `NODE_ENV` is set to `production`. Set <Config>cds.server.index: true</Config> to activate explicitly (not recommended for production).
```
sjvans:
```suggestion
The default generic _index.html_ page is not served if `NODE_ENV` is set to `production`. Set <Config>cds.server.index: true</Config> to activate explicitly also in production-like test environments, for example for deployed PoCs. You should not do this in real production environments!
```
just a thought, i'm fine without as well :)

renejeglinsky: If this config is set, no profile whatsoever disables it again, right?
Was just thinking if that should be config inside a production profile to be explicitly enabled. Then one would see it in that used profile and not be surprised by general config being used.

renejeglinsky (approved):

node.js/cds-server.md
```diff
@@ -261,7 +261,7 @@ The built-in CORS middleware can be enabled explicitly with <Config>cds.server.c
 
 ### Toggle Generic Index Page
 
-The default generic _index.html_ page is not served if `NODE_ENV` is set to `production`. Set <Config>cds.server.index: true</Config> to restore the generic index page in production.
+The default generic _index.html_ page is not served if `NODE_ENV` is set to `production`. Set <Config>cds.server.index: true</Config> to activate explicitly also in production-like test environments, for example for deployed PoCs. You should not do this in real production environments!
```
renejeglinsky:
```suggestion
The default generic _index.html_ page is not served if `NODE_ENV` is set to `production`. Set <Config>cds.server.index: true</Config> to activate explicitly also in production-like test environments, for example for deployed PoCs. You must not do this in real production environments!
```

PR #2087, opened 2025-09-11
[cds^9.4] feat(node.js): translated error messages

johannes-vogel:
delivered with cds 9.4


sjvans (approved):

node.js/cds-i18n.md
```diff
@@ -674,11 +674,11 @@ Ensure you correctly understand how the config option `cds.i18n.folders` work be
 These are the current i18n entries for [`cds.i18n.messages`](#messages) used by the CAP runtime, which you can provide own translations for in your app-specific `_i18n/messages_<locale>.properties` files:
 
 ```properties
-MULTIPLE_ERRORS = Multiple errors occurred. Please see the details for more information.
-ASSERT_FORMAT = Value "{0}" is not in specified format "{1}"
-ASSERT_RANGE = Value {0} is not in specified range [{1}, {2}]
-ASSERT_ENUM = Value {0} is invalid according to enum declaration {{1}}
-ASSERT_NOT_NULL = Value is required
+MULTIPLE_ERRORS = Multiple errors occurred, see details below.
```
renejeglinsky, line 677: Can we replace "below"?

sjvans: @johannes-vogel

johannes-vogel: The snipper refers to the default messages which are delivered already for a while. We should only change the message in the docs if we change it in the code as well.

@renejeglinsky Do you want to revisit the implemented messages?

renejeglinsky: Terms like "above", "below", and so on are not helpful to persons with low vision that use accessibility tools. I think I haven't reviewed the latest changes (at least not its final version). So yes, I'd like to have a look at that.
But to to have the docs in sync with the code let's merge this PR.

PR #2096, opened 2025-09-17
fix: remove internals from outbox api

sjvans:


node.js/queue.md
```diff
@@ -128,8 +112,6 @@ Using the persistent queue, the to-be-emitted message is stored in a database ta
 The optional parameters are:
 
 - `maxAttempts` (default `20`): The number of unsuccessful emits until the message is considered unprocessable. The message will remain in the database table!
-- `parallel` (default `true`): Specifies if messages are sent in parallel (faster, but the order isn't guaranteed).
-- `chunkSize` (default `10`): The number of messages that are read from the database table in one go. Only applies for `parallel !== false`.
 - `storeLastError` (default `true`): Specifies whether error information of the last failed emit is stored in the tasks table.
```
David-Kunz, line 115: Should we also rm `storeLastError`? Should we remove the `legacyLocking` completely and switch it to `false`?

sjvans: imho, they can stay

PR #2098, opened 2025-09-17
Providing Services: changes after testing

renejeglinsky:


guides/providing-services.md
```diff
@@ -457,7 +457,7 @@ To illustrate the above:
 
 Fuzzy search is a fault-tolerant search feature of SAP HANA Cloud, which returns records even if the search term contains additional characters, is missing characters, or has typographical errors.
 
-You can configure the fuzziness in the range [0.0, 1.0]. The value 1.0 enforces exact search.
+You can configure the fuzziness in the range [0.0, 1.0].
```
renejeglinsky, line 460: Without the sentence it feels odd. Would sth like this help?

```suggestion
You can configure the fuzziness in the range [0.0, 1.0]. The value `1.0` enforces the least fuzzy search, if not overwritten by annotations on elements.
```

PR #2112, opened 2025-09-25
lint: move rule docs and samples closer together

chgeo:
The current split into `rules/` and `examples/` folders is a bit hard to grasp.  Move the sample and docs for a rule in one folder, eliminating the `examples` folder.

URLs change slightly:
- `cds-lint/rules/foobar` (before)
- `cds-lint/rules/foobar/` (now)


daogrady (approved):
👍 have been thinking the same. Thanks!

PR #2131, opened 2025-10-07
Describe usage of `concat`

eugene-andreev:


java/working-with-cql/query-api.md
```diff
@@ -1420,6 +1420,17 @@ Select.from(BOOKS).columns(
         .when(b.stock().gt(100)).then("high")
         .orElse("medium").as("stockLevel").type(CdsBaseType.STRING));
 ```
+#### String Expressions
```
agoerler, line 1427: Can we actually think of any other string expression but `concat`?  

Looking at our builder API it's a scalar function. Hence we could put the documentation there. 

When parsing CQN it's a String expression. Also in CQL it's a string expression. Hence have the docu here is also OK. May be add it under [Scalar Functions](https://pages.github.tools.sap/cap/docs/java/working-with-cql/query-api#scalar-functions) as well with a link to `String Expressions`?

agoerler (approved):

java/working-with-cql/query-api.md
```diff
@@ -1420,6 +1424,17 @@ Select.from(BOOKS).columns(
...
+
+* Concat
+
+    Function `concat` creates a string expression to concatenate a specified value to this value.
+
+    ```java
+    // SELECT from Author {name || ' - the Author' as full_name : String}
```
smahati:
```suggestion
    // SELECT from Author {name || ' - the Author' as author_name : String}
```
Edited comment to match the code.

smahati (approved):

PR #2134, opened 2025-10-07
fix profile

Akatuoro:
~~This somehow found it's way in... the build does not work with `with-mtx-sidecar`, the correct profile here is `mtx-sidecar` as also described in the node section.~~

Updating the sidecar package.json to what is currently generated (incl java profile).

### DIFF
--- a/guides/multitenancy/index.md
+++ b/guides/multitenancy/index.md
@@ -175,12 +175,18 @@ cds add multitenancy
      "devDependencies": {
        "@cap-js/sqlite": "^2"
      },
+     "engines": {
+       "node": ">=20"
+     },
      "scripts": {
        "start": "cds-serve",
        "build": "cds build ../.. --for mtx-sidecar --production && npm ci --prefix gen"
      },
      "cds": {
-       "profile": "mtx-sidecar"
+       "profiles": [
+         "mtx-sidecar",
+         "java"
+       ]
      }
    }
    ```


renejeglinsky:
Thanks @Akatuoro ,
what is actually added by `cds add multitenancy`? The details box described what this command adds and if that is also wrong we need to check this with our colleagues.

Akatuoro:
@renejeglinsky , good point, but it's just the documentation. `cds add multitenancy` adds the correct profiles:
```sh
$ cds init bookshop-mt --add java,multitenancy
...
$ cd bookshop-mt
$ cat mtx/sidecar/package.json 
{
  "name": "bookshop-mt-mtx",
  "dependencies": {
    "@cap-js/hana": "^2",
    "@sap/cds": "^9",
    "@sap/cds-mtxs": "^3",
    "@sap/xssec": "^4",
    "express": "^4"
  },
  "devDependencies": {
    "@cap-js/sqlite": "^2"
  },
  "engines": {
    "node": ">=20"
  },
  "scripts": {
    "start": "cds-serve",
    "build": "cds build ../.. --for mtx-sidecar --production && npm ci --prefix gen"
  },
  "cds": {
    "profiles": [
      "mtx-sidecar",
      "java"
    ]
  }
}

```

renejeglinsky:
@swaldmann can you please explain when that has changed? I couldn't find anything relevant in the changelog and I assume the docs were right at one point in time, or?

Akatuoro:
You're right, this is the app config, not the sidecar... Both me and some colleagues missed that :D

I've reverted the original change and instead updated the sidecar package.json with what is currently generated. That was missing the java profile. Future goal would be that we don't need to provide the java profile because it is automatically propagated, but currently there can be errors if the profile is not given.
There's also the engine which is now included in the generated package.json, so I just pasted the whole generated json to be up to date.

chgeo (approved):

guides/multitenancy/index.md
```diff
@@ -175,12 +175,18 @@ cds add multitenancy
...
      "scripts": {
        "start": "cds-serve",
        "build": "cds build ../.. --for mtx-sidecar --production && npm ci --prefix gen"
      },
      "cds": {
-       "profile": "mtx-sidecar"
+       "profiles": [
+         "mtx-sidecar",
+         "java"
+       ]
```
renejeglinsky, line 189:
```suggestion
       "profile": "mtx-sidecar"
```

renejeglinsky: In addition we add then a step outside the details to make Java users add the profile. Can't do it here, but will push that soon.

Akatuoro: @renejeglinsky , the java profile is added by `cds add multitenancy` if used in the context of a java project.
E.g.:
```
cds init bookshop --add java,multitenancy
```
or
```
cds init bookshop --add java
cd bookshop
cds add multitenancy
```

So
```json
       "profiles": [
         "mtx-sidecar",
         "java"
       ]
```
is the generated output.

Since cds-dk 8.0.1 at least. If the project was setup before that or a template was used which was created before version 8 and not updated (that is how I got here), then the user needs to add it.

PR #2163, opened 2025-10-22
fix: grammar

David-Kunz:
Feel free to merge after reviewing.

### DIFF
--- a/cds/cdl.md
+++ b/cds/cdl.md
@@ -979,7 +979,7 @@ Essentially, Compositions are the same as _[associations](#associations)_, just 
 :::
 
 ::: warning Limitations of Compositions of one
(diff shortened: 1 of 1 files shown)


cds/cdl.md
```diff
@@ -979,7 +979,7 @@ Essentially, Compositions are the same as _[associations](#associations)_, just
 :::
 
 ::: warning Limitations of Compositions of one
-Using of compositions of one for entities is discouraged. There is often no added value of using them as the information can be placed in the root entity. Compositions of one have limitations as follow:
+Using `Composition of one` for entities is discouraged. There is often no added value of using them as the information can be placed in the root entity. Compositions of one have limitations as follow:
```
renejeglinsky:
```suggestion
Using compositions of one for entities is discouraged. There is often no added value of using them as the information can be placed in the root entity. Compositions of one have limitations as follow:
```

PR #2167, opened 2025-10-23
Update authorization.md for request response clarity

sjvans:
Clarified behavior of request responses for non-existent entities and unauthorized access.


guides/security/authorization.md
```diff
@@ -517,7 +517,9 @@ Supported features are:
 * [Exists predicate](#exists-predicate) based on subselects.
 
 ::: info Avoid enumerable keys
-In case the filter condition is not met in an `UPDATE` or `DELETE` request, the runtime rejects the request (response code 403) even if the user is not even allowed to read the entity. To avoid to disclosure the existence of such entities to unauthorized users, make sure that the key is not efficiently enumerable.
+In case the filter condition is not met in an `UPDATE` or `DELETE` request, the runtime rejects the request with a `404` response if the entity doesn't exist but a `READ` request would succeed if it did.
+Otherwise, the runtime rejects with `403`, so that the response is consistent whether the entity doesn't exist or the user is not allowed to read it (in both cases a `READ` would result in a `404`).
```
renejeglinsky, line 521: What is "Otherwise" meaning here? Filter condition met? Or filter condition not met but a different request?

renejeglinsky:
```suggestion
For all other requests, if the filter condition is not met, the runtime rejects with `403`, so that the response is consistent whether the entity doesn't exist or the user is not allowed to read it (in both cases a `READ` would result in a `404`).
```

renejeglinsky: @sjvans Could you clarify if my assumption in the previous suggestion is correct?

sjvans: "Otherwise, i.e., the entity does exist but the user may not read it, the runtime rejects with ..."

example:
```
GET /Books(4711) → 404 → doesn't exist or i am not allowed to read it
PUT /Books(4711) → 404 → doesn't exist but i would be allowed to read it
PUT /Books(4711) → 403 → doesn't exist or i would not be allowed to read it
```
i.e., the `403` does not confirm that the entity exists

@BraunMatthias, plz confirm

PR #2182, opened 2025-10-30
Java: Draft queries on runtime views

MattSchur:
Clarify the configuration setting for draft queries on runtime views.

### DIFF
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -409,7 +409,7 @@ Remember to run draft specific queries through the [Draft Service](../fiori-draf
 
 
 ::: tip Draft queries on runtime views
-If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) entities and want to run draft specific queries on these views, set the [*cds.drafts.persistence*](../fiori-drafts#reading-drafts) configuration to `split`.
+If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) entities and want to run draft specific queries in [resolve mode](#rtview-resolve) on these views, set the [*cds.drafts.persistence*](../fiori-drafts#reading-drafts) configuration to `split`.
 :::
 
 ::: warning Avoid draft-enabling runtime views


java/working-with-cql/query-execution.md
```diff
@@ -409,7 +409,7 @@ Remember to run draft specific queries through the [Draft Service](../fiori-draf
 
 
 ::: tip Draft queries on runtime views
-If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) entities and want to run draft specific queries on these views, set the [*cds.drafts.persistence*](../fiori-drafts#reading-drafts) configuration to `split`.
+If you define runtime views on [draft-enabled](../fiori-drafts#reading-drafts) entities and want to run draft specific queries in [resolve mode](#rtview-resolve) on these views, set the [*cds.drafts.persistence*](../fiori-drafts#reading-drafts) configuration to `split`.
```
beckermarc, line 412: Somehow this tip seems to contradict the below warning.
The tip says: When using "resolve" mode make sure to set draft-mode "split"
The warning says: It only works with "cte" mode (which is not really worth a warning anymore, as this is the default)

Maybe we should merge these to boxes.

MattSchur: The tip is for defining runtime views on draft-enabled entities. The warning is for draft-enabling a runtime view (on a non-draft entity) which requires CTE mode and rather should be avoided, since the compiler would generate a draft table for the runtime view and a schema deployment is required on view changes (until we resolve this issue).

beckermarc: Defining views (of any sort) on draft-enabled entities is something I would say we don't support at all today. Is this a use case that is used by applications? I also noticed this in the wording, but thought that this was just a mistake.

The thing is: When defining a (runtime) view on a draft-enabled entity the resulting entity is also draft-enabled (as the @odata.draft.enabled annotation is propagated). So there is not really a difference between those two:
- "defining runtime views on draft-enabled entities" always implies
- "draft-enabling a runtime view" (unless @odata.draft.enabled is explicitly set to `false`, which I guess will even be less supported by our runtime).

renejeglinsky:
@MattSchur How to proceed with this PR?

PR #2190, opened 2025-11-06
node.js: Change 'code' to 'status' in req.reject() examples

etimr:


node.js/events.md
```diff
@@ -482,7 +482,7 @@ This is a convenience variant of the [`req.reject()`](#req-reject) method, with
 
 ```tsx
 function req.reject (
-  code?    : number,
+  status?  : number,
   message? : string,
```
hm23, line 486: Sorry for the confusion - is `message` correct here? In the example below, the second property is 'MISSING_INPUT', which looks like the `code` in the example above...

etimr: I think it can be both, in the example below:
```
req.reject ({ code: 400, message: 'MISSING_INPUT', target: 'title' })
req.reject (400, 'MISSING_INPUT', 'title') // same as above
```
According to the documentation:
- If `status` is given, it is used as the HTTP status code of the response. If `status` is omitted, and `code` is a number in the range of `300...600`, that number is used as the HTTP status code of the response.

- If `code` is given, and a string, it is used to look up a user-readable error message from the [i18n/messages](https://cap.cloud.sap/docs/node.js/cds-i18n) bundles. If `code` is omitted, the given `message` is used as the [i18n](https://cap.cloud.sap/docs/node.js/cds-i18n) key to look up the message, and if found, the original value of `message` is used as `code` in the response.

hm23: 👍 Thanks for the explanation. I don't like these ambiguities, but that's not your fault. 😄

hm23 (approved):

sjvans (approved):

renejeglinsky (approved):

node.js/events.md
etimr: me 2 :( not obvious and ambiguous

PR #2196, opened 2025-11-10
adjust the draft protection snippet to the implementation

rjayasinghe:


beckermarc (approved):

renejeglinsky:
We have started that and you need to add the "AI Review" label. Let me know what you think about the suggestions and how it works. We're continuously improving it (hopefully).

PR #2203, opened 2025-11-13
fix: rm `CREATE X.drafts`

sjvans:


node.js/fiori.md
```diff
@@ -77,13 +77,15 @@ As a result, a new entry to `MyEntity.drafts` is created.
 
 For logical reasons handlers for the `EDIT` event are registered on the active entity, i.e. `MyEntity` in the code above, not on the `MyEntity.drafts` entity.
 
+<!--
```
David-Kunz: We could leave the `NEW` part, no?

sjvans: but why? it's redundant to `NEW`

David-Kunz (approved):

PR #2216, opened 2025-11-19
@assert

StefanHenke:
Describes the basic feature set of the new `@assert` annotation. The full blown CXN documentation is pending and will be provided by Patrice later. Once available, we will link it here. That´s why the expressions in the examples are kept simple by intention.

### DIFF
--- a/cds/cdl.md
+++ b/cds/cdl.md
@@ -856,7 +856,51 @@ Result result = service.run(Select.from("UsingView"), params);
 [Learn more about how to expose views with parameters in **Services - Exposed Entities**.](#exposed-entities){ .learn-more}
 [Learn more about views with parameters for existing HANA artifacts in **Native SAP HANA Artifacts**.](../advanced/hana){ .learn-more}
 
+### Runtime Views { #runtimeviews }
 
+To add or update CDS views without redeploying the database schema, annotate them with [@cds.persistence.skip](../guides/databases#cds-persistence-skip). This advises the CDS compiler to skip generating database views for these CDS views. Instead, CAP resolves them *at runtime* on each request. 
+
+Runtime views must be simple [projections](#as-projection-on), not using *aggregations*, *join*, *union* or *subqueries* in the *from* clause, but may have a *where* condition if they are only used to read.
+
+In CAP Java, runtime views are enabled by default, in Node.js enable them via <Config>cds.features.runtime_views: true</Config>.
+
+[Learn more about runtime views in CAP Java.](../java/working-with-cql/query-execution#runtimeviews) {.learn-more}
+
+By default, runtime views are translated into _Common Table Expressions_ (CTEs) and sent with the query to the database.
+
+For example, given the following CDS model and query:
+
+```cds
+entity Books {
+  key ID     : UUID;
+      title  : String;
+      stock  : Integer;
+      author : Association to one Authors;
+}
+@cds.persistence.skip
+entity BooksWithLowStock as projection on Books {
+    ID, title, author.name as author
+} where stock < 10; // makes the view read only
+```
+```sql
+SELECT from BooksWithLowStock where author = 'Kafka'
+```
+
+The runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database.
+
+```sql
+WITH BOOKSWITHLOWSTOCK_CTE AS (
+    SELECT B.ID,
+           B.TITLE,
+           A.NAME AS "AUTHOR"
+      FROM BOOKS B
+      LEFT OUTER JOIN AUTHOR A ON B.AUTHOR_ID = A.ID
+     WHERE B.STOCK < 10
+)
+SELECT ID, TITLE, AUTHOR AS "author"
+  FROM BOOKSWITHLOWSTOCK_CTE
+ WHERE A.NAME = ?
+```
 
 ## Associations
 
--- a/java/working-with-cql/query-execution.md
+++ b/java/working-with-cql/query-execution.md
@@ -321,13 +321,12 @@ The delete operation is resolved to the underlying `Order` entity with ID *42* a
 
 ### Runtime Views { #runtimeviews }
 
-To add or update CDS views without redeploying the database schema, annotate them with [@cds.persistence.skip](../../guides/databases#cds-persistence-skip). This advises the CDS compiler to skip generating database views for these CDS views. Instead, CAP Java resolves them *at runtime* on each request.
 
-Runtime views must be simple [projections](../../cds/cdl#as-projection-on), not using *aggregations*, *join*, *union* or *subqueries* in the *from* clause, but may have a *where* condition if they are only used to read. On write, the restrictions for [write through views](#updatable-views) apply in the same way as for standard CDS views. However, if a runtime view cannot be resolved, a fallback to database views is not possible, and the statement fails with an error.
+CAP Java provides two modes for resolving [runtime views](../../cds/cdl#runtimeviews) during read operations: [cte](#rtview-cte) and [resolve](#rtview-resolve).
 
-CAP Java provides two modes for resolving runtime views during read operations: [cte](#rtview-cte) and [resolve](#rtview-resolve).
+On write, the restrictions for [write through views](#updatable-views) apply in the same way as for standard CDS views. However, if a runtime view cannot be resolved, a fallback to database views is not possible, and the statement fails with an error.
 
-::: details Changing the runtime view mode
+::: details Changing the runtime view read mode
 To globally set the runtime view mode, use the property `cds.sql.runtimeView.mode` with value `cte` (the default) or `resolve` in the *application.yml*. To set the mode for a specific runtime view, annotate it with `@cds.java.runtimeView.mode: cte|resolve`.
 
 To set the mode for a specific query, use a [hint](#hana-hints):
@@ -337,42 +336,9 @@ Select.from(BooksWithLowStock).hint("cds.sql.runtimeView.mode", "resolve");
 ```
 :::
 
-The next two sections introduce both modes using the following CDS model and query:
-
-```cds
-entity Books {
-  key ID     : UUID;
-      title  : String;
-      stock  : Integer;
-      author : Association to one Authors;
-}
-@cds.persistence.skip
-entity BooksWithLowStock as projection on Books {
-    ID, title, author.name as author
-} where stock < 10; // makes the view read only
-```
-```sql
-SELECT from BooksWithLowStock where author = 'Kafka'
-```
-
-
 #### Read in `cte` mode { #rtview-cte }
 
-This is the default mode since CAP Java `4.x`. The runtime translates the [view definition](#runtimeviews) into a _Common Table Expression_ (CTE) and sends it with the query to the database.
-
-```sql
-WITH BOOKSWITHLOWSTOCK_CTE AS (
-    SELECT B.ID,
-           B.TITLE,
-           A.NAME AS "AUTHOR"
-      FROM BOOKS B
-      LEFT OUTER JOIN AUTHOR A ON B.AUTHOR_ID = A.ID
-     WHERE B.STOCK < 10
-)
-SELECT ID, TITLE, AUTHOR AS "author"
-  FROM BOOKSWITHLOWSTOCK_CTE
- WHERE A.NAME = ?
-```
+In [cte mode](../../cds/cdl#runtimeviews), the runtime translates the view definition into a _Common Table Expression_ (CTE) and sends it with the query to the database. This is the default mode since CAP Java `4.x`.
 
 ::: tip CAP Java 3.10
 Enable *cte* mode with *cds.sql.runtimeView.mode: cte*
@@ -380,7 +346,9 @@ Enable *cte* mode with *cds.sql.runtimeView.mode: cte*
 
 #### Read in `resolve` mode { #rtview-resolve }
 
-The runtime _resolves_ the [view definition](#runtimeviews) to the underlying persistence entities and executes the query directly against the corresponding tables.
+In `resolve` mode, the runtime _resolves_ the view definition to the underlying persistence entities and executes the query directly against the corresponding tables.
+
+For example, the [view definition](../../cds/cdl#runtimeviews) is resolved into the following SQL statement:
 
 ```sql
 SELECT B.ID, B.TITLE, A.NAME AS "author"
--- a/node.js/cds-reflect.md
+++ b/node.js/cds-reflect.md
@@ -108,15 +108,22 @@ Example:
 ```js
 let m = cds.linked`
   namespace my.bookshop;
-  entity Books {...}
-  entity Authors {...}
+  entity Books {}
+  entity Authors {}
+  service CatalogService { 
+    entity Books as projection on my.bookshop.Books;
+    entity Authors as projection on my.bookshop.Authors;
+  }
 `
+
 // Function nature
 let { Books, Authors } = m.entities ('my.bookshop')
 
-
-// Object nature (uses the model's top-level namespace)
-let { Books, Authors } = m.entities
+// Object nature
+let { 
+  'my.bookshop.Books': Books, 
+  'my.bookshop.Authors': Authors 
+} = m.entities
 
 // Array nature
 for (let each of m.entities) console.log (each.name)
--- a/node.js/fiori.md
+++ b/node.js/fiori.md
@@ -10,14 +10,18 @@ See [Cookbook > Serving UIs > Draft Support](../advanced/fiori#draft-support) fo
 [[toc]]
 
 
+
 <!--
 ## Serving `$metadata` Requests
-
-
-
-## Serving `$batch` Requests
-
 -->
+
+
+
+<!--
+## Serving `$batch` Requests
+-->
+
+
 
 ## Draft Entities {#draft-support}
 
@@ -135,6 +139,7 @@ srv.on('someAction', [ MyEntity, MyEntity.drafts ], /*...*/)
 ```
 
 
+
 ## Draft Locks
 
 To prevent inconsistency, the entities with draft are locked for modifications by other users. The lock is released when the draft is saved, canceled or a timeout is hit. The default timeout is 15 minutes. You can configure this timeout by the following application configuration property:
@@ -151,6 +156,7 @@ You can set the property to one of the following:
 :::tip Delete released draft locks
 If the `draft_lock_timeout` has been reached, every user can delete other users' drafts to create an own draft. There can't be two drafts at the same time on the same entity.
 :::
+
 
 
 ## Draft Timeouts
@@ -178,49 +184,49 @@ It can occur that inactive drafts are still in the database after the configured
 :::
 
 
-## Bypassing Drafts
-Creating or modifying active instances directly is possible without creating drafts. This comes in handy when technical services without a UI interact with each other.
 
-To enable this feature, set this feature flag in your configuration:
+## Bypassing Drafts {.deprecated}
 
-```json
-{
-  "cds": {
-    "fiori": {
-      "bypass_draft": true
-    }
-  }
-}
-```
+Use [Direct CRUD](#direct-crud) instead.
 
-You can then create active instances directly:
+Until the next major release (`cds10`), you can still activate the draft bypass without also allowing direct CRUD via <Config>cds.fiori.bypass_draft:true</Config>.
+
+
+
+## Direct CRUD <Beta />
+
+With <Config>cds.fiori.direct_crud:true</Config>, creating or modifying active instances directly is possible without creating drafts.
+This comes in handy when technical services without a UI interact with each other.
+
+That is, you can then create and modify active instances directly:
 
 ```http
 POST /Books
 
 {
-  "ID": 123,
-  "IsActiveEntity": true
+  "ID": 123
 }
 ```
 
-You can modify them directly:
-
 ```http
-PATCH /Books(ID=123,IsActiveEntity=true)
+PATCH /Books(ID=123)
 
 {
   "title": "How to be more active"
 }
 ```
 
-This feature is required to enable [SAP Fiori Elements Mass Edit](https://sapui5.hana.ondemand.com/sdk/#/topic/965ef5b2895641bc9b6cd44f1bd0eb4d.html), allowing users to change multiple objects with the
+For this, the default draft creation behavior by SAP Fiori Elements is redirected to a collection-bound action via annotation `@Common.DraftRoot.NewAction`.
+The thereby freed `POST` request to draft roots without specifying `IsActiveEntity` leads to the creation of an active instance (as it would without draft enablement).
+
+The feature is required to enable [SAP Fiori Elements Mass Edit](https://sapui5.hana.ondemand.com/sdk/#/topic/965ef5b2895641bc9b6cd44f1bd0eb4d.html), allowing users to change multiple objects with the
 same editable properties without creating drafts for each row.
 
 :::warning Additional entry point
 Note that this feature creates additional entry points to your application. Custom handlers are triggered with delta
 payloads rather than the complete business object.
 :::
+
 
 
 ## Programmatic APIs <Beta />
--- a/package-lock.json
+++ b/package-lock.json
@@ -189,7 +189,6 @@
       "integrity": "sha512-PTAFMJOpVtJweExEYYgdmSCC6n4V/R+ctDL3fRQy77ulZM/p+zMLIQC9c7HCQE1zqpauvVck3f2zYSejaUTtrw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@algolia/client-common": "5.38.0",
         "@algolia/requester-browser-xhr": "5.38.0",
@@ -358,7 +357,6 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "Apache-2.0",
-      "peer": true,
       "peerDependencies": {
         "@sap/cds": ">=9.0.0",
         "@sap/cds-dk": "^9",
@@ -561,8 +559,7 @@
       "resolved": "https://registry.npmjs.org/@cspell/dict-css/-/dict-css-4.0.18.tgz",
       "integrity": "sha512-EF77RqROHL+4LhMGW5NTeKqfUd/e4OOv6EDFQ/UQQiFyWuqkEKyEz0NDILxOFxWUEVdjT2GQ2cC7t12B6pESwg==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@cspell/dict-dart": {
       "version": "2.3.1",
@@ -702,16 +699,14 @@
       "resolved": "https://registry.npmjs.org/@cspell/dict-html/-/dict-html-4.0.13.tgz",
       "integrity": "sha512-vHzk2xfqQYPvoXtQtywa6ekIonPrUEwe2uftjry3UNRNl89TtzLJVSkiymKJ3WMb+W/DwKXKIb1tKzcIS8ccIg==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@cspell/dict-html-symbol-entities": {
       "version": "4.0.4",
       "resolved": "https://registry.npmjs.org/@cspell/dict-html-symbol-entities/-/dict-html-symbol-entities-4.0.4.tgz",
       "integrity": "sha512-afea+0rGPDeOV9gdO06UW183Qg6wRhWVkgCFwiO3bDupAoyXRuvupbb5nUyqSTsLXIKL8u8uXQlJ9pkz07oVXw==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@cspell/dict-java": {
       "version": "5.0.12",
@@ -909,8 +904,7 @@
       "resolved": "https://registry.npmjs.org/@cspell/dict-typescript/-/dict-typescript-3.2.3.tgz",
       "integrity": "sha512-zXh1wYsNljQZfWWdSPYwQhpwiuW0KPW1dSd8idjMRvSD0aSvWWHoWlrMsmZeRl4qM4QCEAjua8+cjflm41cQBg==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/@cspell/dict-vue": {
       "version": "3.0.5",
@@ -1568,12 +1562,11 @@
       }
     },
     "node_modules/@eslint/js": {
-      "version": "9.39.1",
-      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-9.39.1.tgz",
-      "integrity": "sha512-S26Stp4zCy88tH94QbBv3XCuzRQiZ9yXofEILmglYTh/Ug/a9/umqvgFtYBAo3Lp0nsI/5/qH1CCrbdK3AP1Tw==",
+      "version": "9.39.2",
+      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-9.39.2.tgz",
+      "integrity": "sha512-q1mjIoW1VX4IvSocvM/vbTiveKC4k9eLrajNEuSsmjymSDEbpGddtpfOoN7YGAqBK3NG+uqo8ia4PDTt8buCYA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       },
@@ -2355,7 +2348,6 @@
       "integrity": "sha512-8fZ4DpLrMHt54nK1lQL07HjHJQ4S2KRcmqwJyo+8a7vFFiH4t1LEYO4Kaa29n0FydAoj5HqOkujJisdUkLEBRg==",
       "dev": true,
       "license": "SEE LICENSE IN LICENSE",
-      "peer": true,
       "dependencies": {
         "@sap/cds-compiler": "^6.3",
         "@sap/cds-fiori": "^2",
@@ -2671,7 +2663,6 @@
       "integrity": "sha512-dVd04UKsfpINUnK0yBoYHDF3xu7xVH4BuDotC/xGuycx4CgbP48X/KF/586bcObxT0HENHXEU8Nqtu6NR+eKhw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@types/body-parser": "*",
         "@types/express-serve-static-core": "^4.17.33",
@@ -2839,7 +2830,6 @@
       "integrity": "sha512-N9lBGA9o9aqb1hVMc9hzySbhKibHmB+N3IpoShyV6HyQYRGIhlrO5rQgttypi+yEeKsKI4idxC8Jw6gXKD4THA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.49.0",
         "@typescript-eslint/types": "8.49.0",
@@ -3320,6 +3310,7 @@
       "integrity": "sha512-PYAthTa2m2VKxuvSD3DPC/Gy+U+sOA1LAuT8mkmRuvw+NACSaeXEQ+NHcVF7rONl6qcaxV3Uuemwawk+7+SJLw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "mime-types": "~2.1.34",
         "negotiator": "0.6.3"
@@ -3334,7 +3325,6 @@
       "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -3385,7 +3375,6 @@
       "integrity": "sha512-8VJKIzheeI9cjuVJhU1hYEVetOTe7LvA+CujAI7yqvYsPtZfVEvv1pg9AeFNtHBg/ZoSLGU5LPijhcY5l3Ea9g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@algolia/abtesting": "1.4.0",
         "@algolia/client-abtesting": "5.38.0",
@@ -3426,6 +3415,22 @@
         "url": "https://github.com/chalk/ansi-regex?sponsor=1"
       }
     },
+    "node_modules/ansi-styles": {
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
+      "integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "color-convert": "^2.0.1"
+      },
+      "engines": {
+        "node": ">=8"
+      },
+      "funding": {
+        "url": "https://github.com/chalk/ansi-styles?sponsor=1"
+      }
+    },
     "node_modules/argparse": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz",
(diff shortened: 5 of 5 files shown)


guides/providing-services.md
```diff
@@ -939,6 +939,42 @@ Cross-service checks are not supported. It is expected that the associated entit
 The `@assert.target` check constraint relies on database locks to ensure accurate results in concurrent scenarios. However, locking is a database-specific feature, and some databases don't permit to lock certain kinds of objects. On SAP HANA, for example, views with joins or unions can't be locked. Do not use `@assert.target` on such artifacts/entities.
 :::
 
+### `@assert` <Beta/>
+
+Annotate an element with `@assert` to define an expression for more complex validations that need to be fulfilled before data gets persisted. The expectation is that in case of a violated validation, the expression returns the error message to be sent to the client, or `null` if the validation passed. With the help of `case when` syntax, it is possible to run several validations one after another and fail early.
```
rjayasinghe: The code examples don't have the `null` case. IIRC this does not need to be contained in the expression in the code, right? If yes, we don't need to mention it here.

agoerler: > The expectation is

could be confused with the expectation of the assert 

> run several validations one after another

this a quite procedural way of looking at an expression

agoerler: We need to mention `null` as you could also reverse the logic:

```cds
@assert (case value > 0 then null else 'value must be positive' end)
```

guides/providing-services.md
```diff
@@ -939,6 +939,42 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
```
agoerler, line 945: please give an example without error function before we make things more complicated

guides/providing-services.md
```diff
@@ -939,6 +939,42 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+  bar : String;
+  @assert: (case
+    when boo > 100 then 'high'
+    when boo > 50  then 'medium'
+  end)
+  boo : Integer;
+  @assert: (case 
+    when length(car) > length(boo) then error('foo.car.invalid', (car, boo)) 
+  end)
+  car: String;
```
agoerler: maybe split this into three examples:

* most simple (one case)
* multiple cases
* error function

guides/providing-services.md
```diff
@@ -939,6 +939,42 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+}
+```
+
+Refer to [Expressions as Annotation Values](../cds/cdl.md#expressions-as-annotation-values) for detailed rules on expression syntax.
+
+::: info Expression Evaluation
+Expressions are evaluated against the after-image of the affected entity. The after-image is the state of the given entity *after* the given request is applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys. The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can be one or more statements being executed.
```
agoerler: I think we can avoid the term "after-image".
```suggestion
Expressions are evaluated *after* the request has been applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys. The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can be one or more statements being executed.
```

StefanHenke:
> Please introduce in smaller steps
> 
> * one case
> * multiple cases
> * error function
> 
> cosider using `? :`
> 
> Please add some examples for moderately more complex expressions.
> 
> Like, e.g.
> 
> * "if an order has items the order's total amount must be larger then 0"
> * "a book's author's name must not be null"

What exactly do you mean with `consider using `? :`?

We agreed with Johannes and Patrice yesterday that we will keep the expressions simple here and wait for the CXL docs. But I can check for a "moderately more complex expression".

rjayasinghe:
> > Please introduce in smaller steps
> > 
> > * one case
> > * multiple cases
> > * error function
> > 
> > cosider using `? :`
> > Please add some examples for moderately more complex expressions.
> > Like, e.g.
> > 
> > * "if an order has items the order's total amount must be larger then 0"
> > * "a book's author's name must not be null"
> 
> What exactly do you mean with `consider using `? :`?
> 
> We agreed with Johannes and Patrice yesterday that we will keep the expressions simple here and wait for the CXL docs. But I can check for a "moderately more complex expression".

I think @agoerler meant that we split up this single block into 3 examples. After each example we explain the nature of the used expression.

StefanHenke:
> > > Please introduce in smaller steps
> > > 
> > > * one case
> > > * multiple cases
> > > * error function
> > > 
> > > cosider using `? :`
> > > Please add some examples for moderately more complex expressions.
> > > Like, e.g.
> > > 
> > > * "if an order has items the order's total amount must be larger then 0"
> > > * "a book's author's name must not be null"
> > 
> > 
> > What exactly do you mean with `consider using `? :`?
> > We agreed with Johannes and Patrice yesterday that we will keep the expressions simple here and wait for the CXL docs. But I can check for a "moderately more complex expression".
> 
> I think @agoerler meant that we split up this single block into 3 examples. After each example we explain the nature of the used expression.

Yes, I have split up the examples already. But was not sure if `?:` is something in addition

guides/providing-services.md
```diff
@@ -939,6 +939,60 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+    when length(boo) > length(far) then error('foo.boo.toolong', (boo, bar)) 
+  end)
+  boo : String;
+}
+```
+
+Refer to [Expressions as Annotation Values](../cds/cdl.md#expressions-as-annotation-values) for detailed rules on expression syntax.
+
+::: info Expression Evaluation
+Expressions are evaluated *after* the request has been applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys. The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can be one or more statements being executed.
```
patricebender:
```suggestion
Expressions are evaluated *after* the request has been applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys.
The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can lead to one or more statements being executed.
```

guides/providing-services.md
```diff
@@ -939,6 +939,60 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+entity Foo {
+  @assert: (case
+    when boo > 100 then 'high'
+    when boo > 50  then 'medium'
+  end)
+  boo : Integer;
+}
+```
+
+If the error message requires parameters, the `error(key, parameters)` function can be used to pass parameters for the message. Each parameter can be represented by an expression. In its simplest form, this is the actual value of an entity field. The evaluating runtime will take care of replacing the placeholders in the message with the provided parameters.
```
patricebender:
```suggestion
If the error message requires parameters, the `error(key, parameters)` function can be used to pass parameters for the message. Each parameter can be represented by an expression.
In its simplest form, this is the actual value of an entity field. The evaluating runtime will take care of replacing the placeholders in the message with the provided parameters.
```

patricebender: [The node runtime is about to add](https://github.com/cap-js/cds-dbs/pull/1421) an `error` function with 3 parameters, the third being the target(s) of the evaluation. Is there also the possibility to specify one or more target(s) with java?

patricebender (commented):
👍 As Adrian stated, it would be nice to have a little more (complex) real life scenarios. I would have probably taken the well known Bookshop for showcasing the different scenarios. But that may be my personal taste.

guides/providing-services.md
StefanHenke: Yes, targets can be specified with a third parameter. We kept it optional so that the function can also be called with two parameters

StefanHenke:
> 👍 As Adrian stated, it would be nice to have a little more (complex) real life scenarios. I would have probably taken the well known Bookshop for showcasing the different scenarios. But that may be my personal taste.

I have improved the examples to use Bookshop related entities. Regarding the ternary operator, we still have to put the `: else` as the compiler complains otherwise.

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
 The `@assert.target` check constraint relies on database locks to ensure accurate results in concurrent scenarios. However, locking is a database-specific feature, and some databases don't permit to lock certain kinds of objects. On SAP HANA, for example, views with joins or unions can't be locked. Do not use `@assert.target` on such artifacts/entities.
 :::
 
+### `@assert` <Beta/>
+
+Annotate an element with `@assert` to define an expression for more complex validations that need to be fulfilled before data gets persisted. If the validation should fail, the expression must return the error message to be sent to the client, or `null` if the validation passed.
```
agoerler: > fulfilled before data gets persisted

I don't like this wording as we check after the data is persisted

* "fulfilled after the data has been persisted" - this is confusing if you consider the commit as part of the insert/update processing; still I like this wording best
* "fulfilled before the transaction is committed" - this is confusing as we don't validate before-commit (but we might adapt the implementation, to do so).
* "fulfilled eventually" - this is too fuzzy and could make readers believe we only support eventual consistency

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+entity Orders : cuid {
+   Items    : Composition of many OrderItems
+                   on Items.parent = $self;
+}
+
+entity OrderItems : cuid {
+    parent    : Association to Orders;  
+    book      : Association to Books;
+    @assert: (case
+      when book.stock <= amount then 'Stock exceeded'
```
agoerler: > 'Stock exceeded'

I think we gave the recommendation that error messages should not contain element names and should be rather imperative like "order less books". See https://github.wdf.sap.corp/cds-java/home/issues/1164

But I am not sure wether this recommendation should also apply to `@ssert` messages.

@beckermarc - what do you think?

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+}
+```
+
+By using multiple `when` sections, multiple conditions can be included in a single annotation. Each condition returns its own error message to precisely describe the error.
+
+```cds
+entity OrderItems : cuid {
+
+    @assert: (case
+      when book.stock = 0 then 'Stock is zero'
```
agoerler:
```suggestion
      when amount <= 0 then 'Amount is not positive'
```

(as the `book.stock = 0` would also be caught by 'book.stock <= amount')

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
 The `@assert.target` check constraint relies on database locks to ensure accurate results in concurrent scenarios. However, locking is a database-specific feature, and some databases don't permit to lock certain kinds of objects. On SAP HANA, for example, views with joins or unions can't be locked. Do not use `@assert.target` on such artifacts/entities.
 :::
 
+### `@assert` <Beta/>
+
+Annotate an element with `@assert` to define an expression for more complex validations that need to be fulfilled before data gets persisted. If the validation should fail, the expression must return the error message to be sent to the client, or `null` if the validation passed.
+
+The returned error can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
+[Learn more about localized messages](./i18n){.learn-more}
```
agoerler: Maybe we could move this further down before we explain the `error` function. Then we can firstly focus on the condition, then on the error message, parameters, targets.

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+    @assert: (case
+      when book.stock = 0 then 'Stock is zero'
+      when book.stock <= amount then 'Stock exceeded'
+    end)
+    quantity  : Integer;
+}
+```
+
+With the help of the function `error(message, parameters, targets)`, it is possible to specify parameters as well as explicit target elements for the message. Each parameter can be represented by an expression.
```
agoerler:
```suggestion
#### The `error` function

With the help of the function `error(message, parameters, targets)`, it is possible to specify parameters as well as explicit target elements for the message. Each parameter can be represented by an expression.
```

agoerler: Maybe use imperative wording like "Use the `error` function to compose parameterized messages and/or to specify message targets"

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+  end)
+  dateOfBirth  : Date;
+  dateOfDeath  : Date;
+}
+```
+
+Refer to [Expressions as Annotation Values](../cds/cdl.md#expressions-as-annotation-values) for detailed rules on expression syntax.
+
+::: info Expression Evaluation
+Expressions are evaluated *after* the request has been applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys. The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can be one or more statements being executed.
```
agoerler:
```suggestion
Expressions are evaluated *after* the request has been applied to the underlying datastore. Affected are the entities of the request's payload. The runtime executes check-statements with the provided expressions and the primary key values for the given entities.
```

guides/providing-services.md
StefanHenke:
```suggestion
      when book.stock <= quantity then 'Stock exceeded'
```

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+    amount    : Decimal(9, 2);
+}
+```
+
+Alternatively, the same condition can be simplified by using the ternary operator:
+
+```cds
+entity OrderItems : cuid {
+
+    @assert: ((book.stock <= amount) ? 'Stock exceeded' : null)
```
StefanHenke:
```suggestion
    @assert: ((book.stock <= quantity) ? 'Stock exceeded' : null)
```

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+```
+
+By using multiple `when` sections, multiple conditions can be included in a single annotation. Each condition returns its own error message to precisely describe the error.
+
+```cds
+entity OrderItems : cuid {
+
+    @assert: (case
+      when book.stock = 0 then 'Stock is zero'
+      when book.stock <= amount then 'Stock exceeded'
```
StefanHenke:
```suggestion
      when book.stock <= quantity then 'Stock exceeded'
```

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+entity OrderItems : cuid {
+    parent    : Association to Orders;  
+    book      : Association to Books;
+    @assert: (case
+      when book.stock <= quantity then 'Stock exceeded'
+    end)
+    quantity  : Integer;
+    amount    : Decimal(9, 2);
+}
+```
```
Akatuoro, line 954: This is quite a big model which distracts from the 3 lines which matter. Either reduce it to only the necessary lines or use `// [!code focus]`

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+Alternatively, the same condition can be simplified by using the ternary operator:
+
+```cds
+entity OrderItems : cuid {
+
+    @assert: ((book.stock <= quantity) ? 'Stock exceeded' : null)
```
Akatuoro: do we need the parentheses here?

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+Refer to [Expressions as Annotation Values](../cds/cdl.md#expressions-as-annotation-values) for detailed rules on expression syntax.
+
+::: info Expression Evaluation
+Expressions are evaluated *after* the request has been applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys. The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can be one or more statements being executed.
+:::
+
+::: warning Limitations
+- Validations will only be enforced in `CREATE` and `UPDATE` statements that contain all key fields (including deep insert and deep update)
+- Only elements with simple types (like, `String`, `Integer`, `Boolean`) can be annotated with `@assert`. Elements typed with structured or arrayed types are not supported
```
Akatuoro: - Does this only mean that I cannot annotate a structured type or does it also mean that I cannot annotate an element inside a structured type?
- What about associations?

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+```
+
+Refer to [Expressions as Annotation Values](../cds/cdl.md#expressions-as-annotation-values) for detailed rules on expression syntax.
+
+::: info Expression Evaluation
+Expressions are evaluated *after* the request has been applied to the underlying datastore. The affected entities are the entities being part of the request and identified by their primary keys. The evaluating runtime constructs and executes statements with the annotation-provided expressions and the respective primary key values for the given entities. Depending on the structure and cardinality of the request data it can be one or more statements being executed.
+:::
+
+::: warning Limitations
+- Validations will only be enforced in `CREATE` and `UPDATE` statements that contain all key fields (including deep insert and deep update)
```
Akatuoro: I don't understand this section. Does this mean that validations will not be enforced in statements other than `CREATE` / `UPDATE`? Or is this about having to provide the full primary key for each entry?

Potential alternative wording:
```suggestion
- All primary key fields need to be contained in the CQN statement for validations to be enforced (including deep insert and deep update)
```

guides/providing-services.md
beckermarc: Maybe 

> to define arbitrary validation expressions that are run after the data has been written to the database. If the validation should fail, the expression must return a String containing the error message to be sent to the client. If at least one such validation fails the transaction is rolled back.

guides/providing-services.md
beckermarc: I guess as error messages are defined in the domain of the application, they can contain terms from the business domain. I think that is the case with "stock" here, which is a valid term in the business domain of orders and just happens to also be the element name. As the target of this error message is also the `amount` I don't feel like this creates a duplication on the UI. Maybe one could however detail it out a little to: `'Available stock of' || books.stock || 'exceeded'`

guides/providing-services.md
beckermarc: Maybe rather: "Enter a positive number". Although such a condition could be also covered with `@assert.range`

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
 :::
 
+### `@assert` <Beta/>
+
+Annotate an element with `@assert` to define an expression for more complex validations that need to be fulfilled before data gets persisted. If the validation should fail, the expression must return the error message to be sent to the client, or `null` if the validation passed.
+
+The returned error can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
+[Learn more about localized messages](./i18n){.learn-more}
+
+The following example ensures that the `quantity` of the ordered book is validated against the actual `stock`. If there is not enough available in the stock, a static error message is returned.
```
schwma:
```suggestion
The following example ensures that the `quantity` of the ordered book is validated against the actual `stock`. If the stock level is insufficient, a static error message is returned.
```

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+    @assert: (case
+      when book.stock = 0 then 'Stock is zero'
+      when book.stock <= quantity then 'Stock exceeded'
+    end)
+    quantity  : Integer;
+}
+```
+
+With the help of the function `error(message, parameters, targets)`, it is possible to specify parameters as well as explicit target elements for the message. Each parameter can be represented by an expression.
+In its simplest form, this is the actual value of an entity field. The evaluating runtime will take care of replacing the placeholders in the message with the provided parameters.
```
schwma:
```suggestion
In its simplest form, this is the actual value of an entity field. The evaluating runtime replaces placeholders in the message with the provided parameters.
```

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+entity OrderItems : cuid {
+    parent    : Association to Orders;  
+    book      : Association to Books;
+    @assert: (case
+      when book.stock <= quantity then 'Stock exceeded'
+    end)
+    quantity  : Integer;
+    amount    : Decimal(9, 2);
+}
```
agoerler: Maybe use manged compositions of many aspects:

```suggestion
entity Books : cuid {
   title : String;
   stock : Integer;
}

entity Orders : cuid {
   Items : Composition of many {
      key itemId   : GUID;
         book     : Association to Books;

         @assert: (case // [!code focus]
            when book.stock <= quantity then 'Stock exceeded' // [!code focus]
         end) // [!code focus]
         quantity : Integer;
         amount   : Decimal(9, 2);
   };
}
```

guides/providing-services.md
schwma: In the Node runtime we have also registered the `@assert` handler on `UPSERT` events. Is this behavior similar in the Java runtime? Perhaps `UPSERT` should be mentioned here as well?

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+Alternatively, the same condition can be simplified by using the ternary operator:
+
+```cds
+entity OrderItems : cuid {
+
+    @assert: ((book.stock <= quantity) ? 'Stock exceeded' : null)
+    quantity  : Integer;
+}
+```
```
sjvans: ternary style/ `else` is not supported by assert collector!

patricebender: but this is parsed as `case … when` statement - it should look identical in the CSN

guides/providing-services.md
rjayasinghe: yes, UPSERT is also supported by the Java runtime.

guides/providing-services.md
rjayasinghe: @agoerler Why not the initial expression with `book.stock = 0`?

guides/providing-services.md
rjayasinghe: I would say that elements inside a structured type should be OK. Structured types itself are not working at the moment.

Regarding associations: following associations in expressions works but is not supported as there are some open questions.

guides/providing-services.md
rjayasinghe: @patricebender do we need the parentheses here?

guides/providing-services.md
```diff
@@ -941,32 +941,27 @@ The `@assert.target` check constraint relies on database locks to ensure accurat
 
 ### `@assert` <Beta/>
 
-Annotate an element with `@assert` to define an expression for more complex validations that need to be fulfilled before data gets persisted. If the validation should fail, the expression must return the error message to be sent to the client, or `null` if the validation passed.
+Annotate an element with `@assert` to define arbitrary validation expressions that are run after the data has been written to the database. If the validation should fail, the expression must return a String containing the error message to be sent to the client. If at least one such validation fails the transaction is rolled back.
```
agoerler:
```suggestion
Annotate an element with `@assert` to define expressions that are validated after the data has been written to the database. If the validation should fail, the expression must return a String containing the error message to be sent to the client. If at least one such validation fails the transaction is rolled back.
```

guides/providing-services.md
```diff
@@ -992,9 +987,16 @@ entity OrderItems : cuid {
     quantity  : Integer;
 }
 ```
+#### The Error Message String
+
+As mentioned above error message returned by the CXL expression inside the annotation can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
```
agoerler:
```suggestion
As mentioned above, the error message returned by the CXL expression inside the annotation can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
```

guides/providing-services.md
sjvans: right, but with an `else`, which causes issues. @beckermarc can provide details

sjvans: however, the collector isn't described here and also not on by default, so i guess it's fine. we want to start out with "do it on service level" anyways, right?

guides/providing-services.md
```diff
@@ -939,6 +939,89 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+         @assert: (case // [!code focus]
+            when book.stock <= quantity then 'Stock exceeded' // [!code focus]
+         end) // [!code focus]
+         quantity : Integer;
+         amount   : Decimal(9, 2);
+   };
+}
+```
+
+Alternatively, the same condition can be simplified by using the ternary operator:
```
DanSchlachter:
```suggestion
Alternatively, an assert with a single `when` .. `then` block can be simplified by using the ternary operator:
```
--> since ternary operator doesn't work with multiple independent conditions.

guides/providing-services.md
```diff
@@ -939,6 +939,89 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+   title : String;
+   stock : Integer;
+}
+
+entity Orders : cuid {
+   Items : Composition of many {
+      key itemId   : GUID;
+         book     : Association to Books;
+
+         @assert: (case // [!code focus]
```
DanSchlachter: code focus only works on the last line on my localhost somehow, this line shows the // [!code focus] on the rendered code block on the vitepress page

guides/providing-services.md
StefanHenke: @Akatuoro: Tried with @rjayasinghe and it works without parenthesis.

guides/providing-services.md
StefanHenke: I think the collector in general has problems with explicit `else null`s. Even `case ... when ... then ... else null` would cause the same error. On the other hand side, the compiler does not allow to write the ternary style without `:else`. Are we confident to improve the collector that it can handle `else null` in the future?

guides/providing-services.md
```diff
@@ -941,31 +941,23 @@ The `@assert.target` check constraint relies on database locks to ensure accurat
...
+         end)
          quantity : Integer;
-         amount   : Decimal(9, 2);
+         ...
    };
 }
 ```
 
-Alternatively, the same condition can be simplified by using the ternary operator:
+Alternatively, an assert with a single `when` .. `then` block can be simplified by using the ternary operator:
```
agoerler: > a single `when` .. `then` block

I think multiple blocks work as well:

(year = 2000 ? : 'Year is 2000' : year = 3000 : 'Year is 3000' : null)

guides/providing-services.md
```diff
@@ -987,35 +979,16 @@ entity OrderItems : cuid {
...
-  end)
-  dateOfBirth  : Date;
-  dateOfDeath  : Date;
-}
-```
 
 Refer to [Expressions as Annotation Values](../cds/cdl.md#expressions-as-annotation-values) for detailed rules on expression syntax.
 
+::: info Error Messages
+As mentioned above, the error message returned by the CXL expression inside the annotation can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
```
agoerler:
```suggestion
The error message returned by the CXL expression inside the annotation can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
```

guides/providing-services.md
StefanHenke: Oh, I haven´t seen your comment in the meeting chat from last week. I would keep the original description and not mention this possibility to not increase complexity at this point.

guides/providing-services.md
StefanHenke: changed

guides/providing-services.md
agoerler: > following associations in expressions works but is not supported as there are some open questions

Like which ?

agoerler (commented):
I think we should mention that `@assert` should be used on service level.

guides/providing-services.md
```diff
@@ -939,6 +939,86 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+With the help of the function `error(message, parameters, targets)`, it is possible to specify parameters as well as explicit target elements for the message. Each parameter can be represented by an expression.
+In its simplest form, this is the actual value of an entity field. The evaluating runtime will take care of replacing the placeholders in the message with the provided parameters.
+
+In the following example, it is expected that the error message with key `error.author.date` is defined to have two parameters which are filled with the concrete values of the elements `dateOfBirth` and `dateOfDeath`. The third parameter ensures that both fields are set as targets and thus marked erroneous.
+
+```cds
+entity Authors : cuid, managed {
+   
+  @assert: (case 
+    when dateOfBirth > dateOfDeath then error('error.author.date', (dateOfBirth, dateOfDeath), (dateOfBirth, dateOfDeath))
```
agoerler: the `error` function will not be release 2025/12

guides/providing-services.md
agoerler: > following associations in expressions works but is not supported as there are some open questions.
resolved

guides/providing-services.md
StefanHenke: I think I have removed that already before. Should be out

guides/providing-services.md
```diff
@@ -939,6 +939,110 @@ Cross-service checks are not supported. It is expected that the associated entit
 The `@assert.target` check constraint relies on database locks to ensure accurate results in concurrent scenarios. However, locking is a database-specific feature, and some databases don't permit to lock certain kinds of objects. On SAP HANA, for example, views with joins or unions can't be locked. Do not use `@assert.target` on such artifacts/entities.
 :::
 
+### `@assert` <Beta/>
+
+Annotate an _element_ with `@assert` to define CXL expressions to be validated _after_ the data has been written to the database. If the validation should fail, the expression will return a String that indicates an error to runtime, or `null` if the validation passes.
```
StefanHenke:
```suggestion
Annotate an _element_ with `@assert` to define CXL expressions to be validated _after_ the data has been written to the database. If the validation should fail, the expression will return a String that indicates an error to the runtime, or `null` if the validation passes. 
```

guides/providing-services.md
```diff
@@ -939,6 +939,110 @@ Cross-service checks are not supported. It is expected that the associated entit
...
 
+### `@assert` <Beta/>
+
+Annotate an _element_ with `@assert` to define CXL expressions to be validated _after_ the data has been written to the database. If the validation should fail, the expression will return a String that indicates an error to runtime, or `null` if the validation passes. 
+
+```cds
+entity OrderItems : cuid {
+        
+  @assert: (case 
+    when quantity <= 0 then 'Quantity must be greater then zero'
```
StefanHenke:
```suggestion
    when quantity <= 0 then 'Quantity must be greater than zero' 
```

guides/providing-services.md
```diff
@@ -939,6 +939,110 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+  quantity : Integer;       
+}
+```
+
+Alternatively, the same condition can be simplified by using the [ternary conditional operator](../releases/archive/2023/march23#ternary-conditional-operator):
+
+```cds
+entity OrderItems : cuid {
+  
+  @assert: (quantity <= 0 ? 'Quantity must be greater then zero' : null)
```
StefanHenke:
```suggestion
  @assert: (quantity <= 0 ? 'Quantity must be greater than zero' : null)
```

guides/providing-services.md
```diff
@@ -939,6 +939,110 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+  quantity  : Integer;
+}
+```
+
+#### Error Messages and Message Targets
+
+If a validation fails the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts) the error is persisted. The exception or error targets the annotated element, which is then highlighted on the Fiori UI.
```
StefanHenke:
```suggestion
If a validation fails, the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts) the error is persisted. The exception or error targets the annotated element, which is then highlighted on the Fiori UI.
```

guides/providing-services.md
```diff
@@ -939,6 +939,109 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+If a validation fails, the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts) the error is persisted. The exception or error targets the annotated element, which is then highlighted on the Fiori UI.
+
+::: info Error Messages
+The error message returned by the CXL expression inside the annotation can be either a static message or a message key to support i18n. If a message key is used, the message is looked up in the message bundle of the service.
+[Learn more about localized messages](./i18n){.learn-more}
+:::
+
+::: info Multiple Targets
+If an assert should target multiple elements you need to annotate all targeted elements.
+:::
```
Akatuoro: Imho we should leave this out until we have a real solution. Currently only single targets are supported -> "multiple targets" is misleading.

guides/providing-services.md
```diff
@@ -939,6 +939,109 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+entity OrderItems : cuid {
+  
+  @assert: (quantity <= 0 ? 'Quantity must be greater than zero' : null)
+  quantity  : Integer;
+}
+```
+
+#### Error Messages and Message Targets
+
+If a validation fails, the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts) the error is persisted. The exception or error targets the annotated element, which is then highlighted on the Fiori UI.
```
Akatuoro: The sentence is about the target -> keep the other things simple
```suggestion
If a validation fails, the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts) the error is persisted. The error targets the annotated element, which is then highlighted on the Fiori UI.
```

guides/providing-services.md
```diff
@@ -939,6 +939,109 @@ Cross-service checks are not supported. It is expected that the associated entit
 The `@assert.target` check constraint relies on database locks to ensure accurate results in concurrent scenarios. However, locking is a database-specific feature, and some databases don't permit to lock certain kinds of objects. On SAP HANA, for example, views with joins or unions can't be locked. Do not use `@assert.target` on such artifacts/entities.
 :::
 
+### `@assert` <Beta/>
+
+Annotate an _element_ with `@assert` to define CXL expressions to be validated _after_ the data has been written to the database. If the validation should fail, the expression will return a String that indicates an error to the runtime, or `null` if the validation passes.
```
Akatuoro: If we write that it is checked after the data is already written, it is easy to misunderstand that the data is already committed then.
```suggestion
Annotate an _element_ with `@assert` to define CXL expressions to be validated _after_ the data has been written to the database but _before_ it is committed. If the validation should fail, the expression will return a String that indicates an error to the runtime, or `null` if the validation passes. 
```

guides/providing-services.md
```diff
@@ -939,6 +939,109 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+#### Complex Asserts
+
+An `@assert` annotation may use other elements from the same entity. This annotation checks that the delivery date of an order is after the order date:
+
+```cds
+entity Order : cuid {
+        
+  @assert: (deliveryDate < orderDate ? 'DELIVERY_BEFORE_ORDER' : null)
+  items : Integer;
```
Akatuoro: Having `items : Integer` here is confusing. Usually this assert would be applied to the deliveryDate.
```suggestion
  deliveryDate : Date;
```

guides/providing-services.md
```diff
@@ -939,6 +939,109 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+```
+
+You can also perform validations based on entities associated via a to-many association. Use an [exists predicate](../cds/cql#exists-predicate) in this case:
+
+```cds
+entity Order : cuid {
+        
+  @assert: (exists items[book.isNotReleased = true] 
+    ? 'Some ordered book is not yet released' : null)
+  items : Integer;
```
Akatuoro: `items : Integer` does not work here since it should be a to-many association or composition. Typically order items are a composition.

```suggestion
  items : Composition of many OrderItems on items.order = $self;
```

guides/providing-services.md
```diff
@@ -965,60 +965,86 @@ entity OrderItems : cuid {
...
+  }
+  
+  entity Orders : cuid {
+    items : Composition of many OrderItems on items.order = $self;
+  }
+  
+  entity OrderItems : cuid {
+    order: Association to Orders;
+    book     : Association to Books;
+    quantity : Integer;
```
Akatuoro: Consistent formatting

```suggestion
    order : Association to Orders;
    book : Association to Books;
    quantity : Integer;
```

guides/providing-services.md
```diff
@@ -965,60 +965,86 @@ entity OrderItems : cuid {
...
+  }
 }
```
Akatuoro, line 1004: If we add the service definition here, we can make each following examples shorter via `annotate OrderService.OrderItems ...`

```suggestion
}

service OrderService {
  entity Orders as projection on db.Orders;
  entity OrderItems as projection on db.OrderItems;
}
```

guides/providing-services.md
```diff
@@ -939,6 +939,135 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+
+Use multiple `when` clauses to check multiple conditions with a single `@assert` annotation. Each condition returns its own error message to precisely describe the error:
+
+```cds
+entity OrderItems : cuid {
+  
+  @assert: (case
+    when book.stock = 0 then 'Stock is zero'
+    when book.stock <= quantity then 'Stock exceeded'
+  end)
```
Akatuoro: This assert should also be on service level since it contains second element / association

agoerler (approved):

guides/providing-services.md
```diff
@@ -939,6 +939,126 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+  quantity @assert: (case // [!code highlight]
+    when book.stock <= quantity then 'Stock exceeded' // [!code highlight]
+  end); // [!code highlight]
+}
+```
+
+You can also perform validations based on entities associated via a to-many association. Use an [exists predicate](../cds/cql#exists-predicate) in this case:
+
+```cds
+annotate OrderService.OrderItems with {
```
DanSchlachter: this is the same code example as above, no exists predicate used

StefanHenke: ups, yeah. That was a copy-paste-error from a refactoring today morning

DanSchlachter (approved):

guides/providing-services.md
```diff
@@ -939,6 +939,125 @@ Cross-service checks are not supported. It is expected that the associated entit
 The `@assert.target` check constraint relies on database locks to ensure accurate results in concurrent scenarios. However, locking is a database-specific feature, and some databases don't permit to lock certain kinds of objects. On SAP HANA, for example, views with joins or unions can't be locked. Do not use `@assert.target` on such artifacts/entities.
 :::
 
+### `@assert` <Beta/>
+
+Annotate an _element_ with `@assert` to define CXL expressions to be validated _after_ the data has been written to the database but _before_ it is committed. If the validation should fail, the expression will return a String that indicates an error to the runtime, or `null` if the validation passes.
```
```suggestion
Annotate an _element_ with `@assert` to define the CXL expressions that are validated _after_ the data has been written to the database but _before_ it is committed. If the validation fails, the expression will return a `String` that indicates an error to the runtime. If the validation passes the expression return `null` 
```

guides/providing-services.md
```diff
@@ -939,6 +939,125 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+[Learn more about localized messages](./i18n){.learn-more}
+:::
+
+
+#### Complex Asserts
+
+::: warning Use complex asserts on service layer
+Like other annotations, `@assert` is propagated to projections. If you annotate an element with `@assert` and the condition uses other elements -  from the same or an associated entity - you must ensure that these elements are available in all projections to which the annotated element is propagated. Otherwise the CDS model won't compile.
+
+It is therefore recommended to use complex asserts on the highest projection, i.e. on the service layer.
```
```suggestion
It is therefore recommended to use complex asserts on the highest projection, that is on the service layer.
```

guides/providing-services.md
```diff
@@ -939,6 +939,125 @@ Cross-service checks are not supported. It is expected that the associated entit
...
+}
+```
+
+#### Background
+
+Expressions are evaluated *after* the request has been applied to the underlying datastore. Affected are the entities of the request's payload. The runtime executes check-statements with the provided expressions and the primary key values for the given entities.
+
+::: warning Limitations
+- All primary key fields need to be contained in the CQN statement for validations to be enforced (including deep insert and deep update)
+- Only elements with simple types (like, `String`, `Integer`, `Boolean`) can be annotated with `@assert`. Elements typed with structured or arrayed types are not supported
```
```suggestion
- All primary key fields need to be contained in the CQN statement for validations to be enforced (including deep insert and deep update).
- Only elements with simple types (like, `String`, `Integer`, `Boolean`) can be annotated with `@assert`. Elements typed with structured or arrayed types are not supported.
```

PR #2231, opened 2025-11-24
Editing "Localized Data" guide

swaldmann:


guides/localized-data.md
```diff
@@ -80,9 +80,6 @@ entity localized.Books as select from Books {*,
   coalesce (localized.descr, descr) as descr
 };
 ```
-::: warning Note:
-In contrast to former versions, with CDS compiler v2 we don't add such entities to CSN anymore, but only on generated SQL DDL output.
-:::
```
swaldmann, line 85: Not really relevant any more what compiler v2 did

guides/localized-data.md
```diff
@@ -124,7 +121,9 @@ entity OpenBookView as select from Books {*}
 Include the `localized` association:
 
 ```cds
-entity ClosedBookView as select from Books { ID, title, descr, localized };
+entity ClosedBookView as select from Books {
+  ID, title, descr, localized
+};
```
swaldmann, line 126: to avoid horizontal scrolling

guides/localized-data.md
chgeo: True

PR #2241, opened 2025-12-02
Node.js: remove Basic Authentication

sjvans:


node.js/authentication.md
```diff
@@ -254,48 +257,10 @@ If you want to restrict these additional logins, you need to overwrite the defau
...
-::: code-group
-```json [package.json]
-"cds": {
-  "requires": {
-    "auth": "basic"
-  }
-}
-```
+::: tip
+The pre-defined mock users can be deactivated by using kind `basic` instead of `mocked`.
```
renejeglinsky: Do we not give any guidance how to create users for the kind `basic`?

sjvans: here: https://github.com/capire/docs/pull/2241/files#diff-33b3dbb1a184af85f839396cbf964f4260920c9eb5df462c01c95585fc1ba73fR208

renejeglinsky: So we're reducing redundancies here and assume users know it from that example? Then maybe we could add this:


```suggestion
The pre-defined mock users can be deactivated by using kind `basic` instead of `mocked`. In that case configure users yourself, as described previously.
```

PR #2245, opened 2025-12-03
review: Transition flow

renejeglinsky:


guides/providing-services.md
```diff
@@ -1390,7 +1393,7 @@ For example, if the current state is `Open` and the target state is `Accepted`,
 This ensures consistent state transitions without custom logic.
 
 ::: tip Generic handlers are not executed for draft entities
-For example, calling `acceptTravel()` on a `Travels` entity that is currently being _edited_, i.e. is in _inactive_ state, has no effect.
+If you, for example, call `acceptTravel()` on a `Travels` entity that is currently being _edited_, which means it is in _inactive_ state, has no effect.
```
```suggestion
If you, for example, call `acceptTravel()` on a `Travels` entity that is currently being _edited_, which means it is in _inactive_ state, that call has no effect.
```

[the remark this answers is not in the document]
renejeglinsky:
```suggestion
For example, if you call `acceptTravel()` on a `Travels` entity that is currently being edited (in _inactive_ state), the call has no effect.
```

guides/providing-services.md
```diff
@@ -1278,7 +1278,7 @@ namespace sap.capire.travels;
 
 entity Travels : managed {
   // [...]
-  Status       : Association to TravelStatus default 'O';
+  Status       : Association to TravelStatus @readonly default 'O';
```
sjvans:
```suggestion
  Status       : Association to TravelStatus default 'O';
```
`@readonly` get's added automatically. now also gone in [xtravels](https://github.com/capire/xtravels/commit/8e129e4f3d96db762cff942c8b05e7f2405e8db8)

sjvans (approved):

renejeglinsky:
How should this be understood then?
image

sjvans:
> How should this be understood then? image

if you annotate the _element_ with `@flow.status` (or `@flow.status: true`), you should also annotate the element with `@readonly` → fully manual mode

if you annotate the _entity_ with `@flow.status: XYZ`, then the element `XYZ` is made read-only automatically

PR #2249, opened 2025-12-03
add brew command for CAP Console

timon-gaebelein:
Adds brew command & updates links to install cap console. Removes note about pre-requisites, since they are all included in v1.2.0 now.


johanneseschrig (approved):

tools/console.md
```diff
@@ -15,9 +15,14 @@ The CAP console is a native desktop app, available for Windows & macOS, designed
 
 ## Setup CAP Console
 
-Download the CAP console from [SAP Tools](https://tools.hana.ondemand.com/#cloud-capconsole) and install it.
```
renejeglinsky, line 18: Will CAP Console still be available to download from there?

timon-gaebelein: Yes, update cycle on there is way slower though, which means that it almost always has to download a patch on first start.

renejeglinsky: So the change is only about our recommendation? Then we could maybe keep the hint towards the tools page in a collapsed details section (for those who know it and want to use it)?

renejeglinsky (approved):

PR #2260, opened 2025-12-09
feat: `CXL` documentation

patricebender:
TBD

### DIFF
--- a/cds/cxl.md
+++ b/cds/cxl.md
@@ -331,6 +331,17 @@ E.g. in the expression `price * quantity`, the `*` operator is a binary operator
 that multiplies the two factors `price` and `quantity`.
 :::
 
+## function { #function }
+
+
+<div class="diagram" >
+<Badge class="badge-inline" type="tip" text="💡 clickable diagram" />
+<div class="diagram" v-html="functionDef"></div>
+</div>
+
+
+CAP supports a set of [portable functions](../guides/databases/cql-to-sql.md#portable-functions) that can be used in all expressions. Those functions are passed through to the underlying database, allowing you to leverage the same functions for different databases, which greatly enhances portability.
+
 ## ref (path expression) { #ref }
 
 A `ref` (short for reference) is used to refer to an element within the model.
@@ -796,18 +807,6 @@ ON
 
 The path expression `author[ age < 40 ].name`
 navigates along the `author` association of the `Books` entity only if the author's age is below 40.
-
-
-## function { #function }
-
-
-<div class="diagram" >
-<Badge class="badge-inline" type="tip" text="💡 clickable diagram" />
-<div class="diagram" v-html="functionDef"></div>
-</div>
-
-
-CAP supports a set of [portable functions](../guides/databases/cql-to-sql.md#portable-functions) that can be used in all expressions. Those functions are passed through to the underlying database, allowing you to leverage the same functions for different databases, which greatly enhances portability.
 
 <style>
 


cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
+import orderingTerm from '../assets/cxl/ordering-term.drawio.svg?raw'
+import overClause from '../assets/cxl/over-clause.drawio.svg?raw'
+import intro from '../assets/cxl/intro.drawio.svg?raw'
+</script>
+
+# Core Expression Language (CXL) { #expressions }
+The Core Expression Language (`CXL`) is a language to express calculations, conditions,
+and other expressions in the context of CDS models and queries.
+**`CXL` extends the SQL expression language**, so many syntax elements from SQL are also available in `CXL`.
```
agoerler:
```suggestion
**`CXL` is inspired by SQL's expression language**, so many syntax elements from SQL are also available in `CXL`.
```

or

```suggestion
**`CXL` is based on SQL's expression language**, so many syntax elements from SQL are also available in `CXL`.
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
+---
+# layout: cds-ref
+shorty: Expressions
+synopsis: >
+  Specification of the Core Expression Notation (CXN) used to capture expressions as plain JavaScript objects.
```
agoerler:
```suggestion
  Specification of the Core Expression Language (CXL) used to capture expressions in the CDS model.
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+
+`CXL` can be used in various places (TODO: Links):
+- In queries as part of the select list or where clause
+- In calculated elements
+- In annotations, where supported
+
+::: info 💡 expressions in CAP are materialized in the context of queries
+No matter where `CXL` is used, it always manifests in queries.
+For example, [a calculated element](./cdl/#calculated-elements) defined in an entity will be resolved
+to the respective calculation in the generated query when the entity is queried.
```
agoerler, line 35: You can use expressions in various other places
* translated to EDMX-expressions
* to define projections between types
* projections can be resolved at runtime (runtime views)
* expressions can be evaluated in memory

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+Just create the sample app as described above and start a repl session within the newly created app by running:
+
+```sh
+cds repl --run .
+```
+
+:::info 💡 All of the example expressions follow the same pattern:
+1. A **CAP Style `CXL`** is shown in the context of a query.
+2. The equivalent **SQL Style `CXL`** is shown.
+3. The resulting **SQL output** is shown.
```
agoerler:
```suggestion
2. The resulting **SQL output** is shown.
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+<div class="diagram" v-html="intro"></div>
+
+
+## expr { #expr }
+
+An expression can hold various elements, such as references, literals, function calls, operators, and more. A few examples, in the context of a select list:
+```cds
+select from Books {
+  42                     as answer,         // literal
+  title,                                    // ref
```
agoerler:
```suggestion
  title,                                    // reference ("ref")
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+  price * quantity       as totalPrice,     // binary operator
+  substring(title, 1, 3) as shortTitle,     // function call
+  author.name            as authorName,     // ref with path segment
+  chapters[number < 3]   as earlyChapters,  // infix filter
```
agoerler:
```suggestion
  chapters[number < 3]   as earlyChapters,  // ref with infix filter
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+  exists chapters        as hasChapters,    // exists
+  count(chapters)        as chapterCount,   // aggregate function
+}
+```
+
+### syntax <Badge class="badge-inline" type="tip" text="💡 clickable diagram" />
+
+<div class="diagram" v-html="expr"></div>
+
+
```
agoerler, line 118: please also explain the plain vanilla ref (i.e. one segment, no filter)

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+  <div v-html="pathSegment"></div>
+</div>
+
+### path expression in the select list
+
+A path expression can be used to navigate to any element of the associations target:
+
+:::code-group
+```js [CQL]
+> await cds.ql`SELECT from Books { author.name, genre.name }` // [!code focus]
```
agoerler:
```suggestion
> await cds.ql`SELECT from Books { author.name, genre.name }` // [!code highlight]
```

(I think the readability is much better with "code highlight")

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+[
+  { author_name: 'Emily Brontë', genre_name: 'Drama' },
+  { author_name: 'Charlotte Brontë', genre_name: 'Drama' },
+  { author_name: 'Edgar Allen Poe', genre_name: 'Mystery' },
+  { author_name: 'Edgar Allen Poe', genre_name: 'Romance' },
+  { author_name: 'Richard Carpenter', genre_name: 'Fantasy' }
```
agoerler: > `SELECT from Books { author.name, genre.name }`

This is a very fortunate example:

```cds
entity MyBooks as SELECT from Books { author.name, genre.name }
```

would result in an entity MyBooks with a duplicate element `name` -> 💣

To avoid the conflict, we could use

```sql
SELECT from Books { title, author.name }
```

here we don't get a conflict but the nasty issue that if used in a view definition's query specification we get the elements `title` and `name`. But if the very same query is executed by the CAP Node.js runtime the result type has the elements `title` and `author_name`.

Also in CAP Java the query 

```java
Select.from("Books").columns("title", "author.name")
```

returns, when executed a result with elements `title` and `name` - just as the CDS Compiler does.

We could fix this/avoid the confusion by using

```sql
SELECT from Books { title, author.name as author }
```

Please note that this is confusion is only caused by how the default display name of an "inline" is determined. This difference should be explained in the CAP Node.js docu. For expressions, the difference is actually not relevant.

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+  sap_capire_bookshop_Books as B
+  left JOIN sap_capire_bookshop_Authors as author ON author.ID = B.author_ID
+  left JOIN sap_capire_bookshop_Genres as genre ON genre.ID = B.genre_ID
+```
+:::
+
+In this example, we select the names of the authors and genres of books.
+Both `author` and `genre` are associations on the `Books` entity.
+
+::: info 💡 Associations are **forward declared joins**
```
agoerler: >forward declared

I have no formal background of computer science. But my understanding of "forward declared" is roughly that you declare something (e.g. an identifier) so that you can use it (in your code or model) but define the details only later. In this sense I don't think that we are using the term "forward declared" correctly as we don't state that an association is a "join" and define the details (i.e. the on condition) later.

Instead, I think, an association defines a navigation path with specified foreign key-relationship. We can use the knowledge on the fk-relationship in different ways:
* in the on condition of a  join
* in the condition that correlates a subquery to a main query
* to select related entities by fk values with an additional query

We use the term "forward declared joins" often. But also, in my experience, many wonder what the meaning might be.

See: https://en.wikipedia.org/wiki/Forward_declaration

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+An association can be used just as a table alias.
```
agoerler: > An association can be used just as a table alias.

No. An association relates an entity to a particular set of instances of the associated entity type. E.g. a person could have a home address and a work address, which have to be distinguished.

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+They are declared **before** they are used (e.g. in an entity definition)
```
agoerler: > They are declared **before** they are used (e.g. in an entity definition)

Same is true for the non-association elements.

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+Once an association is traversed in a query, the respective join is added automatically.
```
agoerler: > the respective join

or a condition that correlates a subquery to a main query ...

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+:::
```
agoerler: In essence, association declare a navigation path  with specified FK relationship, which can be used in various ways.

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+
```
agoerler, line 437: maybe add more examples where path expression can be used (where, order by, ...)

I think we should also refs from the perspective of set theory.

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+This is especially useful for to-many relations.
+
+In the example a path expression combined with an [infix-filter](#infix-filter),
+allows to select all authors that have written at least one book in the `Fantasy` genre.
+
+:::code-group
+```js [CQL]
+> await cds.ql`
+  SELECT from ${Authors} as FantasyAuthors { name }
+  where exists books.genre[name = 'Fantasy']` // [!code focus]
```
agoerler:
```suggestion
  where exists books.genre[name = 'Fantasy']` // [!code highlight]
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+path-expressions can also be used after the `exists` predicate to check for the existence.
+This is especially useful for to-many relations.
+
+In the example a path expression combined with an [infix-filter](#infix-filter),
+allows to select all authors that have written at least one book in the `Fantasy` genre.
+
+:::code-group
+```js [CQL]
+> await cds.ql`
+  SELECT from ${Authors} as FantasyAuthors { name }
```
agoerler:
```suggestion
  SELECT from ${Authors} { name }
```

The alias is not used and does not add any value. Please also adjust the SQL below.

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+???
+:::
+
+### conclusion
+
+A `ref` can be used to reference an element.
+It is possible to navigate along [path segments](#path-segment) to reference elements within the model.
+This is not limited to an entities own elements, but can also be used to navigate associations to elements of related entities.
+
+This is only the tip of the iceberg.
```
agoerler: > tip of the iceberg

sound a bit like danger looming below the surface

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+
+::: info 💡 TODO
+???
+:::
+
+### conclusion
+
+A `ref` can be used to reference an element.
+It is possible to navigate along [path segments](#path-segment) to reference elements within the model.
+This is not limited to an entities own elements, but can also be used to navigate associations to elements of related entities.
```
agoerler: Much more: a "ref" references a set of data. E.g. the value of an element. Or the value of an element of an associated entity. Or an entity. Or a set of entities. Or a set of associated entities...

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+and the author is younger than 40 years.
+
+:::code-group {4}  <!-- hier Zeilennummer(n) anpassen -->
+```js [CAP Style]
+> await cds.ql`
+  SELECT from ${Books} { title }
+  where startswith(
+    author[ years_between(dateOfBirth, dateOfDeath) < 40 ].name,
+    'Emily'
+  )`
```
agoerler: I think this example is rather complicated and not so typical. It does the following:

A book has a set of one associated author. You filter this set with the condition `years_between(dateOfBirth, dateOfDeath) < 40`, which might yield in an empty set. From this set with cardinality 0,1 we access the element `name`, which might be `null`. We use this element in the `startswith` function, which evaluates to `null` if one argument is `null`. Hence we effectively obtain a condition that is roughly "where the author's name starts with `Emily` and the author is younger than 40 years". 

I think much clearer would be just to use a filtered expand on the select list:

```sql
Select from Authors { name, books[year > 2000] { title } }
```

expand only books that have been written after the year 2000.

If you really want o express "books where the author's name starts with `Emily`
and the author is younger than 40 years" I'd use this query:

```sql
Select from Books {title} 
       where exists author[years_between(dateOfBirth, dateOfDeath) < 40 and 
                           startswith(name, 'Emily')]
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+        ) + (
+          CASE
+            WHEN (CAST(STRFTIME('%Y%m', author.dateOfDeath) AS INTEGER) < CAST(STRFTIME('%Y%m', author.dateOfBirth) AS INTEGER)) THEN
+              (CAST(STRFTIME('%d%H%M%S%f0000', author.dateOfDeath) AS INTEGER) > CAST(STRFTIME('%d%H%M%S%f0000', author.dateOfBirth) AS INTEGER))
+            ELSE
+              (CAST(STRFTIME('%d%H%M%S%f0000', author.dateOfDeath) AS INTEGER) < CAST(STRFTIME('%d%H%M%S%f0000', author.dateOfBirth) AS INTEGER)) * -1
+          END
+        )
+      ) / 12
+    ) < ?
```
agoerler: I think this is way to complicated to illustrate path expressions. Why not use something like

`address[kind = 'home'].street`

this would result in SQL, which readers can actually understand

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+The join for this path expression is generated as usual and enhanced with the infix filter condition `years_between(dateOfBirth, dateOfDeath) < 40`.
+
+
+::: info 💡 Standard functions
+the `years_between` and `startswith` functions are in the [set of CAPs standard functions](../guides/databases.md#standard-database-functions) and are translated to the respective SQL to get the desired result.
+:::
+
+
+### use an infix-filter to make an association more specific
+
```
agoerler, line 810: > make an association more specific

i.e. to navigate to a subset of the associated entities

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+### binary operator { #binary-operator }
+
+<div class="diagram">
+<div v-html="binaryOperator"></div>
+</div>
+A binary operator is an operator that operates on two operands.
+
+E.g. in the expression `price * quantity`, the `*` operator is a binary operator
+that operates on the two operands `price` and `quantity`.
```
agoerler:
```suggestion
E.g. in the expression `price * quantity`, the `*` operator is a binary operator
that multiplies the two factors `price` and `quantity`.
```

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+
+## literal value { #literal-value }
+
+<div class="diagram" >
+<div v-html="literalValue"></div>
+</div>
+
+TODO
```
agoerler: don't forget to specify the literal formats for

* Date
* Time
* Timestamp
* Binary

cds/cxl.md
```diff
@@ -0,0 +1,431 @@
...
+TODO: some text
+
+## over-clause <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> { #over-clause }
+
+<div class="diagram" v-html="overClause"></div>
+
+## type-ref <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> { #type-ref }
+
+TODO
+
```
agoerler: I think this is too specialized for the CXL docu

cds/cxl.md
Akatuoro: Query does not mean that it is a database query. There can be multiple vehicles for an expression. Conceptually though, it can be understood as part of a query - whether it is either sent to the database, converted to edmx (and then sent to the backend again), or evaluated in memory.
This is an important point though. Expressions are not only meant for database queries.

Keeping this open -> happy for suggestions on how to formulate this.

cds/cxl.md
Akatuoro: [ChatGPT agrees](https://chatgpt.com/share/693bd2cf-bae4-8006-bf95-b43fa488df81). Not that that is always a good indicator...

Reading the wiki page, I'd also agree that the term does not fit for associations, particularly compared with its typical usage. In associations we fully define the join, even if we may later add to it through infix filters.

What do you think about "predefined joins" or "predefined joins for path navigation"?

I think we need to discuss this further

Akatuoro: [ChatGPT agrees](https://chatgpt.com/share/693bd2cf-bae4-8006-bf95-b43fa488df81). Not that that is always a good indicator...

Reading the wiki page, I'd also agree that the term does not fit for associations, particularly compared with its typical usage. In associations we fully define the join, even if we may later add to it through infix filters.

What do you think about "predefined joins" or "predefined joins for path navigation"?

I think we need to discuss this further

cds/cxl.md
Akatuoro: In my opinion "code focus" provides better readability... a stronger focus on the line that matters

cds/cxl.md
```diff
@@ -0,0 +1,528 @@
...
+  <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> 
+  <div v-html="ref"></div>
+</div>
+
+
+### simple element reference
+
+
+:::code-group
+```js [CQL] {3,4,5}
```
renejeglinsky: Why only 3,4,5? Have 6 and 7 been added later and just missed to be included here?

Image


```suggestion
```js [CQL] {3-7}
```

cds/cxl.md
```diff
@@ -0,0 +1,528 @@
...
+<Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> 
+<div v-html="infixFilter"></div>
+</div>
+
+### enhancing path expression with filter conditions
+
+In this case we want to select all books where the author's name starts with `Emily`
+and the author is younger than 40 years.
+
+:::code-group
```
renejeglinsky, line 766: This code group breaks with the explained style in the samples chapter:

Image

cds/cxl.md
Akatuoro: If it is part of the diagrams, it would be good to also have an example for it

cds/cxl.md
patricebender: was a copy paste error or something like that :)

cds/assets/cxl.excalidraw.svg
swaldmann, line 1: Unused? Can be removed?

cds/assets/cxl/intro.drawio.svg
swaldmann, line 1: Can we use a rendered monospace font instead of \`exists\` and \`null\`?

cds/assets/cxl.excalidraw.svg
Akatuoro: Yes, that was the initial sketch

cds/assets/cxl/intro.drawio.svg
Akatuoro: Yes :)

assets/cxl/sets-intersection.drawio.svg
swaldmann, line 1: Took this as an example to try it with some (pseudo) syntax coloring. For comparison one graphic with and one without:

Image

I kinda like it and could do this tomorrow for the other graphics too if you want? @patricebender @Akatuoro

patricebender: I like it :) We figured that it is quite hard to get the indentation right within the text boxes though. Daniel suggested to use VS Code screenshots. If we can get it right with real text I'd prefer that over screenshots though.

patricebender:
> There's one example where it's not the text but the box that is clickable. Is that on purpose? image

actually, I want to have clickable boxes. Will check!

assets/cxl/sets-intersection.drawio.svg
swaldmann: Yes, draw.io seems to aggressively remove all spaces.

I was able to set the left padding/margin for each indented line, which also persists in the rendered view:

Image

Using text also has the advantage that draw.io nowadays supports distinguishing between bright/dark mode, so dark mode also looks quite nice:

Image

You can fine-tune the dark mode colors but I was pretty happy with draw.io's auto-generated ones.

Just pushed the new graphics in [f82b0d7](https://github.com/capire/docs/pull/2260/commits/f82b0d79f1f84224e11deeaabd818f4bf439c9db)

cds/cxl.md
```diff
@@ -0,0 +1,953 @@
...
+```
+:::
+
+In this example, the ordering term sorts books by price in descending order and places rows with `null` prices at the end.
+
+
+## type-ref <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> { #type-ref }
+
+::: info 💡 learn more about type references [here](./cdl.md#type-references)
+:::
```
swaldmann: Please use the dedicated `{.learn-more}` link instead for for such cases.

→ **Learn more** on \<internal-capire\>/resources/authoring#learn-more-link 💡

We usually advise not to link to "here", but to give more of a hint where you're linking to. You can also omit the `.md` in the URL:

```suggestion
[Learn more about type references in CDL](./cdl#type-references){ .learn-more }
```

cds/cxl.md
```diff
@@ -0,0 +1,953 @@
...
+and other expressions in the context of CDS models and queries.
+**`CXL` is based on the SQL expression language**, so many syntax elements from SQL are also available in `CXL`.
+
+`CXL` can be used in various places:
+- In [CQL](./cql#path-expressions) (select list, where clause, …)
+- In [CDL](./cdl)
+  + In [calculated elements](./cdl/#calculated-elements)
+  + In [annotations](./cdl.md#expressions-as-annotation-values)
+
+::: info 💡 expressions in CAP are materialized in the context of queries
```
swaldmann:
```suggestion
::: info 💡 Expressions in CAP are materialized in the context of queries
```

assets/cxl/sets.drawio.svg
swaldmann, line 1: @Akatuoro this is a merged/sketch version of the other graphics? I see it's not used in the `.md`. This can be removed now in favor of the syntax-colored individual ones?

assets/cxl/sets-leftjoin.drawio.svg
swaldmann, line 1: Whenever I see Venn diagrams explaining JOINs I'm reminded of this article/rant, where the author makes the case that set theory is a false analogy as JOINs are rather a cross product filtered by a predicate with an optional `UNION`:
https://blog.jooq.org/say-no-to-venn-diagrams-when-explaining-joins/

Food for thought 🤔  The JOIN diagrams shown there could also come in helpful in some Capire guides.

assets/cxl/sets-intersection.drawio.svg
patricebender: great thanks for your help 🥇

Akatuoro: This looks great, thanks!

assets/cxl/sets-leftjoin.drawio.svg
Akatuoro: Thanks for the link! This topic also came up when writing the theory section, because the inner join is indeed $\\{ (a, b) \in \text{Authors} \times \text{Books} | a.\text{id} = b.\text{author\\_id} \\}$ -> a cross product with a filter.

Intersections (and thus Venn diagrams) are possible though if the sets are made union-compatible:

The key point of the blog entry in regards to why this can not be shown as a Venn diagram is that the sets (Authors and Books) are of a different type and are not *union-compatible*. So $\text{Authors} \cap \text{Books}$ is indeed not correct.
Rather, we implicitely look at the cross product
$$X = \\{ \space (a_1, \dots, a_n, b_1, \dots, b_m) \in \\{ \text{Authors} \cup \\{(\omega, \dots, \omega)\\} \\} \times \\{ \text{Books} \cup \\{(\omega, \dots, \omega)\\} \\} - \\{(\omega, \dots, \omega)\\} \space \\}$$
with each circle representing:
$$\text{Authors}' = \\{ (a_1, \dots, a_n, b_1, \dots, b_m) \in X | a_\text{id} \ne \omega \\}$$
$$\text{Books}' = \\{ (a_1, \dots, a_n, b_1, \dots, b_m) \in X | b_\text{id} \ne \omega \\}$$

Getting the actual authors back is now only an exercise of projection:
$$\\{ (a_1, \dots, a_n) \\} = \\{ \pi_{a_1, \dots, a_n}(\text{Authors}') \\} = \text{Authors}$$
-> `Select distinct a.* from Authors a full join Books b on 1 = 1`

Since the sets are independent of the projection (what we select), any join operations simply filter on the cross product, which imho can be nicely displayed using venn diagrams. So I do disagree with the author of the blog even though they have a point. I also think the join diagrams could come in helpful.

One of the key points in cql / cxl is that path navigation allows us to work on sets and only consider the select list (projection in relational algebra) at the end. When actually writing relational algebra, the tuple needs to be defined explicitely. CQL and SQL also handle null values nicely. When writing relational algebra, null values need to be explicitely included.

cds/cxl.md
```diff
@@ -98,140 +118,628 @@ select from Books {
...
+ORDER BY
+  author.dateOfBirth ASC
 ```
 :::
 
-::: info 💡 TODO
-???
+In this example, we select all books and order them by the date of birth of their authors.
+The table alias for the `author` association is used in the order by clause of the SQL query.
+
```
agoerler, line 488: please add a hint that you should any path expression used in order by should also be on the select list

assets/cxl/sets.drawio.svg
Akatuoro: Yes, we initially drew the merged version and later separated them -> removed

cds/cxl.md
```diff
@@ -98,140 +118,628 @@ select from Books {
...
+### after `exists` predicate
+
+path expressions can also be used after the `exists` predicate to check for the existence.
```
agoerler:
```suggestion
Path expressions can also be used in the `exists` predicate to check wether the set that is referenced by the path is not empty.
```

cds/cxl.md
```diff
@@ -98,140 +118,628 @@ select from Books {
...
+This is especially useful for to-many relations.
+
+E.g., to select all authors that have written **at least**  one book:
```
agoerler:
```suggestion
E.g., to select all authors that have written **at least** one book:
```

cds/cxl.md
```diff
@@ -98,140 +118,628 @@ select from Books {
...
+
+::: info 💡 Learn more about the `exists` predicate [here](./cql.md#exists-predicate)
 :::
 
-### conclusion
+### theory of path expressions
+
+
+
+Every entity defines a set of all possible instances:
```
agoerler:
```suggestion
Every entity defines a set of all instances:
```

cds/cxl.md
```diff
@@ -98,140 +118,628 @@ select from Books {
...
+$${ b \in \text{Books} }$$
+
+A simple select query on Books returns the complete set → all books.
```
agoerler:
```suggestion
A simple select from Books returns the complete set → all books.
```

assets/cxl/infix-filter-full.drawio.svg
agoerler, line 1: some (but not all) DBs allow offset without limit

cds/assets/cxl/literal-value.drawio.svg
agoerler, line 1: true/false -> Boolean-literal ?

What about timestamp ?

cds/assets/cxl/ref.drawio.svg
agoerler, line 1: why no infix filter for structured elements?

assets/cxl/sets-expand.drawio.svg
agoerler, line 1: The diagram is not correct as there is no intersection between the set of authors and the set of books. You need something like this:

Screenshot 2025-12-23 at 17 39 03

Authors are associated with books. There's a subset of books with `stock > 100`. And a subset of authors have a book within this subset.

assets/cxl/sets-intersection.drawio.svg
agoerler, line 1: See above. The set of authors and the set of books don't intersect.

assets/cxl/sets-leftjoin.drawio.svg
agoerler: > explaining JOINs

But we don't want to explain joins! We have sets of instances of some entity type that are associated with some other instances of in a different set. Sometimes our SQL backends use a JOIN in SQL to compute a cross product of these sets. But this is an implementation detail of the SQL backends.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
 By clicking on the individual blocks, you can get more information about the respective building block.
 
 The following diagram illustrates how to read the diagrams:
 
 <div class="diagram" v-html="intro"></div>
 
+### theoretical background
+
+CAP did not re-invent when it comes to expressions.
+It rather builds upon well-known concepts from relational databases and SQL.
```
agoerler: > well-known concepts from relational databases and SQL

not quite. We discourage from using explicit joins or unions. Instead we promote using path expressions.

Where are path expressions and infix filters inspired from?

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+<Badge class="badge-inline" type="tip" text="💡 clickable diagram" />
+<div  v-html="expr"></div>
+</div>
 
+TODO: Some samples --> Where can we use expressions?
 
-TODO: some text and more examples
+- annotation expression
+- calculated element
+- one in select
```
agoerler: view definitons
at runtime

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
 
 <div class="diagram">
   <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> 
   <div v-html="ref"></div>
 </div>
 
-## path segment { #path-segment }
+::: info 💡 Leaf elements
+Leaf elements as opposed to associations and structured elements represent scalar values, such as strings, numbers, dates, as well as the array and map types.
+They typically manifest as columns in database tables.
```
agoerler, line 358: > They typically manifest as columns in database tables.

... if you execute the expression against a relational database 
... or in a [path expression](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31361034) in OData

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
 
-<div class="diagram">
-  <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> 
-  <div v-html="pathSegment"></div>
-</div>
+In its simplest form, a `ref` can be used to reference an element:
+
+:::code-group
+```js [CQL] {1}
+> await cds.ql`SELECT from Books { title }` // [!code focus]
```
agoerler, line 367: > await cds.ql`SELECT from Books { title }`

I think this `await cds.ql` adds little value. This chapter is about explaining the QL. We think it's helpful to explain the QL in the context of a Select statement. But why do we need to put this Select statement` in the context of `await cds.ql`? This makes it confusing, I think.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+    sap_capire_bookshop_Books AS Books
+    LEFT JOIN sap_capire_bookshop_Authors AS author -- The table alias for association 'author'
+        ON author.ID = Books.author_ID;
 ```
 :::
 
-In this example, we select the names of the authors and genres of books.
-Both `author` and `genre` are associations on the `Books` entity.
+In this example, we select all books together with the name of their author.
+`author` is an association in the `Books` entity.
```
agoerler:
```suggestion
The association `author` that is defined in the `Books` entity relates a book to it's author.
```

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+]
 ```
 
 ```sql [SQL]
-SELECT FantasyAuthors.name
-FROM sap_capire_bookshop_Authors AS FantasyAuthors
-WHERE EXISTS (
+SELECT Authors.name
+FROM sap_capire_bookshop_Authors as Authors
+WHERE exists (
```
agoerler: CAP Java would use an inner join here. This semi-join might be advantageous performance wise. But it's a bit harder to understand the navigation from the set of books to the associated set of authors. 

In CAP Java we do:

```sql
Select T1.* from Books T0 inner join Authors T1 on T0.author_ID = T1.ID
```

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+A path expression can also be used in the `from` clause of a query to navigate to a related entity:
 
 :::code-group
 ```js [CQL]
-> await cds.ql`
-  SELECT from ${Authors} as FantasyAuthors { name }
-  where exists books.genre[name = 'Fantasy']` // [!code focus]
-
-[ { name: 'Richard Carpenter' } ]
+> await cds.ql`SELECT from Books:author { name }` // [!code focus]
```
agoerler: > SELECT from Books:author { name }

Shouldn't it be `Books.author` ? What's the difference?

I think this example is not typical. Here we select all authors which are reachable by navigation from any book.

I'd prefer an example like 

```sql
Select from Order[id = 17].items
```

This much closer to what you find in "real life".

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+  WHERE Books.author_ID = Authors.ID
+)
 ```
 :::
 
-::: info 💡 TODO
-???
+TODO explanation
+
+This is equivalent to writing `SELECT from Authors where exists books`. When combining this with [infix filters](#infix-filter), it allows for quite concise queries.
```
agoerler: > This is equivalent to writing `SELECT from Authors where exists books`

provided there is a `books` association in `Authors` ...

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+:::
 
-This is only the tip of the iceberg.
-A path expression can be much more complex. For example, the individual [path segments](#path-segment)
-themselves can contain expressions by applying [infix-filters](#infix-filter).
-More samples are shown in the upcoming sections.
+In this example, we select all books and order them by the date of birth of their authors.
+The table alias for the `author` association is used in the order by clause of the SQL query.
 
+### after `exists` predicate
```
agoerler:
```suggestion
### in `exists` predicate
```

In the example below `exists books` is the exists predicate. `exists` is merely a keyword.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+
```
agoerler, line 490: We need to explain thoroughly that you may use path expressions that navigate to-many association only with exists!

In `where`, `order by`, and on the select list you can only use path expressions that follow to-one paths (or substructure access), i.e. "single-valued" path expressions.

Unfortunately, we missed to reject "collection-valued" in the places mentioned above. Using "collection-valued" path expressions e.g. on the select list can result in an unexpected cross product of the associated entity sets.

You can use "collection-valued" path expressions on the select list only in expands.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
 <div class="diagram">
 <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> 
 <div v-html="infixFilter"></div>
 </div>
 
-### enhancing path expression with filter conditions
+
+::: info 💡 infix notation as a way to influence auto-generated subqueries
+Within an infix, more than than just a simple `WHERE` condition can be specified.
+It is also possible to use other query modifiers, such as `GROUP BY`, `HAVING`, `ORDER BY`, `LIMIT`, and `OFFSET`.
```
agoerler: > Within an infix, more than than just a simple `WHERE` condition can be specified.
It is also possible to use other query modifiers, such as `GROUP BY`, `HAVING`, `ORDER BY`, `LIMIT`, and `OFFSET`.

I think this is danger zone! There's certainly on runtime support in CAP Java here.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+<Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> 
+<div v-html="infixFilterFull"></div>
+</div>
+
+  ::: warning
+  Query modifiers other than `WHERE` may only be used in the context of nested projections
+  and `exists` predicates. They are ignored for regular path expressions.
+
+
+:::
```
agoerler: danger zone!

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+### applied to `expand`
+
+Further narrow down the result set of a path expression by applying an infix filter condition to
+[nested-expands](./cql#nested-expands):
+
+::: code-group
+```js [CQL] {3}
+> await cds.ql`
+    SELECT from Authors { name,
+      books[ price < 19.99 ] as cheapBooks {
```
agoerler:
```suggestion
      books[price < 19.99] as cheapBooks {
```

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+      inner JOIN sap_capire_bookshop_Genres as genre
+        ON genre.ID = books.genre_ID
+    WHERE books.author_ID = Authors.ID
+      and genre.name = 'Fantasy'
+  )
+```
+:::
+
+> Note how the infix filter condition `genre.name = 'Fantasy'` is applied to the
+subquery following the `exists` predicate for the `books` association.
```
agoerler:
```suggestion
the `exists`-subquery for the `books` association in SQL .
```

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+    ]
+  },
+  { name: 'Richard Carpenter', cheapBooks: [] }
+]
+```
+
+```sql [SQL]
+SELECT Authors.name,
+  (
+    SELECT jsonb_group_array(jsonb_insert('{}', '$."title"', title)) as _json_
```
agoerler: > jsonb_group_array(jsonb_insert('{}', '$."title"', title)) as _json_

??? This CAP Node.js SQL is likely confusing to the reader.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+  ) as cheapBooks
+FROM sap_capire_bookshop_Authors as Authors
+```
+:::
+
+<div class="node">
+
+::: info 💡 JSON functions
+
+In this example, the runtime makes use of JSON functions to aggregate the related `books` into a JSON array.
```
agoerler:
```suggestion
In this example, the CAP Node.js runtime makes use of JSON functions to aggregate the related `books` into a JSON array. The CAP Java runtime avoids using JSON functions but executes multiple selects for the same result:

```sql
Select T0.* from Authors T0; 
Select T1.title, T1.price from Authors T0 
         left outer join Books T1 
         on T0.ID = T1.author_ID and T1.price < 19.99;
```
```

File: cds/cxl.md [hunk 51]
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+> TODO: Link to guide about JSON functions, What about java?
+:::
+
+</div>
+
+
+### with query modifiers
+
+It is also possible to influence the generated subquery,
+by adding other query modifiers, such as `GROUP BY`, `HAVING`, `ORDER BY`, `LIMIT`, and `OFFSET`:
```
agoerler: This is new to me. I don't think we support this in CAP Java.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+          and cheapBooks.price < 19.99
+        ORDER BY title DESC -- ORDER BY added to subquery
+      )
+  ) as cheapBooks
+FROM sap_capire_bookshop_Authors as Authors
+```
+:::
+
+
+TODO: add some short explanation and the limitations
```
agoerler: Limitation for CAP Java:

we support expand with limit and order by programatically but if a nested projection uses limit and order by this would not be supported, I'm afraid.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+:::
+
+TODO: explanation and intro to the next sample with [query modifiers](#with-query-modifiers)
+
+
+:::code-group
+
+```js [CQL] {3,4,5}
+await cds.ql`
+  SELECT from Books[
```
agoerler: > Books[
    where stock > 0
    group by genre
    order by genre asc
  ]
  
  Not sure how I understand which set this is. Is this a set of entities, which consist only have a single element "genre"? Why is this set sorted. This is all a bit confusing.
  
  I have never seen these constructs.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+```js
+await cds.ql`
+  SELECT from Books
+  {
+    avg(price) as avgPrice,
+    genre.name
+  }
+  where stock > 0
+  group by genre
+  order by genre asc`
```
agoerler:
```suggestion
  group by genre.name
  order by genre.name`
```

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+WHERE Books.stock > 0
+GROUP BY Books.genre_ID
+ORDER BY Books.genre_ID ASC
+```
+:::
+
+::: details The above is equivalent to…
+
+Using the infix notation to specify the query modifiers is just
+syntactic sugar:
```
agoerler: No. Using the infix notation to specify the query modifiers makes it really hard to understand (and explain) what the infix does. I think the plain infix filter is easy to understand though.

cds/cxl.md
```diff
@@ -62,176 +69,638 @@ cds repl --run .
...
+
+```js
+await cds.ql`
```
agoerler: this example is easy to understand

cds/cxl.md
```diff
@@ -304,106 +758,172 @@ the `years_between` and `startswith` functions are in the [set of CAPs standard
...
-::: code-group
-```js [CAP Style]
-> await cds.ql`
-  SELECT from Authors {
-    name,
-    string_agg(books.title, ', ' ORDER BY books.title DESC) as titles
-  }
-  GROUP BY books.author.ID`
 
+## ordering term <Badge class="badge-inline" type="tip" text="💡 clickable diagram" /> { #ordering-term }
```
agoerler: This is related to the query language and not to the expression language, I think. Move to CQL chapter?

cds/cxl.md
```diff
@@ -304,106 +758,172 @@ the `years_between` and `startswith` functions are in the [set of CAPs standard
...
+Filters progressively narrow down the set:
+
+$$\text{highstock} = \{ b \in \text{Books} \mid b.\text{stock} > 100 \}$$
+
+With the infix filter notation, we write it as `Books[stock > 100]`.
+An association defines a relationship between two sets:
+
+$$\text{books} = \{ (a,b) \in \text{Books} \times \text{Authors} \mid b.\text{author\_id} = a.\text{id} \}$$
+
+We can select this set using the path expression `Authors:books` in the [from clause](#in-from-clause).
```
agoerler: why not `Authors.books` ?

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+FROM sap_capire_bookshop_Authors as Authors
+WHERE exists (
+  SELECT 1
+  FROM sap_capire_bookshop_Books as Books
+  WHERE Books.author_ID = Authors.ID
+)
+```
+:::
+
+TODO explanation
```
agoerler: The path expression `Books:author` denotes the set of all author entities that can be reached by following the association `author` from any entity `Books`. I.e. the set of authors that have written a book.

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+
+This is equivalent to writing `SELECT from Authors where exists books`. When combining this with [infix filters](#infix-filter), it allows for quite concise queries.
+
+### in the where clause {#in-where-clause}
+
+A path expression can also be used as part of the where clause to filter based on elements of related entities:
```
agoerler:
```suggestion
A single-valued path expression can also be used as part of the where clause to filter based on elements of related entities:
```

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+  sap_capire_bookshop_Books AS Books
+  LEFT JOIN sap_capire_bookshop_Genres AS genre
+    ON genre.ID = Books.genre_ID
+WHERE
+  genre.name = ?
+```
+:::
+
+In this example, we select all books that belong to the `Fantasy` genre.
+The table alias for the `genre` association is used in the where clause of the SQL query.
```
agoerler:
```suggestion
```

This implementation detail does not apply to CAP Java. In CAP Java, we use table aliases T0, T1, ..., which is also an implementation detail not worth to mention.

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+  sap_capire_bookshop_Books AS Books
+  LEFT JOIN sap_capire_bookshop_Authors AS author
+    ON author.ID = Books.author_ID
+ORDER BY
+  author.dateOfBirth ASC
+```
+:::
+
+In this example, we select all books and order them by the date of birth of their authors.
+The table alias for the `author` association is used in the order by clause of the SQL query.
```
agoerler:
```suggestion
```

see above

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+### applied to `expand`
+
+Further narrow down the result set of a path expression by applying an infix filter condition to
+[nested-expands](./cql#nested-expands):
+
+::: code-group
+```js [CQL] {3}
+> await cds.ql`
+    SELECT from Authors { name,
+      books[ price < 19.99 ] as cheapBooks {
```
agoerler:
```suggestion
      books[ price < 20.00 ] as cheapBooks {
```

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+```sql [SQL]
+SELECT Authors.name,
+  (
+    SELECT jsonb_group_array(jsonb_insert('{}', '$."title"', title)) as _json_
+    FROM (
+        SELECT cheapBooks.title
+        FROM sap_capire_bookshop_Books as cheapBooks
+        WHERE Authors.ID = cheapBooks.author_ID
+          and cheapBooks.price < 19.99
+      )
```
agoerler: In CAP Java, we'd execute two queries here instead. Therefore the SQL is misleading and hard to understand.

Maybe rather show the translation to OData, which is more straightforward to understand. (Then please do without the `as cheapBooks`, which is only avaliable with the [nest transformation](https://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs03/odata-data-aggregation-ext-v4.0-cs03.html#Transformationnest), I think:

```sql
SELECT from Authors { name, books[ price < 20.00 ] { title, price } }
```

-> 

```
GET Authors?$select=name&$expand=books($filter=price lt 20.00;$select=title,price)
```

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+```
+:::
+
+<div class="node">
+
+::: info 💡 JSON functions
+
+TODO: This is database specific
+
+In this example, the runtime makes use of JSON functions to aggregate the related `books` into a JSON array.
```
agoerler:
```suggestion
In this example, the CAP Node.js runtime makes use of JSON functions to aggregate the related `books` into a JSON array.
```

cds/cql-draft.md
```diff
@@ -17,6 +17,221 @@ import functionArgs from '../assets/cxl/function-args.drawio.svg?raw'
...
+This is because SQL databases do not have a native concept of nested result sets.
+
+> TODO: Link to guide about JSON functions, What about java?
```
agoerler: > What about java?

How CAP Java executes this depends the [expand optimization mode](https://pages.github.tools.sap/cap/docs/releases/2023/sep23#optimized-expand) used. We always do multiple queries:

### join (default): 

```sql
SELECT T0.name FROM Authors
SELECT T1.title, T1.price FROM Authors T0 JOIN Books T1 ON T1.author_id = T0.ID AND T1.price < 20.00
```

### parent-keys:

```sql
SELECT T0.ID, T0.name FROM Authors T0 -- yields <set of author IDs>
SELECT T0.tile, T0.price FROM Books T0 WHERE T0.author_id IN (<set of author IDs>) AND T0.price < 20.00
```

### load-single (legacy):

```sql
SELECT T0.ID, T0.name FROM Authors T0 -- yields <set of author IDs>
for(id in <set of author IDs>): 
  SELECT T0.tile, T0.price FROM Books T0 WHERE T0.author_id = id AND T0.price < 20.00
```

### subquery

```sql
SELECT T0.ID, T0.name FROM Authors T0 -- yields <set of author IDs>
SELECT T0.tile, T0.price FROM BOOKS T0 WHERE T0.ID IN (SELECT T1.ID FROM Authors T0 JOIN Books T1 ON T1.author_id = T0.ID AND T1.price < 20.00)
```

(Not 100% sure about `subquery`)

cds/cxl.md
```diff
@@ -204,109 +204,84 @@ FROM
...
-The condition can manifest in multiple ways:
-- In the on condition of a join
-- In the condition that correlates a subquery to a main query
-- To select related entities with an additional query
-:::
-
-### in the from clause {#in-from-clause}
-
-A path expression can also be used in the `from` clause of a query to navigate to a related entity:
+When navigating along a to-many association to a leaf element, the result is flattened:
```
agoerler, line 415: > is flattened

No. You create a cross product of the associated entities. This is nothing we should promote as it might have slipped in by accident. Also it might work only in the translation to SQL. The example below will result in a left outer join of books and authors. Also authors which have not written a book will be included.

Please either expand books from authors or start the selection from books:

```sql
Select from Authors { books { title } as title, name as author }
```

(gives a nested result)

or 

```
Select from Books { title, author.name as author }
```

(will not include authors that have written no books)

The important point here is: the expression language CXL does not make any assumption that it is translated to SQL or how it's converted to SQL. 

In your example below, the path expression `books.title` is just the set of titles of the books which are associated to a particular author via the `Authors` entity's `books` association. Hence the query 

```sql
SELECT from Authors { books.title as title, name as author }
```

should return something like

```
[
  { title: ['Wuthering Heights'], author: 'Emily Brontë' },
  { title: ['Jane Eyre'], author: 'Charlotte Brontë' },
  { title: ['Eleonora', 'The Raven'], author: 'Edgar Allen Poe' },
  { title: ['Catweazle'], author: 'Richard Carpenter' }
]
```

which it doesn't. Instead it returns the "flattened" result which you mention. I think it does so by accident!

To make things worse think about:

```sql
SELECT FROM Authors { 
   name, 
   books.title book, 
   addresses.{city, street} as address, 
   phonenumbers.{country_code, area_code, extension}
}
```

This would result in an unwanted, unexpected 4-way left outer join. 

### I think there is no valid use case for path expression involving to-many associations besides in expands and `exists`!

cds/cxl.md
```diff
@@ -204,109 +204,84 @@ FROM
...
+Since books is a to-many association, we get a flattened result: one entry per author and book title.
 
-This is equivalent to writing `SELECT from Authors where exists books`. When combining this with [infix filters](#infix-filter), it allows for quite concise queries.
+In annotation expressions, the result should often only contain one entry per entry in the annotated entity.
+This can be achieved using the [exists](#after-exists-predicate) predicate.
 
-### in the where clause {#in-where-clause}
 
-A path expression can also be used as part of the where clause to filter based on elements of related entities:
+::: info 💡 Associations are **forward-declared joins**
```
agoerler: I still think  we should sunset the phrase "forward-declared joins"

cds/cxl.md
```diff
@@ -204,109 +204,84 @@ FROM
...
+They provide a convenient way to navigate between related entities without having to define the join conditions manually.
 
-:::code-group
-```js [CQL]
-> await cds.ql`SELECT from Books { title } where genre.name = 'Fantasy'` // [!code focus]
-[ { title: 'Catweazle' } ]
-```
+The join condition is defined **ahead of time** as part of the association.
```
agoerler:
```suggestion
The foreign-key relationship is defined **ahead of time** as part of the association.
```

Akatuoro: I do think join condition is accurate as it does not need to be a foreign key relationship, although it pretty much always includes one. A `join` is the SQL terminology and well understood, even if cxl can also manifest in other ways.

cds/cxl.md
Akatuoro: This was an ask by Daniel -> we should discuss this in our next review meeting

cds/cxl.md
Akatuoro: @danjoa the forward declaration terminology according to [Wikipedia](https://en.wikipedia.org/wiki/Forward_declaration) does not quite match the behavior of associations.

Let's discuss this in our next review meeting.
Earlier discussion: https://github.com/capire/docs/pull/2260#discussion_r2605857105

cds/cql-draft.md
Akatuoro: Thanks for listing these. Imho we should include the explanation around how expands are executed in a database specific section. While a developer should not need to interact with them or at most simply choose an optimization mode, it does help in the understanding of what is actually happening -> demystifying.

assets/cxl/sets-expand.drawio.svg
Akatuoro: Correct. Intersections are possible though if the sets are made union-compatible, see https://github.com/capire/docs/pull/2260#discussion_r2635260008

The more I look at this diagram, the more I like it 👍 It showcases the foreign keys as pointers or "paths" which we can navigate along through path expressions. We can then have multiple diagrams for different queries, each highlighting the selected circles and graying out anything which is not involved in the current query.

Could you share the diagram source or simply commit it as `.drawio.svg`?

agoerler: This was just a sketch done with PowerPoint. Let's check wether I can convert it somehow ...

Akatuoro: Ok, then I can also redraw it as a drawio svg :D I don't think we're going to include PowerPoint in the docs anytime soon

cds/assets/cxl/ref.drawio.svg
Akatuoro: "A filter can only be provided when navigating along associations" - CDS (compiler)(expr-unexpected-filter)

cds/cxl.md
Akatuoro: I also want to get rid of the surrounding cds.ql for the snippets, but it does make them easily reproducible via `cds repl .` in the bookshop repo... as described in the how-to section above.

Ideally, we could just give the cql sample and
- paste it into a repl, maybe `cds repl --ql .`
- execute it directly in the browser

For now though, cds.ql for easily trying it out.

cds/cxl.md
Akatuoro: Yes, but that is more confusing than helpful :D Also, the leaf element would be more akin to a property inside an OData path expression

cds/assets/cxl/literal-value.drawio.svg
Akatuoro: true/false is more explicit than boolean-literal.

Timestamp, yes

cds/cxl.md
```diff
@@ -129,11 +116,29 @@ select from Books {
 <div  v-html="expr"></div>
 </div>
 
-TODO: Some samples --> Where can we use expressions?
+Expressions can be used in various places, for example in annotations:
+
+```cds
+annotate AdminService.Authors:dateOfDeath with @assert: (case
+  when dateOfDeath > $now then 'Cannot be in the future'
```
agoerler: > dateOfDeath > $now

`$now` is a timestamp whereas `dateOfDeath` is a (local) date hence the types  are incompatible.

We should introduce `$today`, which would be quite handy.

cds/cxl.md
```diff
@@ -129,11 +116,29 @@ select from Books {
...
+  when dateOfDeath < dateOfBirth then 'Enter a date after date of birth'
+end);
+```
+
+Or in entity defintions for adding calculated elements:
+
+```cds
+extend Authors with {
+  age = years_between(dateOfBirth, coalesce(dateOfDeath, $now));
```
agoerler: > coalesce(dateOfDeath, $now)

see above. `dateOfDeath` and  `$now` have incompatible types.

cds/cxl.md
```diff
@@ -202,122 +207,97 @@ FROM
...
-
-### in the from clause {#in-from-clause}
-
-A path expression can also be used in the `from` clause of a query to navigate to a related entity:
+When navigating along a to-many association to a leaf element, the result is flattened:
 
 :::code-group
 ```js [CQL]
-> await cds.ql`SELECT from Books:author { name }` // [!code focus]
+> await cds.ql `SELECT from Authors { books.title as title, name as author }` // [!code focus]
```
agoerler, line 419: please void the navigation over the to-many association `books` as it introduces the before mentioned join

cds/cxl.md
```diff
@@ -202,122 +207,97 @@ FROM
...
-  WHERE Books.author_ID = Authors.ID
-)
+  LEFT JOIN sap_capire_bookshop_Books as books
+  ON books.author_ID = Authors.ID
 ```
 :::
 
-TODO explanation
+In this example, we select the book titles together with each author.
+Since books is a to-many association, we get a flattened result: one entry per author and book title.
```
agoerler:
```suggestion
Since books is a to-many association, we get a _joined_ result that repeats every author (name) for every associated book.
```

::: tip Use expand to read to-many associations as structured result
To avoid that the author data is duplicated rather use an expand

```sql
SELECT from Authors { name as author, books { title }  }
```

This query will produce a structured result:

```
[
  { author: 'Emily Brontë', books : [ { title: 'Wuthering Heights' } ] },
  { author: 'Charlotte Brontë', books : [ { title: 'Jane Eyre' } ] },
  { author: 'Edgar Allen Poe', books : [ { title: 'Eleonora' }, { title: 'The Raven' }  ] },
  { author: 'Richard Carpenter', books : [ { title: 'Catweazle' } ] }
]
```
:::

cds/cxl.md
```diff
@@ -202,122 +207,97 @@ FROM
...
 
-This is equivalent to writing `SELECT from Authors where exists books`. When combining this with [infix filters](#infix-filter), it allows for quite concise queries.
+In annotation expressions, the result should often only contain one entry per entry in the annotated entity.
```
agoerler: ?

cds/cxl.md
```diff
@@ -202,122 +207,97 @@ FROM
...
+This can be achieved using the [exists](#in-exists-predicate) predicate.
 
-### in the where clause {#in-where-clause}
 
-A path expression can also be used as part of the where clause to filter based on elements of related entities:
+::: tip Associations are **forward-declared joins**
+They provide a convenient way to navigate between related entities without having to define the join conditions manually.
```
agoerler:
```suggestion
They provide a convenient way to navigate between related entities without having to explicitly specify the join condition (foreign-key relationship) during query execution.
```

cds/cxl.md
```diff
@@ -355,178 +334,99 @@ This allows to filter the target of an association based on certain criteria.
...
-> REVISIT: this does work in the node runtime, but the compiler does not yet support it. How about java?
+Exist predicates with infix filters can also be nested.
+Here we select all authors that have written at least one book in the `Fantasy` genre:
 
 :::code-group
 ```js [CQL]
 > await cds.ql`
     SELECT from Authors { name }
-    where exists books[genre.name = 'Fantasy']` // [!code focus]
+    where exists books[exists genre[name = 'Fantasy']]` // [!code focus]
```
agoerler, line 571: `exists books[genre.name = 'Fantasy']` should also be fine. As `genre.name` is a single-valued path expression there should be no need to wrap this into another exist in CXL.

cds/cxl.md
```diff
@@ -687,66 +551,65 @@ FROM sap_capire_bookshop_Authors as Authors
 :::
 
 
-> TODO: explanation about json functions again? Learn more link? Specific additions?
-
-
 ### between path segments
 
-TODO: simpler example a la `SELECT from Books { title, author[age < 40].name }`
+Assuming you have the [calculated element](#in-calculated-element) age in place on the Authors entity:
```
agoerler:
```suggestion
Assuming you have the [calculated element](#in-calculated-element) `age` in place on the Authors entity:
```

cds/cxl.md
```diff
@@ -687,66 +551,65 @@ FROM sap_capire_bookshop_Authors as Authors
...
+
+```cds
+extend Authors with {
+  age = years_between(dateOfBirth, coalesce(dateOfDeath, $now));
```
agoerler: the types of `dateOfDeath` and  `$now` are incompatible

cds/cxl.md
```diff
@@ -687,66 +551,65 @@ FROM sap_capire_bookshop_Authors as Authors
...
-    ON B.author_ID = author.ID
-    AND FLOOR(
+  sap_capire_bookshop_Books as Books
+LEFT JOIN
+  sap_capire_bookshop_Authors as author
+ON
+  author.ID = Books.author_ID AND floor(
+  (
+    (
+      (cast(strftime('%Y', coalesce(author.dateOfDeath,session_context('$now'))) as Integer) - cast(strftime('%Y', author.dateOfBirth) as Integer)) * 12
```
agoerler, line 791: This SQL is quite complicated and distracts from understanding the CXL expression. Maybe just use a stored calculated element ("on write"). This would then simplify the SQL of the query execution.

cds/cxl.md
```diff
@@ -687,66 +551,65 @@ FROM sap_capire_bookshop_Authors as Authors
...
 :::code-group
-```js [CQL] {4}
-> await cds.ql`
-  SELECT from Books { title }
-  where startswith(
-    author[ years_between(dateOfBirth, dateOfDeath) < 40 ].name,
-    'Emily'
-  )`
+```js [CQL]
+> await cds.ql `SELECT from Books { title, author[age < 40].name as author }` // [!code focus]
```
agoerler, line 768: This query does not match the explanation above: "In this case we want to select all books where the author's name starts with `Emily` and the author is younger than 40 years."

Instead, we select _all_ books but the author is only included in the result if the author's age is blow 40.

cds/cxl.md
```diff
@@ -687,66 +551,65 @@ FROM sap_capire_bookshop_Authors as Authors
...
+            (cast(strftime('%d%H%M%S%f0000', coalesce(author.dateOfDeath,session_context('$now'))) as Integer) < cast(strftime('%d%H%M%S%f0000', author.dateOfBirth) as Integer)) * -1
+        end
+      )
+    )
+  ) / 12) < ?
+  ```
 :::
 
-The path expression `author[ years_between(dateOfBirth, dateOfDeath) < 40 ].name`
+The path expression `author[ age < 40 ].name`
```
agoerler, line 808: > The path expression `author[ age < 40 ].name` navigates along the `author` association of the `Books` entity.

The path expression `author[ age < 40 ].name` navigates along the `author` association only if the author's age is below 40.

or

The `author` association of the `Books` entity is filtered for authors with age below 40.

cds/cxl.md
```diff
@@ -687,66 +551,65 @@ FROM sap_capire_bookshop_Authors as Authors
...
 navigates along the `author` association of the `Books` entity.
 
-The join for this path expression is generated as usual and enhanced with the infix filter condition `years_between(dateOfBirth, dateOfDeath) < 40`.
-
-
-::: info 💡 Standard functions
-the `years_between` and `startswith` functions are in the [set of CAPs standard functions](../guides/databases.md#standard-database-functions) and are translated to the respective SQL to get the desired result.
-:::
-
+The join for this path expression is generated as usual and enhanced with the infix filter condition `age < 40`.
```
agoerler: > The join

You are again explaining in terms of join and SQL. From my point of view the join is an implementation detail. The importand part is that `author[age < 40]` is a filtered association that gives access to a book`s associated `Authors` entity only if this author's age is below 40. You don't need SQL or joins to understand this. Just set theory.

cds/cxl.md
```diff
@@ -771,7 +634,7 @@ that operates on the single operand `price`. It negates the value of `price`.
 </div>
 
 
-::: info 💡 A binary operator is an operator that operates on two operands.
+::: info A binary operator is an operator that operates on two operands.
```
agoerler: Here would be the place to document which operators there are. That `=` does a three-valued comparison whereas `==` does a two-valued comparison. What's the precedence of operators. That we support `||` for string concatenation. This is all not documented so far. But it's very fundamental. We can't explain infix filters without explaining filter predicates.

cds/cxl.md
Akatuoro: We tried this in java and it did not work:
```java
		CqnSelect select = Select.from(Authors_.class).columns(a -> a.name())
		.where(a -> a.books()
			.filter(b -> b.genre().name().eq("Fantasy"))
			.exists());
		Result res = db.run(select);
```

```json
{
  "SELECT": {
    "from": { "ref": ["AdminService.Authors"] },
    "columns": [{ "ref": ["name"] }],
    "where": [
      {
        "xpr": [
          "EXISTS",
          {
            "ref": [
              {
                "id": "books",
                "where": [
                  { "ref": ["genre", "name"] },
                  "=",
                  { "val": "Fantasy" }
                ]
              }
            ]
          }
        ]
      }
    ]
  }
}
```
```
Path {\"ref\":[\"genre\",\"name\"]} in infix filter is not supported 
```

Also, we like the sequence of examples. We first have a very simple infix filter and then we show that we can also nest infix filters. We'd like to also add `exists books[genre.name = 'Fantasy'] ` again once the compiler and Java supports it. Currently, it only works in cds.ql.

cds/cxl.md
Akatuoro: The SQL is only shown when navigating explicitly to the SQL tab in the code group. Yes, it is very complex, but you usually don't see it... unless you want to

cds/cxl.md
Akatuoro: I see CQL as very SQL-like and many parts of CQL are also the same as SQL. So in my opinion it makes sense to draw the analogy.

Still, I agree that this sentence is not needed with the adjusted sentence above which you suggested.

cds/cxl.md
Akatuoro: Yes, we still need to document the operators (and also the literals) more.

Here's also some explanation for operators, but interestingly only for the node impl variant:
https://cap.cloud.sap/docs/guides/databases?impl-variant=node#standard-operators

Java then has its own explanation in the Java CQL docs:
https://cap.cloud.sap/docs/java/working-with-cql/query-api#comparison-operators

I like the way it is documented in Java, we can probably do something similar here. Do you want to contribute this?

cds/cxl.md
Akatuoro: It's about the point that you typically don't want to do the flattening / joining in annotation expressions. Since your point is that the joining should not be done anyways, this sentence would be unnecessary. But we'll have to discuss the flattening / joining together in a meeting

cds/cxl.md
Akatuoro: As [mentioned](https://github.com/capire/docs/pull/2260#discussion_r2667589016), it was requested by Daniel -> need to discuss this together. Until then, I'm keeping it in the PR

cds/assets/cxl/binary-operator.drawio.svg
agoerler, line 1: The new shape looks much better :-)

cds/assets/cxl/literal-value.drawio.svg
agoerler, line 1: The vertical alignment of the text is off

cds/cxl.md
```diff
@@ -139,6 +139,174 @@ Or as part of a query:
 SELECT from Books { title } where genre.name = 'Fantasy'
 ```
 
+### in model definitions
+
+Expressions can be used to define calculated elements.
+Typically, this is done in a column of a query. CAP
```
agoerler:
```suggestion
Typically, this is done on the select list of a query. CAP
```

cds/cxl.md
```diff
@@ -139,6 +139,174 @@ Or as part of a query:
...
+
+the `years_between` function is one of CAPs portable functions that calculates the number of years between two dates.
+
+
+[Learn more about calculated elements](./cdl.md#calculated-elements){ .learn-more }
+[Learn more about CAPs portable functions](../guides/databases/cql-to-sql.md#portable-functions){ .learn-more }
+
+### in queries
+
+Expressions can be used in various parts of a query, e.g., in the select list, in the where clause, in order by clauses, and more:
```
agoerler:
```suggestion
Expressions can be used in various parts of a query, e.g., on the select list, in the where clause, in order by clauses, and more:
```

cds/cxl.md
```diff
@@ -656,36 +865,7 @@ that multiplies the two factors `price` and `quantity`.
...
-| `sum(x)`                          | Returns the sum of all values of `x`.                                                                                                                   |
-| `average(x)`                      | Returns the average (mean) value of `x`.                                                                                                                |
-| `count(x)`                        | Returns the count of non-null values of `x`.                                                                                                            |
-| `countdistinct(x)`                | Returns the count of distinct non-null values of `x`.                                                                                                   |
-
+CAP supports a set of [portable functions](../guides/databases/cql-to-sql.md#portable-functions) that can be used in all expressions. Those functions are passed through to the underlying database, allowing you to leverage the same functions for different databases, which greatly enhances portability.
```
agoerler: Rather define the portable functions here (as they are DB-independent and also apply for to remote OData calls). Then link from the database guide to here.

cds/assets/cxl/literal-value.drawio.svg
Akatuoro: Not in the rendered version :D but I adjusted it so that it is centered in both drawio and the rendered version

cds/cxl.md
```diff
@@ -110,96 +109,59 @@ select from Books {
...
 Expressions can be used to define calculated elements.
-Typically, this is done in a column of a query. CAP
+Typically, this is done on the select list of a query. CAP
 also allows to define calculated elements directly in the model:
 
 ```cds
-extend Authors with {
-  age = years_between(dateOfBirth, coalesce(dateOfDeath, date( $now )));
+extend Books with {
+  total = price * stock;
```
agoerler, line 132: please add a type
```suggestion
  total = price * stock : Decimal(8,2);
```

cds/assets/cxl/literal-value.drawio.svg
patricebender: I found it quite weird that I had to use "align to bottom" in order to make it "align centered" in the rendered version 😄 thx Johannes!

cds/cxl.md
```diff
@@ -308,6 +273,75 @@ The `@assert` annotation lets you capture the intent via an expression, without
...
+> cds.parse.expr ` true `
+{ val: true }
+> cds.parse.expr ` false `
+{ val: false }
+> cds.parse.expr ` Date'2026-01-01' `
+{ val: '2026-01-01', literal: 'date' }
+> cds.parse.expr ` Time'08:42:15.000' `
+{ val: '08:42:15.000', literal: 'time' }
+> cds.parse.expr ` TimeStamp'2026-01-14T10:30:00Z' `
+{ val: '2026-01-14T10:30:00Z', literal: 'timestamp' }
```
agoerler, line 302: This is ISO 8601. Is this portable on all DBs? I.e., will 

```cds
entity Start : cuid {
  start : Timestamp default TimeStamp'2026-01-14T10:30:00Z'
}
``` 

actually work?  

Or rather use ISO/IEC 9075

```suggestion
> cds.parse.expr ` TimeStamp'2026-01-14 10:30:00' `
{ val: '2026-01-14 10:30:00', literal: 'timestamp' }
```

?

agoerler: CAP Java additionally supports these literal types

* number
* x
* hex
* boolean
* string
* vector

cds/cxl.md
```diff
@@ -531,6 +518,9 @@ WHERE exists (
 
 [Learn more about the `exists` predicate.](./cql.md#exists-predicate){.learn-more}
 
+The `exists` predicate can be further enhanced by [combining it with infix filters](#exists-infix-filter).
```
agoerler, line 521: wrap as "tip"?

PR #2223, opened 2025-11-21
Revised security guide

BraunMatthias:

PR #2264, opened 2025-12-10
Adding node specific Snippets & Sections (CAP User)

PDT42:
Adding node specific snippets and sections to the revised security guide, originally added in https://github.com/capire/docs/pull/2223.

### DIFF
--- a/guides/security/cap-users.md
+++ b/guides/security/cap-users.md
@@ -993,7 +993,7 @@ srv.before('READ', srv.entities.Books, req => {
 })
 ```
 
-In addition to the request context, information about the current user can similarly be retrieved from the global [`cds.context`](#TODO), which provides access to the current `cds.EventContext`: 
+In addition to the request context, information about the current user can similarly be retrieved from the global [`cds.context`](../../node.js/events#cds-context), which provides access to the current [`cds.EventContext`](../../node.js/events#cds-event-context): 
 
 ```js
 const cds = require('@sap/cds')
@@ -1001,7 +1001,7 @@ const { user, tenant } = cds.context
 ```
 
 :::tip
-Prefer local req objects in your handlers for accessing event context properties, as each access to cds.context happens through [AsyncLocalStorage.getStore()](#TODO), which induces some minor overhead.
+Prefer local req objects in your handlers for accessing event context properties, as each access to cds.context happens through [AsyncLocalStorage.getStore()](https://nodejs.org/api/async_context.html#asynclocalstoragegetstore), which induces some minor overhead.
 :::
 
 Setting `cds.context` usually happens in inbound authentication middlewares or in inbound protocol adapters.
@@ -1018,6 +1018,8 @@ Depending on the configured [authentication](./authentication) strategy, CAP der
 
 </div>
 
+
+<!-- TODO: Pseudo Role Checks in Node? cds.User.is('privilged')? -->
 
 ### Customizing Users { #customizing-users }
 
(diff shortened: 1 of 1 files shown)


guides/security/cap-users.md
```diff
@@ -646,21 +724,55 @@ c.s.c.s.a.c.AmsRuntimeConfiguration      : Configured AmsUserInfoProvider
...
 "accessResult":"or( eq($app.Genre, "Mystery") eq($app.Genre, "Fantasy") )"}.
 ```
 
+</div>
+
+<div class="impl node">
+
+```sh
+[ams] - Computing AMS filter conditions for ...
+[ams] - Privilege check for ... on ... was granted. {...
```
finkmanAtSap:
```suggestion
[ams] - Privilege check for ... on ... was conditional. {...
```

The suggestion is the more common log message encountered by developers. When a privilege check event is logged by the AMS CAP plugin, it is to compute filter conditions. Most of the time, the user will not have unrestricted access (=granted) but filtered access (=conditional) in this case.

guides/security/cap-users.md
```diff
@@ -839,27 +957,25 @@ The `UserInfo` object is not modifyable, but during request processing, a new `R
...
-|---------------|---------------------|----------------------------------|-------------------------|--------------------|
-| _Logon name_     | `getName()`    | `user_name`                      | `sub`                   | `$user`  |
-| _Tenant_   | `getTenant()`  | `zid`                            | `zone_uuid`             | `$user.tenant` |
-| _Attributes_ | `getAttributeValues(String attr)` | `xs.user.attributes.<attr>` | All non-meta attributes | `$user.<attr>` |
-| _Roles_     | `getRoles()` and `hasRole(String role)` | `scopes`  | n/a - injected via AMS | String in `to`-clause |
+| User Property | UserInfo Getter                         | XSUAA JWT Property          | IAS JWT Property        | `@restrict`-annotation |
+|---------------|-----------------------------------------|-----------------------------|-------------------------|------------------------|
+| _Logon name_  | `getName()`                             | `user_name`                 | `sub`                   | `$user`                |
+| _Tenant_      | `getTenant()`                           | `zid`                       | `zone_uuid`             | `$user.tenant`         |
```
finkmanAtSap: Is it really `zone_uuid` for the tenant in IAS? Afaik, the zone concept has been deprecated and app_tid is to be used. zone_uuid is typically the same value like app_tid though and still contained in the token for backward-compatibility

guides/security/cap-users.md
```diff
@@ -1108,6 +1221,187 @@ Prefer using [Remote Services](#remote-services) built on Cloud SDK rather than
...
+
+Setting `cds.context` usually happens in inbound authentication middlewares or in inbound protocol adapters.
+During processing, you can set it programmatically or spawn a new root transaction providing a context argument to achieve a [switch of the current user](#switching-users--switching-users-node).
+
+Depending on the configured [authentication](./authentication) strategy, CAP derives a default set of user claims containing the user's name, tenant, attributes and assigned roles:
+
+| User Property | UserInfo Getter                     | XSUAA JWT Property          | IAS JWT Property        | `@restrict`-annotation |
+|---------------|-------------------------------------|-----------------------------|-------------------------|------------------------|
+| _Logon name_  | `user.id`                           | `user_name`                 | `sub`                   | `$user`                |
+| _Tenant_      | `req.tenant` / `cds.context.tenant` | `zid`                       | `zone_uuid`             | `$user.tenant`         |
```
finkmanAtSap:
```suggestion
| _Tenant_      | `req.tenant` / `cds.context.tenant` | `zid`                       | `app_tid`             | `$user.tenant`         |
``

File: guides/security/cap-users.md [hunk 4]
```diff
@@ -1108,6 +1221,187 @@ Prefer using [Remote Services](#remote-services) built on Cloud SDK rather than
...
+| _Attributes_  | `attr(attr)`                        | `xs.user.attributes.<attr>` | All non-meta attributes | `$user.<attr>`         |
+| _Roles_       | `roles()` and `is(role)`            | `scopes`                    | n/a - injected via AMS  | String in `to`-clause  |
```
finkmanAtSap: AMS does not affect the behaviour of roles(). Our last information is that only `user.is` is part of the public API and that `roles` is an internal property that we cannot expect to exist on all user objects, e.g. those created by custom auth middlewares. So we only decorate `user.is` based on ams roles.

We should sync if this has changed, so that both methods behave consistently.

guides/security/cap-users.md
finkmanAtSap: It looks like it is indeed `app_tid`

```suggestion
| _Tenant_      | `getTenant()`                           | `zid`                       | `app_tid`             | `$user.tenant`         |
```

PDT42: Good catch! And thanks for doing the research!

guides/security/cap-users.md
PDT42: You are right! `roles()` is not included in the [public documentation](https://cap.cloud.sap/docs/node.js/authentication#cds-user). I will remove `roles()` from this table accordingly 👌

renejeglinsky:
We'd like to avoid adding more toggled content, as the usage of the toggle brings a lot of problems and is discouraged. Would you like to prepare the content or should I give it a try and propose a change?

BraunMatthias:
> We'd like to avoid adding more toggled content, as the usage of the toggle brings a lot of problems and is discouraged. Would you like to prepare the content or should I give it a try and propose a change?

please let's optimize on the target branch after having merged this branch - otherwise we'll have to face lots of merge conflicts in the next days.

renejeglinsky:
Agreed :)

PR #2277, opened 2025-12-12
Tree view

stewsk:


advanced/fiori.md
```diff
@@ -745,9 +745,9 @@ Cache Control feature is currently supported on the Java runtime only.
 
 ## Hierarchical Tree Views
 
-Recursive hierarchies are parent-child hierarchies, where each entity references its parent and through that defines the hierarchical structure. A common example is a company organization structure or HR reporting, where each employee entity references another employee as a direct report or manager.
+Recursive hierarchies are parent-child hierarchies: each entity references its parent and through that defines the hierarchical structure. A common example is a company organization structure or HR reporting, where each employee entity references another employee as a direct report or manager.
```
renejeglinsky: It could be just me, but do we need to make the distinction clearer between hierarchy and composition? I guess it's clear through the context of the Fiori guide, but maybe it's short-sighted to say: hierarchy is needed for UIs. In addition I think hierarchy has not this "contained-in" relationship. Is that obvious?

renejeglinsky: Maybe we could add a info box about parent-child is not equal to contained-in?

renejeglinsky: But again, if it's just me, we can skip it.

renejeglinsky: Well, I think users in this guide will have touched the domain modelling guide already. In case they didn't, maybe only a learn more towards the domain modelling guide is enough at this point.

advanced/fiori.md
```diff
@@ -774,19 +775,64 @@ service AdminService {
...
+> Note: construct the `hierarchyQualifier` with the following pattern: <br>
+> `<entity name in service>Hierarchy`
+
+You can now start the server with `cds watch` and see the hierarchical tree view in action in the [_Browse Genres_](http://localhost:4004/fiori-apps.html#Genres-display) app.
+
+![Fiori UI with hierarchical tree view.](assets/hierarchical-tree-view.png) {style="filter: drop-shadow(0 2px 5px rgba(0,0,0,.40));"}
+
+The compiler automatically expands the shortcut annotation `@hierarchy` to the
+following annotate and extend statements. If you cannot use the `@hierarchy` annotation,
+e.g. because you only have an unmanaged parent association, you can write them yourself.
```
renejeglinsky: I understand that the following is the the manual approach. Maybe we spent a heading for this?


```suggestion
The compiler automatically expands the shortcut annotation `@hierarchy` to the
following annotate and extend statements. 

### Manual Approach

The following documents what happens behind the scenes, done by the compiler as described before. You can also use it, if you cannot use the `@hierarchy` annotation, for example, because you only have an unmanaged parent association.
```

stewsk: There are other situations (besides "unmanaged assoc") where the compiler doesn't generate the hierarchy stuff, thus the heading should be  it more general. What about "Manual approch" or similar?

renejeglinsky: Edited my previous suggestion. Please have a look. Stating that it also describes the "Behind the scenes" is correct?

advanced/fiori.md
eugene-andreev: Hierarchies are hierarchies :)

may be
```suggestion
Recursive hierarchies are parent-child related structures
```
?

advanced/fiori.md
```diff
@@ -774,19 +775,64 @@ service AdminService {
...
+
+> Note: construct the `hierarchyQualifier` with the following pattern: <br>
+> `<entity name in service>Hierarchy`
+
+You can now start the server with `cds watch` and see the hierarchical tree view in action in the [_Browse Genres_](http://localhost:4004/fiori-apps.html#Genres-display) app.
+
+![Fiori UI with hierarchical tree view.](assets/hierarchical-tree-view.png) {style="filter: drop-shadow(0 2px 5px rgba(0,0,0,.40));"}
+
+The compiler automatically expands the shortcut annotation `@hierarchy` to the
+following annotate and extend statements. If you cannot use the `@hierarchy` annotation,
```
eugene-andreev:
```suggestion
following `annotate` and `extend` statements. If you cannot use the `@hierarchy` annotation,
```

advanced/fiori.md
renejeglinsky:
```suggestion
Recursive hierarchies are parent-child related structures: each entity references its parent and through that defines the hierarchical structure. A common example is a company organization structure or HR reporting, where each employee entity references another employee as a direct report or manager.
```

advanced/fiori.md
```diff
@@ -774,19 +775,67 @@ service AdminService {
...
+
+> Note: construct the `hierarchyQualifier` with the following pattern: <br>
+> `<entity name in service>Hierarchy`
+
+You can now start the server with `cds watch` and see the hierarchical tree view in action in the [_Browse Genres_](http://localhost:4004/fiori-apps.html#Genres-display) app.
+
+![Fiori UI with hierarchical tree view.](assets/hierarchical-tree-view.png) {style="filter: drop-shadow(0 2px 5px rgba(0,0,0,.40));"}
+
+The compiler automatically expands the shortcut annotation `@hierarchy` to the
+following annotate and extend statements.
```
renejeglinsky:
```suggestion
following `annotate` and `extend` statements. 
```

advanced/fiori.md
stewsk: This section comes from the original tree view docu by the runtimes, haven't touched it.
I don't care if we give an explanation or not -> @renejeglinsky to decide

renejeglinsky (approved):

PR #2292, opened 2025-12-17
Review of @assert docs

renejeglinsky:


guides/providing-services.md
```diff
@@ -941,7 +941,7 @@ The `@assert.target` check constraint relies on database locks to ensure accurat
 
 ### `@assert` <Beta/>
 
-Annotate an _element_ with `@assert` to define the CXL expressions that are validated _after_ the data has been written to the database but _before_ it is committed. If the validation fails, the expression will return a `String` that indicates an error to the runtime. If the validation passes the expression return `null` 
+Annotate an element with `@assert` to define CXL expressions that validate data after you write it to the database but before the system commits it. If validation fails, the expression returns a `String` that indicates an error to the runtime. If validation passes, the expression returns `null`.
```
StefanHenke:
```suggestion
Annotate an element with `@assert` to define CXL expressions that are validated _after_ the data has been written to the database but _before_ it is committed it. If validation fails, the expression returns a `String` that indicates an error to the runtime. If validation passes, the expression returns `null`. 
```

Personally, I would leave the first part as it was because the proposal sounds like the user would have influence on this. This is not the case.

StefanHenke (approved):
just a small comment

guides/providing-services.md
smahati:
```suggestion
The `@assert` annotation lets you define a CXL expression that is used to validate data _after_ it's been written to the database but _before_ it's committed. If validation fails, the expression returns a `String` that indicates an error to the runtime. If validation passes, the expression returns `null`. 
```

guides/providing-services.md
```diff
@@ -965,23 +965,23 @@ entity OrderItems : cuid {
 
 #### Error Messages and Message Targets
 
-If a validation fails, the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts) the error is persisted. The error targets the annotated element, which is then highlighted on the Fiori UI.
+If validation fails, the transaction is rolled back with an exception. If you use [Fiori draft state messages](../advanced/fiori#validating-drafts), the error is persisted. The error targets the annotated element, which is then highlighted on the Fiori UI.
```
smahati: The transaction is rolled back only in the default case, should we mention that it's not rolled back when using fiori draft state messages? (unless I'm mistaken)

renejeglinsky: @StefanHenke Is Maha's understanding correct that the transaction in not rolled back when using Fiori draft?
```suggestion
In general, if validation fails, the transaction is rolled back with an exception. But, if you use [Fiori draft state messages](../advanced/fiori#validating-drafts), the error is persisted. The error targets the annotated element, which is then highlighted on the Fiori UI.
```

StefanHenke: Well, it´s a bit more complicated actually :) But the proposal is fine I would say. The important piece is that if you use the draft state messages, this message is really persisted and that part of the transaction is not rolled back. But I guess explaining all the details would reveal our implementation and overwhelm the reader. So, keeping it on that layer is good.

renejeglinsky: Alright, we'll find out through issues/incidents if we need to explain the "more complicated" stuff as well. Thanks!

PR #2319, opened 2026-01-15
Small additions and fixes to initial-data.md

qmacro:
Some tweaks to language specifics, a couple of additions and clarifications, and the removal of what seems to be an accidental repeat of a paragraph on the "in the neighbourhood" concept.

### DIFF
--- a/.vitepress/theme/components/indexFilter.ts
+++ b/.vitepress/theme/components/indexFilter.ts
@@ -1,25 +1,23 @@
+const { base, themeConfig: { sidebar }} = global.VITEPRESS_CONFIG.site
 import { join } from 'node:path'
-import { ContentData, DefaultTheme, SiteConfig} from 'vitepress'
-type SBItem = DefaultTheme.SidebarItem
-
-// @ts-ignore
-const site = (global.VITEPRESS_CONFIG as SiteConfig<DefaultTheme.Config>).site
-const sidebar = site.themeConfig.sidebar! as SBItem[]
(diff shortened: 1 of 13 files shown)


guides/databases/initial-data.md
```diff
@@ -158,13 +158,13 @@ myproject/
 
 All `db/data/*.csv` and  `srv/data/*.csv` files are automatically loaded, because they are located in a `data` folder next to `.cds` model sources.
 
-Additionally, this _in-the-neighborhood-of-models_ technique enables reuse packages, which's main purpose is to provide initial data. Such packages get installed via `npm` or `mvn`, and hence their content will reside under `node_modules` folders, or Maven `target` folders. As long as they also provide models through `.cds` sources, their data will be found and loaded.
+This is especially useful for remote service definitions imported with `cds import` (by default into `srv/external/`) so that data for such services can also be served when mocking.
```
renejeglinsky, line 161: This describes sth different than it did before, doesn't it? Is that on purpose?

qmacro: My addition is intended as just that, an addition. But the deletion (of "Additionally, this ...") is also deliberate, as it occurs pretty much verbatim in the section that follows this, with its own heading.

danjoa: That's correct ... I copied that paragraphs so often, and ultimately lost all the traces out of sight

danjoa:
Thanks DJ! 👍

PR #2321, opened 2026-01-16
[review] Security Guide

renejeglinsky:
General remarks on:
- [Remote Authentication](https://github.com/capire/docs/pull/2321#issuecomment-3758980683)


guides/security/authorization.md
```diff
@@ -413,7 +413,25 @@ The [restrict annotation](#restrict-annotation) for an entity allows you to enfo
 In addition, you can define a `where`-condition that further limits the set of accessible instances. 
 This condition, which acts like a filter, establishes *instance-based authorization*.
 
-### Filter Conditions { #filter-consitions }
+### Filter Conditions
+
+For instance, a user is allowed to read or edit `Orders` (defined with the `managed` aspect) that they have created:
```
renejeglinsky, line 417: This idea was to move the example first, as we often do it in capire. We could do the same in other sections as well, if we agree on doing so. This is just an example to show what I mean.

guides/security/authentication.md
```diff
@@ -309,7 +309,7 @@ Integration tests running in production profile should verify that unauthenticat
  - cross-landscape user propagation (including on-premise)
  - streamlined SAP and non-SAP system [integration](https://help.sap.com/docs/cloud-identity-services/cloud-identity-services/integrating-service) (due to [OpenId Connect](https://openid.net/connect/) compliance)
 
-IAS authentication is best configured and tested in the Cloud, so let's enhance the started bookshop sample application with a deployment descriptor for SAP BTP, Cloud Foundry Runtime (CF).
+IAS authentication is best configured and tested in the Cloud, so let's enhance the [previously started bookshop sample application](#mock-user-authentication) with a deployment descriptor for SAP BTP, Cloud Foundry Runtime (CF).
```
renejeglinsky: It feels kind of hidden that we started an example. WDYT about making "Set up a Sample" an own section so that it can be found easily?

guides/security/authentication.md
```diff
@@ -426,7 +426,11 @@ The startup log should confirm the activated IAS authentication:
 </div>
 
 <div class="node">
+
+```sh
 TODO
+```
```
renejeglinsky, line 482: Needs to be filled

guides/security/authentication.md
```diff
@@ -1337,7 +1341,9 @@ In such architectures, CAP authentication is obsolete and can be deactivated ent
 </div>
 
 <div class="node">
+
 TODO
+
```
renejeglinsky, line 1397: Needs to be filled

guides/security/authentication.md
```diff
@@ -140,9 +140,9 @@ curl http://localhost:4004/odata/v4/admin/Books --verbose
 results in a `401` error response from the server indicating that the anonymous user has been rejected due to missing authentication.
 This is true for all endpoints including the web application page at `/index.html`.
 
-Mock users require **basic authentication**, hence sending the same request on behalf of mock user `alice` (password: `basic`) with
+Mock users require **basic authentication**, hence sending the same request on behalf of mock user `alice` (no password) with
 ```sh
-curl http://alice:basic@localhost:4004/odata/v4/admin/Books
+curl http://alice:@localhost:4004/odata/v4/admin/Books
```
renejeglinsky: @PDT42 Is this correct or does Alice have a password? Please also check the curl command for correctness. Thanks :)

guides/security/authorization.md
```diff
@@ -267,16 +267,30 @@ Restrictions can be defined on different types of CDS resources, but there are s
 
 Unsupported privilege properties are ignored by the runtime. Especially, for bound or unbound actions, the `grant` property is implicitly removed (assuming `grant: '*'` instead). The same also holds for functions:
 
-```cds
+::: code-group
```
renejeglinsky: I feel it does make sense to show a wrong version and the resulting model after implicitly removing it. WDYT?
Please check the resulting model. Or is there a better/right way to show that?

guides/security/authorization.md
```diff
@@ -262,21 +262,35 @@ Restrictions can be defined on different types of CDS resources, but there are s
 | entity          |  <Y/>   | <Y/> | <Y/><sup>1</sup>  |               |
 | action/function |  <Na/>  | <Y/> | <Na/><sup>2</sup> | = `@requires` |
 
-> <sup>1</sup>For bound actions and functions that are not bound against a collection, Node.js supports instance-based authorization at the entity level. For example, you can use `where` clauses that *contain references to the model*, such as `where: CreatedBy = $user`. For all bound actions and functions, Node.js supports simple static expressions at the entity level that *don't have any reference to the model*, such as `where: $user.level = 2`.
+> <sup>1</sup>For bound actions and functions that are not bound against a collection, Node.js supports instance-based authorization at the entity level, see [link] (somewhere in Node.js docs)<br>
```
renejeglinsky: I felt that there was too much text that could be part of the runtime docs. I'm not saying is has to but maybe we can find a place for that and set a link here. Would be preferred.

guides/security/authorization.md
```diff
@@ -262,21 +262,35 @@ Restrictions can be defined on different types of CDS resources, but there are s
...
 > <sup>2</sup> For unbound actions and functions, Node.js supports simple static expressions that *don't have any reference to the model*, such as `where: $user.level = 2`.
```
renejeglinsky, line 266: Should it say "bound and unbound actions and functions"? In consequence we should add `<sup>2</sup>` also in the table for `entity`.
See: 

Image


```suggestion
> <sup>2</sup> For bound and unbound actions and functions, Node.js supports simple static expressions that *don't have any reference to the model*, such as `where: $user.level = 2`.
```

guides/security/authorization.md
renejeglinsky: @BraunMatthias What do you think about changing this?

guides/security/authentication.md
renejeglinsky: @PDT42 or @sjvans Would you be able to fill it?

guides/security/authentication.md
renejeglinsky: @PDT42 or @sjvans would you be able to fill it?

guides/security/authentication.md
PDT42, line 154: Can't create a suggestion on the [original comment](https://github.com/capire/docs/pull/2321/files/f11482fc2a6e653f175d001bce27c083113395dc#r2698482727). 

Default user `alice` does not have a password. Running this curl worked for me after setting up a `tiny-sample`. I think we should use the catalog service, since its available in `sample` as well as `tiny-sample`. 

```suggestion
curl http://alice:@localhost:4004/odata/v4/catalog/Books
```

PDT42: See https://github.com/capire/docs/pull/2321#discussion_r2717555037

guides/security/authentication.md
PDT42: Within our logs, I think this would be the closest we get ... I am not aware of an output that _names_ the service binding, our auth middleware is going to use. 

```suggestion
```sh
[cds] - using auth strategy { 
    kind: 'ias',
    impl: 'node_modules/@sap/cds/lib/srv/middlewares/auth/ias-auth.js'
}
```

guides/security/authentication.md
PDT42: I am not aware of a functional equivalent in node 🤔 People could add a custom auth middleware?

guides/security/authentication.md
renejeglinsky: The context here is that unrestricted services are not authenticated in non-production profiles, see line 130ff. So I think that's why the admin service is used here.

guides/security/authentication.md
renejeglinsky: If there is no way to turn off authentication, we could also just remove this div here.

guides/security/authentication.md
renejeglinsky, line 1385: @BraunMatthias How about moving this note into the java div with the hint how to turn off authentication? If there's is no equivalent in Node.js then we could move it into the Java div as it doesn't apply to Node.js at all. Right?

guides/security/authorization.md
renejeglinsky: Agreed to do this via phone

guides/security/authorization.md
renejeglinsky: @PDT42 Do you have an opinion about it?

guides/security/authorization.md
renejeglinsky: @BraunMatthias What do you think?

guides/security/authentication.md
sjvans: yes, you'd need to do this in a custom auth middleware. please hint to that instead of removing the div.

guides/security/authentication.md
```diff
@@ -466,10 +473,16 @@ The startup log should confirm the activated IAS authentication:
...
 
 <div class="node">
-TODO
+
+```sh
+[cds] - using auth strategy { 
+    kind: 'ias',
+    impl: 'node_modules/@sap/cds/lib/srv/middlewares/auth/ias-auth.js'
+}
+
```
renejeglinsky, line 483:
```suggestion
```

```

File: guides/security/authentication.md [hunk 10]
```diff
@@ -466,10 +473,17 @@ The startup log should confirm the activated IAS authentication:
...
+[cds] - using auth strategy { 
+    kind: 'ias',
+    impl: 'node_modules/@sap/cds/lib/srv/middlewares/auth/ias-auth.js'
+}
+```
+
 </div>
 
-::: tip
+::: tip Local at this point?
```
renejeglinsky:
```suggestion
::: tip
```

guides/security/authentication.md
renejeglinsky: Will remove the div for now. Please fill it in another PR.

guides/security/authentication.md
```diff
@@ -1382,7 +1392,9 @@ In such architectures, CAP authentication is obsolete and can be deactivated ent
 </div>
 
 <div class="node">
+
 TODO
+
 </div>
```
renejeglinsky:
```suggestion
<!--
<div class="node">

TODO

</div>
-->
```

guides/security/authorization.md
renejeglinsky:
```suggestion
> <sup>1</sup>For bound actions and functions that are not bound against a collection, Node.js supports instance-based authorization at the entity level. For example, you can use `where` clauses that *contain references to the model*, such as `where: CreatedBy = $user`. For all bound actions and functions, Node.js supports simple static expressions at the entity level that *don't have any reference to the model*, such as `where: $user.level = 2`.
```

guides/security/authorization.md
renejeglinsky:
```suggestion
```

guides/security/authorization.md
```diff
@@ -262,21 +262,35 @@ Restrictions can be defined on different types of CDS resources, but there are s
...
+```cds [Resulting model]
+service CatalogService {
+  entity Products as projection on db.Products { ... }
+  actions {
+    @(requires: 'Admin') // is already in implicit {grant: '*'}
+    action addRating (stars: Integer);
+  }
+   //unsupported property is removed, means implicit { grant: '*'}
+  function getViewsCount @(restrict: [{ to: 'Admin' }]) () returns Integer; 
 }
```
renejeglinsky:
```suggestion
```cds
service CatalogService {
 entity Products as projection on db.Products { ... }
 actions {
   @(requires: 'Admin')
   action addRating (stars: Integer);
 }
  function getViewsCount @(restrict: [{ grant: 'READ' to: 'Admin' }]) () returns Integer;
}

```

guides/security/authorization.md
```diff
@@ -262,21 +262,35 @@ Restrictions can be defined on different types of CDS resources, but there are s
...
+
 ```
+:::
```
renejeglinsky:
```suggestion
```

guides/security/authorization.md
```diff
@@ -274,11 +274,10 @@ service CatalogService {
     @(requires: 'Admin')
     action addRating (stars: Integer);
   }
-  function getViewsCount @(restrict: [{ to: 'Admin' }]) () returns Integer;
+  function getViewsCount @(restrict: [{ grant: 'READ' to: 'Admin' }]) () returns Integer;
```
renejeglinsky:
```suggestion
  function getViewsCount @(restrict: [{ grant: 'READ', to: 'Admin' }]) () returns Integer;
```

guides/security/authentication.md
sjvans: → https://github.com/capire/docs/pull/2362

sjvans:
> > > > ### Remote Authentication
> > > > That guide is based on a Java sample only, right? Are there plans for Node.js as well?
> > > 
> > > 
> > > @sjvans How would you see that? Is there something to fill from the Node.js perspective?
> > 
> > 
> > @renejeglinsky the same concepts apply, but configs (e.g., `application.yaml` vs `package.json`) differ. we could create a node version/ input for one if you'd like
> 
> Please do so in a new PR. I think it would add value. Matthias was not so sure, that's why I asked you.

added to our backlog

PR #2336, opened 2026-01-22
Remove java mtx classic

BraunMatthias:


java/_menu.md
```diff
@@ -22,7 +22,6 @@
 # [Change Tracking](change-tracking)
 # [Transactional Outbox](outbox)
 # [Multitenancy](multitenancy)
-  ## [Multitenancy (Classic)](multitenancy-classic) <!-- INTERNAL -->
```
renejeglinsky, line 25: This is the only relevant change in this PR, or? ALL the others are in your review PR already (and not up to date). Regarding this, it would be best to merge this PR first and then the review PR with the more recent changes. Right?

BraunMatthias: yesm just wanted to remove the classix mtx from the menu but keep the file for now

PR #2350, opened 2026-01-29
Update CQL to SQL translation details in documentation

renejeglinsky:
@danjoa Some edits by AI that I wanted to get a review on.

---

Clarified CQL behavior regarding != operator and its translation in SQL databases. Updated note on database-specific behavior for portable functions.

### DIFF
--- a/guides/databases/initial-data.md
+++ b/guides/databases/initial-data.md
@@ -81,7 +81,21 @@ Common rules apply to text content in `.csv` files, like:
 | Numeric content should be treated as texts                       | -> enclose in double quotes |
 | Boolean values should be treated as text                         | -> enclose in double quotes |
 
+## Data Loading Order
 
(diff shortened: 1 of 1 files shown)


guides/databases/initial-data.md
```diff
@@ -81,7 +81,21 @@ Common rules apply to text content in `.csv` files, like:
...
+```
+db/data/
+├── sap.capire.bookshop-Authors.csv       # Loads first (parent)
+└── sap.capire.bookshop-Books.csv         # Loads second (references Authors)
+```
+
+The filename prefix ensures correct order: `Authors` loads before `Books` alphabetically. If you encounter foreign key violations, verify your file naming follows dependency order.
+
+> [!tip] Large Datasets
+> For datasets with thousands of rows, consider using database-specific bulk loading tools instead of CSV files. The CSV mechanism loads all data in a single transaction, which can cause memory issues and long startup times.
```
renejeglinsky, line 98: Another proposal by AI. I don't think that this is true, but how is the actual behavior?

PR #2362, opened 2026-02-04
security: deactivate cap authentication for node.js

sjvans:


guides/security/authentication.md
sjvans, line 1386: @BraunMatthias what happens with authorization checks in this mode, i.e., are they (1) skipped or (2) still enforced and fail or (3) still enforced but always pass because this auth mode "never" results in some kind of "privileged mode"?

BraunMatthias (approved):

renejeglinsky:
superseeded by / integrated in #2759

PR #2363, opened 2026-02-04
[CDS/CXL] Edit CXL guide

renejeglinsky:
You can ignore the changes to the databases guides. They should not have been in this PR/branch.


cds/cxl.md
```diff
@@ -340,11 +340,11 @@ that multiplies the two factors `price` and `quantity`.
 </div>
 
 
-CAP supports a set of [portable functions](../guides/databases/cap-level-dbs#portable-functions) that can be used in all expressions. Those functions are passed through to the underlying database, allowing you to leverage the same functions for different databases, which greatly enhances portability.
+CAP supports a set of [portable functions](../guides/databases/cap-level-dbs#portable-functions) that can be used in all expressions. The CAP compiler automatically translates these functions to database-specific native equivalents, allowing you to use the same functions across different databases, which greatly enhances portability.
 
-## ref (path expression) { #ref }
+## ref — Element References & Path Expressions { #ref }
```
Akatuoro:
```suggestion
## ref - Path Expressions { #ref }
```

Currently it's cut off and uses a different hyphen:
Image

Akatuoro:
> > The cds repl - sql log pattern is only used once. Is it an inconsistency or on purpose?
> > image
> 
> `sql log` was on purpose in this case, as the code contains more than just an SQL statement but multiple lines of one transaction. But IMO we could also just use `SQL` to be in line with the rest of the code groups. wdyt @Akatuoro ?

We can also use SQL here. The retrieval is different, but in the end it is SQL.


> > Overall `cds repl` is only used twice in a codeblock. In all other cases it's `CQL`. Why is that?
> 
> it should always be `CQL` :)

`cds repl` is always used when we do not focus on the actual `CQL`.

The first time, we use the [Node.js fluent style CQL API](https://cap.cloud.sap/docs/node.js/cds-ql#constructing-queries) (as compared to [CQL](https://cap.cloud.sap/docs/cds/cql) or the [CAP Java fluent style CQL API](https://cap.cloud.sap/docs/java/working-with-cql/query-api)). This imho always gets confusing when we talk about "CQL", but sometimes we mean the actual language, other times we mean a fluent style API - they may support different things in different ways...... Here we explicitly use the Node.js CQL API because the actual CQL does not support `INSERT`.

The second time, we run `cds.parse.expr` which is not CQL.

PR #2408, opened 2026-02-18
fix menu title of Core Eventing in CAP

sjvans:


sjvans:
@renejeglinsky this should be in before we publish

PR #2410, opened 2026-02-23
fix(code-group): scrollbars appear in max window mode

joergmann:
### Symptom
code-group expands horizontally when window is maximized

### ❌ Before: code group expands horizontally
image

### ✅ After: code group shows scroll bar
image


joergmann:
Closing in favour of reformulate the content to fit the box or at least expand not too wide

PR #2428, opened 2026-03-03
Fix toggles

Akatuoro:
Fixes impl-variant toggle behavior in regards to:
- layout shift on load
- outline inconsistencies (both node and java headers shown on load)

Good page for testing: http://localhost:5173/docs/guides/databases/postgres#deploy
Chrome Performance CLS: 0.01 -> 0.00


renejeglinsky (approved):
Funny that I'm the codeowner in that area. We should fix that ;)

PR #2471, opened 2026-03-26
Update Authentication Documentation

samyuktaprabhu:
This PR updates the Authentication documentation to clarify the behavior of mock user authentication. 
- Updated documentation for `cds.security.authentication.mode` to reflect the new auto-derived behavior: when not explicitly configured, the mode defaults to `model-relaxed` in local development (mock users enabled) and `model-strict` in  production, aligning with secure-by-default principles
- Removed the tip suggesting manual configuration of `model-relaxed` for local development as this is now handled automatically


rjayasinghe (dismissed):

samyuktaprabhu (commented):
Hi @renejeglinsky,

Do you happen to know what caused the linting failure? I’m having trouble understanding what the error is referring to.

Thanks!

PR #2474, opened 2026-03-30
Document tenant-specific IAS host injection in RequestContext

StefanHenke:
Added instructions for injecting tenant-specific IAS host in background threads.

### DIFF
--- a/guides/security/cap-users.md
+++ b/guides/security/cap-users.md
@@ -1529,13 +1529,13 @@ Avoid iterating through all subscriber tenants to perform tenant-specific tasks.
 Instead, prefer a task-based approach which processes specific subscriber tenants selectively.
 :::
 
-When creating a new RequestContext in a background thread (without propagating an existing request), you must manually inject the tenant-specific IAS host to use IAS-based Remote Services. Unlike inherited request contexts that automatically carry authentication details, a fresh asynchronous request lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
+To use IAS-based Remote Services in background executions, you must in addition manually inject the tenant-specific IAS host into the created Request Context. Unlike inherited request contexts that automatically carry authentication details, a fresh Request Context lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
 
 ```java
 TenantInto tenantInfo = ...;
 String tenantHost = tenantInfo.get("subscriber").get("tenantHost");
 runtime.requestContext().systemUser(tenantId).modifyUser(user->user.setAdditionalAttribute("iss", tenantHost)).run((reqContext) -> {
-…
+… //call remote service
 });
 ```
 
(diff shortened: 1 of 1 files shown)


BraunMatthias (approved):

guides/security/cap-users.md
```diff
@@ -1529,6 +1529,16 @@ Avoid iterating through all subscriber tenants to perform tenant-specific tasks.
 Instead, prefer a task-based approach which processes specific subscriber tenants selectively.
 :::
 
+When creating a new RequestContext in a background thread (without propagating an existing request), you must manually inject the tenant-specific IAS host to use IAS-based Remote Services. Unlike inherited request contexts that automatically carry authentication details, a fresh asynchronous request lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
```
renejeglinsky: I assume "background thread" and "background executions" (from previous paragraph) mean the same thing.
If that is correct, I assume the difference between those two paragraphs is the target: persistence vs. remote service. Is that correct?

StefanHenke: @renejeglinsky : Yes you are right. I rephrased the sentence a bit to make it bit clearer.

guides/security/cap-users.md
```diff
@@ -1529,6 +1529,16 @@ Avoid iterating through all subscriber tenants to perform tenant-specific tasks.
 Instead, prefer a task-based approach which processes specific subscriber tenants selectively.
 :::
 
+To use IAS-based Remote Services in background executions, you must in addition manually inject the tenant-specific IAS host into the created Request Context. Unlike inherited request contexts that automatically carry authentication details, a fresh Request Context lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
```
renejeglinsky: I feel that this would be clearer (as long as it's correct).

```suggestion
To use IAS-based Remote Services in background executions, you must in addition manually inject the tenant-specific IAS host into the created Request Context. As background executions need a fresh Request Context, they cannot inherit request contexts that automatically carry authentication details and thereby lack the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
```

StefanHenke:
```suggestion
To use IAS-based Remote Services in background executions, you might in addition need to manually inject the tenant-specific IAS host into the created Request Context. If the background execution is initialized with a fresh Request Context, it will not carry inherited authentication details from a previous Request Context and thereby lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
```

@renejeglinsky: It is not given that every background execution needs a fresh request context. But this would be to detailed to be described here. I changed your proposal a bit. Hoping that this is now better understandable. if not, let´s maybe have a short call

renejeglinsky: Thanks @StefanHenke , I appreciate your proposal. Seems clearer to me.

renejeglinsky (approved):

PR #2479, opened 2026-03-31
Deprecations and removals in CAP Java 5

StefanHenke:
Added sections for adjusted, deprecated, and removed properties, as well as removed Java APIs in migration documentation.


rjayasinghe (approved):

java/migration.md
```diff
@@ -66,6 +67,39 @@ CAP Java 5.0 increased some minimum required versions:
...
+| --- | --- | --- | --- |
+| `abc` | false | true | Any description. |
+
+### Deprecated Properties
+
+The following properties have been deprecated and might be removed in a future major version:
+
+- `abd`
+
+The functionality provided by these properties is enabled by default and there is no reason to switch these off.
```
renejeglinsky: This sentence sounds  as if we ask the user to not switch off a deprecated property, or? Isn't the exact opposite the case?

StefanHenke: @renejeglinsky : I basically copied this sentence over from the previous migration guide. The meaning behind is: Typically, these deprecated properties can be used to toggle a newly introduced or incompatible feature. After some time, we enable these features by default and deprecate the corresponding property. This still gives applications in worst case (safety net) the possibility to switch off the features in the current release. This might give some time for adapting to the new behavior. However, once the property is deleted in the next major, it cannot be switched off anymore and applications have to live with the new feature behavior. That´s why the properties should not be switched off. Maybe only after our explicit recommendation for concrete stakeholders.

renejeglinsky: Ok, is there Is a significant difference between switching it off and not using it. Switching off would mean to not use the new default whereas not using it (that is, delete the property, right?) would mean using the default.
I haven't considered to use it to switch off default behavior.

```suggestion
The functionality provided by these properties is enabled by default. If you enabled and tested it already and want to use the default, you can safely remove the property. If you want to stay with the previous behavior und start testing the new default now, you can set it to `false`.
```

StefanHenke: @renejeglinsky: With new default (activated), stakeholders will automatically test the new behavior. Only if they face issues, they might get back to the previous behavior by setting it to `false` which we, however, strongly discourage. That´s why I wouldn´t mention this here.

What about:

```suggestion
The functionality provided by these properties is enabled by default. This reflects its intended behavior once the properties are deleted in future releases. 
```

StefanHenke: see https://github.com/capire/docs/pull/2479/changes#r3084352819

rjayasinghe (approved):

renejeglinsky (approved):

chgeo:
@agoerler @StefanHenke @rjayasinghe shall we merge this?

StefanHenke:
> @agoerler @StefanHenke @rjayasinghe shall we merge this?

I would vote for it. We will definitely add more things for the official 5.0 release. But it matches the current state

agoerler (approved):

PR #2504, opened 2026-04-15
Fixing outdated dependencies

danjoa:


swaldmann:
CI is currently failing as `eslint-plugin-vue` is removed, so I adjusted the ESLint config in https://github.com/capire/docs/pull/2504/commits/13c108631e7ee32592f03210544843347a37c31d.

package.json
```diff
@@ -23,24 +23,22 @@
...
+    "cspell": "^10",
     "express": "^5",
     "fflate": "^0.8.2",
     "globals": "^17.4.0",
     "markdownlint-cli": ">=0.35.0",
     "markdownlint-rule-search-replace": "^1.1.1",
     "monaco-editor": "^0.55.1",
     "sass": "^1.62.1",
-    "vite": "^7.3.1",
+    "vite": "^8",
```
Akatuoro: vite 8 is a bigger change since they use a new bundler (rolldown). It does not yet work with vite-plugin-cds, I've already been looking into adding support.
Should imho be moved to a separate PR.

As a side note: vitepress is also [still using vite 7](https://github.com/vuejs/vitepress/blob/9da1e3e70f41b7b8cb81f791307c778d65854f7a/package.json#L116).

chgeo: Yes.  I vote for keeping vite 7 for now and tackle vitepress 2-alpha with vite 8 later.

chgeo: We're back to vite 7.

chgeo (approved):

PR #2514, opened 2026-04-17
docs: node streaming API documentation

BobdenOs:
Currently the documentation includes all the APIs we support for streaming, but doesn't include information on how the javascript language enables these APIs to be consumed and provide the best outcomes by using streaming.


qmacro (commented):
I've made some changes and submitted a commit. Thanks.

node.js/streaming.md
```diff
@@ -0,0 +1,138 @@
...
+    row.descr = row.description
+    // ...
+  }
+}, INSERT.into(table))
+```
+
+As an alternative to this, you could simply define the transformation in your CDS model, as the `INSERT` and `UPSERT` queries are capable of doing the transformation directly in the database:
+
+```cds
+entity BooksRenamed projection on Books {
```
danjoa:
```suggestion
entity BooksRenamed as projection on Books {
```

danjoa (dismissed):

danjoa:
Should we include that into the release?

sjvans (approved):

sjvans:
> Should we include that into the release?

@danjoa  do you mean a release notes entry?

renejeglinsky (approved):

PR #2515, opened 2026-04-17
Fix (?) CXL constraint for no-description in constraints.md

qmacro:
In [the constraints section on @assert](https://cap.cloud.sap/docs/guides/services/constraints#assert-constraint), I couldn't understand what this was for:

```cds
Description @assert: (case
  when Description then 'Description must be specified'
  ...
);
```

As this code stands, it doesn't make much sense to me. I think perhaps the condition is meant to be `Description is null`.


renejeglinsky (approved):
This is then in line with what we have in xtravels: https://github.com/capire/xtravels/blob/e4a311614c62d5fea1795f99f62dd1329563d0e7/srv/travel-constraints.cds#L8

Thanks DJ!

PR #2518, opened 2026-04-21
Revert "Document tenant-specific IAS host injection in RequestContext…

StefanHenke:
As it turned out that the API provided by Subscription Manager does not offer the required field, we need to revert the docu. Once there is a final decision/solution, we will update the docu again

… (#2474)"

This reverts commit 66b451882b226922a59ddfc4c4862e1f314f8836.

### DIFF
--- a/guides/security/cap-users.md
+++ b/guides/security/cap-users.md
@@ -1529,16 +1529,6 @@ Avoid iterating through all subscriber tenants to perform tenant-specific tasks.
 Instead, prefer a task-based approach which processes specific subscriber tenants selectively.
 :::
 
-To use IAS-based Remote Services in background executions, you might in addition need to manually inject the tenant-specific IAS host into the created Request Context. If the background execution is initialized with a fresh Request Context, it will not carry inherited authentication details from a previous Request Context and thereby lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
-
-```java
-TenantInto tenantInfo = ...;
-String tenantHost = tenantInfo.get("subscriber").get("tenantHost");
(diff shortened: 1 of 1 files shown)


renejeglinsky (approved):

guides/security/cap-users.md
```diff
@@ -1529,16 +1529,6 @@ Avoid iterating through all subscriber tenants to perform tenant-specific tasks.
 Instead, prefer a task-based approach which processes specific subscriber tenants selectively.
 :::
 
-To use IAS-based Remote Services in background executions, you might in addition need to manually inject the tenant-specific IAS host into the created Request Context. If the background execution is initialized with a fresh Request Context, it will not carry inherited authentication details from a previous Request Context and thereby lacks the IAS host information required for remote service calls. The host can be retrieved from a `TenantInfo` object provided for example by the `TenantProviderService`.
-
-```java
-TenantInto tenantInfo = ...;
-String tenantHost = tenantInfo.get("subscriber").get("tenantHost");
```
BraunMatthias, line 1536: should we leave with tenantInfo = ...

StefanHenke: @BraunMatthias : You mean removing ´tenantInfo.get("subscriber").get("tenantHost");` and leaving a bit unclear where to get the tenantHost? I think it depends a bit if and when we will get an automatic solution which works out of the box. However, people will then ask where they can read the tenantHost from. We would need to mention the xsuaa api which the stakeholder has to call manually.

StefanHenke: @renejeglinsky : As Matthias is currently on vacation, I propose to get this revert commit merged as the shown code is not correct. We have meanwhile requested a more central solution from underlying layers which would make this completely obsolete anyways.

renejeglinsky (approved):

PR #2531, opened 2026-04-23
Remove sections for cds property changes in 4.9

StefanHenke:
Updated migration documentation with adjusted property defaults and removed properties.


chgeo (approved):
I have just replaced the list of old/new methods with a table, which reads nicer.

PR #2537, opened 2026-04-27
move plugin brownfield guide one level higher

rjayasinghe:


java/_menu.md
```diff
@@ -24,7 +24,7 @@
 # [Multitenancy](multitenancy)
 # [Security](security)
 # [Spring Boot Integration](spring-boot-integration)
-## [CAP plugins w/o CDS Model](cap-plugins-in-spring-boot-apps)
+# [BTP services in Brownfield Apps](cap-plugins-in-spring-boot-apps)
```
BraunMatthias: I'd prefer menus structure like this:

Building Plugins
Integrating CAP Apps
-> UCL
Integrating None-CAP Apps
Migration Guide

BraunMatthias (changes requested):

java/_menu.md
rjayasinghe: Great, now I remember what we discussed. 👍🏽

java/btp-services-in-non-cap-apps.md
```diff
@@ -4,7 +4,7 @@ synopsis: >
 status: released
 ---
 
-# Use CAP Plugins in Spring Boot Applications without a CDS Model
+# Integrate non-CAP Applications
```
smahati, line 7:
```suggestion
# Integrate non-CAP Applications
##### Use CAP Plugins in Spring Boot Applications without a CDS Model
```
If we use a fifth level header, it won't appear in the UI and it will preserve links to it and any bookmarks people might have.

PR #2539, opened 2026-04-27
Patch release 2026.4.1

renejeglinsky:
@chgeo I have no idea why there are so many commits. Can you explain that?


renejeglinsky:
I assume this is because at some point we used squash merge to merge into cap.cloud.sap.
Will merge this.

PR #2550, opened 2026-05-04
Troubleshooting empty HDI if subscribed on Windows

vkozyura:


get-started/get-help.md
```diff
@@ -283,6 +283,18 @@ npm add -D tar
 ```
 On macOS and Linux, the built-in implementation continues to be used.
 
+### Why are tenant HDI containers empty after subscription on Windows?
```
renejeglinsky: This is only valid for hybrid testing, right?

vkozyura: yes, in production you usually sibscribe in CF cockpit

stewsk: what difference does this make? If you have built your app on windows, then the resources.tgz is missing (or empty), and then nothing gets deployed to the HDI container, no  matter how you subscribe.
So I think this applies also to production.
Or am I wrong here?

stewsk: What I mean: it doesn't matter how you subscribe, but how the app is built

vkozyura: @stewsk: Agree, then we should change from "troubleshooting empty HDI" - just copied it from the issue :) - to "issues building MTX resources". And for the last one we already have troubleshooting item above - I will modify it.

ecklie (approved):
Yes, the problem is caused by the build. It also has impact on the productive app. We need to check if we can improve the build and add a consistency check.

get-started/get-help.md
ecklie: But the visible impact is on the deployed application

vkozyura: @ecklie: In some cases we obtain an explicite error "tar: Error is not recoverable" by cds build. In some other cases cds build may silently fail, which can be seen only on the deployed application. My question for now: In troubleshooting guide - should it be one or two sections?

renejeglinsky: For the error message we do have the troubleshooting info already. So this can be found. How would we describe the other case? How would one notice on the deployed application if the build doesn't throw an error?

vkozyura: > How would we describe the other case? How would one notice on the deployed application if the build doesn't throw an error?

This is exactly what the newly added section does: "Why are tenant HDI containers empty after building app on Windows?"

My question was if we should merge those two section?

get-started/get-help.md
```diff
@@ -283,6 +283,18 @@ npm add -D tar
 ```
 On macOS and Linux, the built-in implementation continues to be used.
 
+### Why are tenant HDI containers empty after building app on Windows?
```
renejeglinsky: I might be wrong or simplify it too much, but to me this reads as if there's is already a first analysis done and I found that HDI containers are empty.
What is it that makes me investigate? Like "Can't get/read data from customer tenant"?. So folks who have not done an initial analysis and found the empty containers, what would they look for in our troubleshooting?

vkozyura: Here the ticket describibg the issue: https://github.tools.sap/cap/issues/issues/19999.

stewsk: in my case the "symptom" was that I got funny runtime errors from the Node.js runtime saying that table xyz is not there (don't remember the precise error message)

ecklie: Right. It is not after building an app but after subscription.
"Why are tenant HDI containers empty after tenant subscription"
It think the error message is something like "table not found". I will check it.

ecklie: SqlError: invalid table name:  Could not find table/view ...

get-started/get-help.md
```diff
@@ -283,6 +283,18 @@ npm add -D tar
...
+On Windows, cds build --production may silently fail to create resources.tgz in the MTX sidecar build output. 
+The build log will incorrectly report the file as written. 
+
+Install the tar library to fix this:
+
+```sh
+npm add -D tar
+```
+On macOS and Linux, the built-in implementation continues to be used.
+
```
renejeglinsky: I'd also suggest to move this entry to the MTXS section of this guide. WDYT?

```suggestion
### Why do I get "SqlError: invalid table name" after subscription?

Most likely you're using Windows to build your project. On Windows there's a known issue, where `cds build --production` may silently fail to create the _resources.tgz_ in the MTX sidecar build output. After deployment and subscription, you can then notice the mentioned SqlError or similar error messages that point to tables/views not being available.

:::warning The build log will incorrectly report the file as written. 
:::

To fix this on Windows, install the tar library:

```sh
npm add -D tar

Even with this dependency added, on macOS and Linux the built-in implementation continues to be used.

```

renejeglinsky: Image

stewsk: The error message I got (cds watch on node RT) was
"Could not find table/view ...... in schema ......:

get-started/get-help.md
vkozyura: ok, two different errors -> two sections

get-started/get-help.md
vkozyura: @renejeglinsky: i changed the title. In your suggestion I didn't like "Most likely you're using Windows" - using Windows is rather an exceptional situation. i merged your changes from the suggestion.

renejeglinsky: @vkozyura If I understand the issue correctly, Windows is the "most likely" case when someone encounters this error, or? There might be other situations where a similar error occurs, but this is then not OS-specific. Probably.  We don't know for sure. And this is what I wanted to express.

get-started/get-help.md
```diff
@@ -283,6 +283,23 @@ npm add -D tar
 ```
 On macOS and Linux, the built-in implementation continues to be used.
 
+### How to fix "SqlError: invalid table name: Could not find table/view ..." while building app on Windows
+
+On Windows, cds build --production may silently fail to create resources.tgz in the MTX sidecar build output.
```
renejeglinsky:
```suggestion
### How to fix "SqlError: invalid table name: Could not find table/view ..."?

On Windows there's a known issue, where `cds build --production` may silently fail to create the _resources.tgz_ in the MTX sidecar build output.
```

get-started/get-help.md
```diff
@@ -283,6 +283,23 @@ npm add -D tar
 ```
 On macOS and Linux, the built-in implementation continues to be used.
 
+### How to fix "SqlError: invalid table name: Could not find table/view ..." while building app on Windows
+
+On Windows, cds build --production may silently fail to create resources.tgz in the MTX sidecar build output. 
+The build log will incorrectly report the file as written.
```
renejeglinsky: Duplicate of the following warning.

```suggestion
```

renejeglinsky (approved):

PR #2559, opened 2026-05-07
Discourage `i18n.folders` config

daogrady:
Discourage users from changing the `cds.i18n.folders` configuration, which lead to frequent confusion among users and often did not match their expectations/ lead to bugs.


joergmann (approved):

node.js/cds-i18n.md
```diff
@@ -657,7 +657,7 @@ cds.env.i18n = {
 
 ::: danger
```
renejeglinsky: Maybe we could also emphasize this in the note title?

```suggestion
::: danger Do not switch defaults without proper evaluation
```


```suggestion
::: tip Stick to the defaults!
```

chgeo (approved):

smahati (approved):

PR #2569, opened 2026-05-11
fix title ref in hyperlink

qmacro:
Looks like the heading title changed, but this instance of a reference to it ([Instance-based Access Control](https://cap.cloud.sap/docs/guides/security/authorization#instance-based-auth)) wasn't updated.


qmacro:
I'll close this PR and include the change in the existing PR I have open.

PR #2580, opened 2026-05-18
Add `httpc` language alias for php

renejeglinsky:
This is how it looks like:

image


chgeo (approved):

chgeo:
@renejeglinsky I decided to use `httpc` as alias name, similar to `csvc` (CSV with comments).

renejeglinsky:
> @renejeglinsky I decided to use `httpc` as alias name, similar to `csvc` (CSV with comments).

Makes absolute sense. Thanks! :)

PR #2585, opened 2026-05-20
clarify @mandatory is not validated in deep updates

mariayord:
Updated documentation regarding @mandatory annotation. There is no validation on @mandatory in deep updates usecase. 

Removed section with `@FieldControl`, as it is mentioned already in `Field Control `below

Fixes: https://github.tools.sap/cap/cdsnode/issues/2637


guides/services/constraints.md
```diff
@@ -368,13 +368,9 @@ service Sue {
...
-In addition to server-side input validation as introduced above, this adds a corresponding `@FieldControl` annotation to the EDMX so that OData / Fiori clients would enforce a valid entry, thereby avoiding unnecessary request roundtrips:
-
-```xml
-<Annotations Target="Sue.Books/title">
-  <Annotation Term="Common.FieldControl" EnumMember="Common.FieldControlType/Mandatory"/>
-</Annotations>
-```
+::: warning Deep updates not covered
+`@mandatory` validation is **not** applied to nested data in deep update operations. To enforce mandatory values in those cases, use the [`not null` constraint](../../cds/cdl#constraints) instead.
+:::
```
renejeglinsky, line 373:
```suggestion
:::

[Learn more about `@Fieldcontrol` annotation with `@mandatory`.](#mandatory-1){.learn-more}
```

renejeglinsky:
@johannes-vogel @agoerler could you have a quick check here?

renejeglinsky:
@rjayasinghe Have you reviewed this PR?

rjayasinghe:
**The changes in this PR reflect the Node.js runtime — they do not hold for CAP Java.**

The statement *"`@mandatory` validation is **not** applied to nested data in deep update operations"* is accurate for Node.js, but as written it reads as cross-runtime and is misleading for Java. Since these docs cover both runtimes, could we scope it per-runtime?

In **CAP Java**, `@mandatory` validation (`MandatoryHandler` → `NotNullValidator`) *does* recurse into nested data on deep updates, following the same cascade rules as persistence — i.e. into **compositions** by default (and any association carrying `@cascade.update`/`@cascade.all`). The only relaxation is partial-update semantics. Concretely, for a deep update:

| Nested `@mandatory` field | CAP Java behavior |
|---|---|
| **omitted** from the payload | tolerated (partial-update semantics) |
| explicit `null` / `""` in a **composition** | **rejected** (`VALUE_REQUIRED`) |
| explicit `null` / `""` in an **association** target | tolerated |

So for Java it's not *"no validation on nested data"* — it's *"a **missing** mandatory value is not enforced on update, but an explicitly provided `null` inside a cascading composition still is."*

This is backed by the unit tests in `MandatoryHandlerTest` (the composition-null case is asserted to fail; I've added coverage for the omitted-field case confirming it's tolerated).

Suggested framing: keep the current sentence under a Node.js tab, and for Java state that mandatory *is* validated in nested cascading targets, with only a *missing* value left unenforced (use `not null` if you also need presence enforced).

mariayord:
@rjayasinghe As I understand it, Java does not validate @mandatory, except for compositions where the value is explicitly set to null or ''.
@renejeglinsky In that case, I will document this as the Node.js behavior.
@rjayasinghe  For Java, it's up to you whether you want to document this partially supported feature.

guides/services/constraints.md
```diff
@@ -368,13 +368,11 @@ service Sue {
 }
 ```
 
-In addition to server-side input validation as introduced above, this adds a corresponding `@FieldControl` annotation to the EDMX so that OData / Fiori clients would enforce a valid entry, thereby avoiding unnecessary request roundtrips:
+::: warning Deep updates not covered
```
renejeglinsky: Would that be enough to merge this PR?

```suggestion
::: warning Deep updates not covered in Node.js
```

mariayord: @renejeglinsky It looks like Node.js also supports the same use case as the Java runtime, so we can provide common documentation. Could you please check if the changed docu is ok ?

guides/services/constraints.md
```diff
@@ -368,13 +368,14 @@ service Sue {
 }
 ```
 
-In addition to server-side input validation as introduced above, this adds a corresponding `@FieldControl` annotation to the EDMX so that OData / Fiori clients would enforce a valid entry, thereby avoiding unnecessary request roundtrips:
+::: warning Mandatory Values in Compositions
+For insert/update, a mandatory value is **not** enforced when the property is missing from the request.
+However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions.
+For associations, no mandatory-value validation is performed.
+To enforce mandatory values in those cases, use the [`not null` constraint](../../cds/cdl#constraints) instead.
+:::
```
renejeglinsky: Just an idea, let me know what you think.


```suggestion
#### ... in Compositions
For insert/update, a mandatory value is **not** enforced when the property is missing from the request.
However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions.

#### ... in Associations
No mandatory-value validation is performed. To enforce mandatory values in associations, use the [`not null` constraint](../../cds/cdl#constraints) instead.

```

Does that mean, in associations when the `not null` constraint is active, if a request is missing a property it will fail? Or where is the difference in behavior between associations and compositions when following this guidance?

mariayord: If the payload does not contain the mandatory value at all, in either a composition or an association, the validation is not performed. In this case, `not null` should be used to enforce validation.

renejeglinsky: Would that be correct? Or how is it with a provided `null/''` value in associations?

```suggestion
#### ... in Compositions
For insert/update, a mandatory value is **not** enforced when the property is missing from the request.
However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.

#### ... in Associations
For insert/update, a mandatory value is **not** enforced when the property is missing from the request. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
```

mariayord: Validation is done only in case of compositions and provided value `null/''`

renejeglinsky: So my proposal is correct, or?

mariayord: Not exactly , because in Association section is not clear if the validation is done by `null/''`

renejeglinsky: Like this?

```suggestion
#### ... in Compositions
For insert/update, a mandatory value is **not** enforced when the property is missing from the request.
However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
#### ... in Associations
For insert/update, a mandatory value is **not** enforced when the property is missing from the request. Even an explicitly provided `null/''` value is accepted. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
```

mariayord: @renejeglinsky , I am not sure if it is better to repeat the same statement in two sections, or just to point what is valid for both and what is the difference. But if you think such format is better for capire, I am OK.

renejeglinsky: These days being explicit is better -> AI ;) 
We assume a lot of knowledge in our suers already in capire but here I feel being explicit (null is accepted) is better.
Thanks!

guides/services/constraints.md
```diff
@@ -368,13 +368,13 @@ service Sue {
 }
 ```
 
-In addition to server-side input validation as introduced above, this adds a corresponding `@FieldControl` annotation to the EDMX so that OData / Fiori clients would enforce a valid entry, thereby avoiding unnecessary request roundtrips:
+#### ... in Compositions
+For insert/update, a mandatory value is **not** enforced when the property is missing from the request.
+However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
+#### ... in Associations
+For insert/update, a mandatory value is **not** enforced when the property is missing from the request. Even an explicitly provided `null/''` value is accepted. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
```
renejeglinsky, line 377: Small correction-

```suggestion
However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.

#### ... in Associations

For insert/update, a mandatory value is **not** enforced when the property is missing from the request. Even an explicitly provided `null/''` value is accepted. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
```

renejeglinsky (approved):

renejeglinsky:
@rjayasinghe Could you check/approve?

guides/services/constraints.md
```diff
@@ -368,13 +368,15 @@ service Sue {
...
+#### ... in Compositions
+For insert/update, a mandatory value is **not** enforced when the property is missing from the request.
+However, an explicitly provided `null/''` value is validated and rejected. This rule applies to compositions and cascading compositions. To enforce mandatory values, use the [`not null` constraint](../../cds/cdl#constraints) instead.
 
-```xml
-<Annotations Target="Sue.Books/title">
-  <Annotation Term="Common.FieldControl" EnumMember="Common.FieldControlType/Mandatory"/>
-</Annotations>
-```
+#### ... in Associations
```
johannes-vogel, line 375: I don't think we need to repeat the section for associations. In associations, only the foreign key is respected. all other elements are not validated, it's not in particular specific for mandatory.

rjayasinghe (approved):

rjayasinghe:
> @rjayasinghe Could you check/approve?

LGTM

guides/services/constraints.md
renejeglinsky: In a previous version I had the impression that there was a notable change for associations:
image

If we'd remove the whole section about associations we'd loose the information about the `not null` constraint. Or should we move that somewhere else?

renejeglinsky: Should this sentence, for example, be moved to the general section?

PR #2592, opened 2026-05-26
Review Serving UIs > Fiori

renejeglinsky:


danjoa (approved):
👍 Thanks @renejeglinsky !

Note: [I reverted the prior change of `php` to `httpc` back to `php`](https://github.com/capire/docs/pull/2592/commits/b8ccaacbda74cc76e785c9b60183a53c3f9f719a) (for a reason, let's keep it that way pls)

chgeo (approved):

PR #2595, opened 2026-05-27
`<div id>` instead of h6 title

danjoa:
Since its latest release VSCode's built-in link checker supports `<div id=... />` tags for alternative link targets, which we should preferably use instead of h6 titles.

### DIFF
--- a/guides/uis/fiori.md
+++ b/guides/uis/fiori.md
@@ -241,7 +241,7 @@ With that, all UIs on all services exposing `Books` will automatically receive V
 
 
 ## Fiori Draft Support
-###### Draft Support
+<div id="draft-support" />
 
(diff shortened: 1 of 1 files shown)


chgeo (approved):

renejeglinsky:
Does our BLC recognize this?

chgeo:
> Does our BLC recognize this?

It should.  Just give it a try.

danjoa:
> > Does our BLC recognize this?
> 
> It should. Just give it a try.

Yes, I ran it before pushing this

danjoa:
@renejeglinsky waiting for your additional approval as code owner

renejeglinsky (approved):

PR #2596, opened 2026-05-28
Add Node.js part to Outbound Authentication

vkozyura:


guides/security/remote-authentication.md
```diff
@@ -105,26 +109,64 @@ cds:
...
 
+::: details Java configuration explained
+
 The `type` property activates the protocol for exchanging business data and must be offered by the provider [CDS service](https://github.com/capire/xflights-java/blob/6fc7c665c63bb6d73e28c11b391b1ba965b8772c/srv/data-service.cds#L24).
 The `model` property needs to match the fully qualified name of the CDS service from the imported model.
 You can find CDS service definition of `sap.capire.flights.data` in file `target/cds/capire/xflight-data/service.cds` resolved during CDS build step.
 The `binding.name` needs to point to the shared identity instance and `options.url` together with `http.suffix` provides the required location of the remote service endpoint.
 Finally, `onBehalfOf: systemUser` specifies that the remote call is invoked on behalf of a technical user in context of the tenant.
 
+:::
```
PDT42, line 139: Move this block to right after the Java specific example?

guides/security/remote-authentication.md
```diff
@@ -105,26 +109,64 @@ cds:
...
+
+::: details Node.js configuration explained
+
+The `kind` property activates the protocol for exchanging business data (same as `type` in Java).
+The key in `cds.requires` must match the fully qualified CDS service name (e.g., `sap.capire.flights.data`), which is the same name used in `cds.connect.to()`.
+The `credentials.url` provides the full endpoint URL including the service path `/hcql/data`.
+The `forwardAuthToken: true` enables direct forwarding of the incoming JWT token to the provider service - this is suitable for co-located services that share the same identity instance, as the token is already valid for the provider.
+
+:::
```
PDT42: IMO, this could / should point to: https://cap.cloud.sap/docs/guides/integration/reuse-and-compose#configuring-required-services and https://cap.cloud.sap/docs/guides/integration/reuse-and-compose#binding-required-services, instead.

guides/security/remote-authentication.md
```diff
@@ -105,26 +109,64 @@ cds:
...
+
 ::: tip
-On behalf of `systemUser` works both in pure single tenant and in pure multitenant scenarios.
+On behalf of `systemUser` (Java) works both in pure single tenant and in pure multitenant scenarios.
```
PDT42: In the updated structure, this jump back to Java `systemUser` is very disconnected. What is the / Is there an equivalent in Node.js?

guides/security/remote-authentication.md
```diff
@@ -150,11 +192,12 @@ For different [user propagation](./cap-users#remote-services) modes the remote s
 The provider service authorization needs to align with the configured user propagation.
 :::
 
-Additionally, to establish the co-located setup, the microservice needs to share the same identity instance:
+Additionally, to establish the co-located setup, the microservice needs to share the same identity instance.
+This is configured in the MTA deployment descriptor (applies to both Java and Node.js):
```
PDT42: "applied to both Java and Node.js" should be the default for any shared guide. IMO, only the opposite should be noted.

guides/security/remote-authentication.md
```diff
@@ -150,11 +192,12 @@ For different [user propagation](./cap-users#remote-services) modes the remote s
...
 
 ::: code-group
 
-```yaml [/srv/srv/main/resources/application.yaml]
+```yaml [mta.yaml]
```
PDT42, line 197: `mta.yaml` makes sense to me, reading `resources` .. but his change still looks suspicious.

guides/security/remote-authentication.md
```diff
@@ -172,11 +215,18 @@ resources:
...
 cd ./xflights_java
 cds up
 ```
 
+```sh [Node.js]
+cd ./xflights
+cds up
+```
+:::
+
```
PDT42, line 226: Do we really need to tell people how to deploy multiple times?

guides/security/remote-authentication.md
```diff
@@ -220,7 +270,7 @@ As a consequence, external services can run cross-regionally; even non-BTP syste
 A prerequisite for external service calls is a trust federation between the consumer and the provider system.
 
 A seamless integration experience for external service communication is provided by [IAS App-2-App](#app-to-app) flows, which are offered by CAP via remote services.
-Alternatively, remote services can be configured on top of [BTP HTTP Destinations](../services/consuming-services#using-destinations) that offer [various authentication strategies](https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/http-destinations) such as SAML 2.0 as required by many S/4 system endpoints.
+[BTP HTTP Destinations](../services/consuming-services#using-destinations) offer [various authentication strategies](https://help.sap.com/docs/connectivity/sap-btp-connectivity-cf/http-destinations) such as SAML 2.0 as required by many S/4 system endpoints. Both CAP Java and CAP Node.js support IAS App-2-App via configuration - Java uses service bindings with `ias-dependency-name`, while Node.js uses BTP Destinations with `tokenService.body.resource`.
```
PDT42: Using "BTP HTTP Destinations" is implicitly miss-leading, IMO. The config options apply to all the destination types, I think ... no need to make people question: What's the difference between HTTP and non-HTTP destinations - if there is none.

PDT42: "Both CAP Java and CAP Node.js support IAS App-2-App via configuration - Java uses service bindings with `ias-dependency-name`, while Node.js uses BTP Destinations with `tokenService.body.resource`." ... to do what?

guides/security/remote-authentication.md
```diff
@@ -235,11 +285,13 @@ Technically, the connectivity component uses [IAS App-2-App flows](https://help.
...
 
 :::tip
-CAP offers a simplified App-2-App setup by leveraging remote services that require:
+CAP offers App-2-App setup by leveraging remote services that require:
 - Identity instances for provider and consumer
 - Configured IAS dependency from consumer to provider
 - URL pointing to the provider
 - Principal propagation mode (optional)
+
+**Note:** Both CAP Java and CAP Node.js handle IAS App-2-App via configuration.
```
PDT42: Seems somewhat redundant.

guides/security/remote-authentication.md
```diff
@@ -261,15 +313,37 @@ resources:
...
+      service-plan: application
       config:
         display-name: xflights
-        provided-apis: # [!code ++:5]
+        oauth2-configuration:
+          token-policy:
+            access-token-format: jwt
+        provided-apis:
         - name: data-consumer
           description: Grants technical access to data service API
```
PDT42: Will this render as valid yaml?

guides/security/remote-authentication.md
```diff
@@ -333,22 +414,125 @@ cds:
           ias-dependency-name: data-consumer
 ```
 
+```json [Node.js: package.json]
```
PDT42, line 412: The process for Node.js is WAY more elaborate than the Java one, but doesn't get its own introduction ... "You only need to add the information about the IAS dependency to be called." definitely doesn't sound accurate.

guides/security/remote-authentication.md
```diff
@@ -333,22 +414,125 @@ cds:
...
+| Token Service URL | `https://<IAS tenant>/oauth2/token` |
+| Token Service URL Type | `Dedicated` |
+
+**Additional Property (required for IAS App-2-App):**
+
+| Property | Value |
+|----------|-------|
+| `tokenService.body.resource` | `urn:sap:identity:application:provider:name:data-consumer` |
+
+The key is `tokenService.body.resource` which passes the `resource` parameter to IAS, triggering App-2-App token scoping and adding the `ias_apis` claim.
```
PDT42, line 482: Why not tell people how they can specify the destination in their `mta` instead? 

EDIT: Justified in `**Note:** MTA cannot resolve cross-service credential references like `${generated>xtravels-ias/clientid}`. The destination must be created manually in BTP Cockpit or via Destination Service API.` - later on. Is that note accurate?

guides/security/remote-authentication.md
```diff
@@ -333,22 +414,125 @@ cds:
...
+
+**3. Authentication type depends on use case:**
+- **OAuth2ClientCredentials**: For technical user scenarios (no user context)
+- **OAuth2JWTBearer**: For user propagation (requires user token from authorization_code flow)
+
+Both support `tokenService.body.resource` for IAS App-2-App scoping.
```
PDT42: Why are these two listed, but none of the other auth types?

guides/security/remote-authentication.md
```diff
@@ -333,22 +414,125 @@ cds:
...
+
+| Property | Value |
+|----------|-------|
+| Name | `xflights-ias-app2app` |
+| Type | `HTTP` |
+| URL | `https://<xflights-srv url>` |
+| Proxy Type | `Internet` |
+| Authentication | `OAuth2ClientCredentials` or `OAuth2JWTBearer` |
+| Client ID | Consumer IAS client ID |
+| Client Secret | Consumer IAS client secret |
```
PDT42, line 466: Where do people get these?

guides/security/remote-authentication.md
```diff
@@ -361,7 +545,7 @@ Open the Administrative Console for the IAS tenant (see prerequisites [here](./a
 
 1. Select **Applications & Resources** > **Applications**. Choose the IAS application of the `xtravels` consumer from the list.
 2. In **Application APIs** select **Dependencies** and click on **Add**.
-3. Type `data-consumer` as dependency name (needs to match property value `ias-dependency-name`) and pick provided API `data-consumer` from the provider IAS application `xflights`.
+3. Type `data-consumer` as dependency name (needs to match property value `ias-dependency-name` in Java, or the dependency name in `tokenService.body.resource` for Node.js) and pick provided API `data-consumer` from the provider IAS application `xflights`.
```
PDT42: "... in Java, or the dependency name in `tokenService.body.resource` for Node.js" - what is the "dependency name"?

guides/security/remote-authentication.md
```diff
@@ -392,11 +576,11 @@ To do so, assign a proper AMS policy (e.g., `admin`) to the test user as describ
...
 
-- **Don't write custom integration logic** for consumed services.
-Leverage CAP's remote service architecture instead to ensure a seamless integration experience.
+- **Don't write custom integration logic** for consumed services unless necessary.
+Leverage CAP's remote service architecture when possible to ensure a seamless integration experience.
 
-- **Don't implement connectivity layer code** (for example, to fetch or exchange tokens).
-Instead, rely on the shared connectivity component, which ensures centralized and generic processing of outbound requests.
+- **Don't implement connectivity layer code** (for example, to fetch or exchange tokens) unless necessary.
+Rely on the shared connectivity component, which ensures centralized and generic processing of outbound requests.
```
PDT42: IMO these changes just weaken the "Pitfalls" section.

PDT42 (changes requested):
Perhaps this guide should use the Java / Node.js switch, if that is still a thing? The examples and required steps seem pretty disconnected.

renejeglinsky:
> Perhaps this guide should use the Java / Node.js switch, if that is still a thing? 

It's not a thing anymore. But code groups that have Java/Node are now remembered by selection without the toggle.

guides/security/remote-authentication.md
vkozyura: The previous label [/srv/srv/main/resources/application.yaml] was wrong - that path is for Java Spring config, not MTA

guides/security/remote-authentication.md
vkozyura: 1. Java did it 2. makes it obvious which app is being deployed -> keep

guides/security/remote-authentication.md
vkozyura: the note is accurate and based on real experience

guides/security/remote-authentication.md
vkozyura: changed

guides/security/remote-authentication.md
vkozyura: info added

guides/security/remote-authentication.md
PDT42: I don't disagree in principle: But this code-block is part of this document four separate times ...

vkozyura: but for two different projects: xflights and xtravels :) anyway I don't want to change the java part in this pull request

guides/security/remote-authentication.md
PDT42: Conceptually, should the destination be using the creds from the service binding or a dedicated service key?

guides/security/remote-authentication.md
```diff
@@ -333,22 +406,127 @@ cds:
...
+            access-token-format: jwt
+
+  - name: xtravels-destination
+    type: org.cloudfoundry.managed-service
+    parameters:
+      service: destination
+      service-plan: lite
+```
+
+**Note:** MTA cannot resolve cross-service credential references like `${generated>xtravels-ias/clientid}`. The destination must be created manually in BTP Cockpit or via Destination Service API.
```
PDT42: Perhaps: Move this to the beginning of "2. BTP Destination" or right before "Client ID and Client Secret are obtained from the consumer's ..."? Provide the reason for what should be done, before doing it?

PDT42 (dismissed):
I am still having trouble with the structure: I think intermixing the Java and Node.js guides like this doesn't provide much benefit and is at least a bit confusing  ... However, the documented process (without trying to go through it myself) makes sense to me.

guides/security/remote-authentication.md
vkozyura: moved

guides/security/remote-authentication.md
vkozyura: done

guides/security/remote-authentication.md
```diff
@@ -105,36 +105,73 @@ cds:
...
 cds up
 ```
 
+```sh [Node.js]
+cd ./xtravels
+cds up
+```
+:::
+
 ❗Note that CF application `xtravels-srv` will not start successfully as long as `xflights` is not deployed yet (step 3).
```
renejeglinsky: This is how this part now looks like.

Image

WDYT? @PDT42 @vkozyura

renejeglinsky: We could/should do that throughout the document. I think this makes a nicer flow

guides/security/remote-authentication.md
renejeglinsky, line 163: What's the result of this? If we state this, I think it's clearer what we talk about in the rest of the document:
- `xtravels-srv` server
- `xtravels-ias` service instance
- `xtravels-db` service instance?
- more?

guides/security/remote-authentication.md
```diff
@@ -90,6 +87,7 @@ Make sure that you've prepared a [local environment for CF deployments](../deplo
 
 As client, `xtravels-srv` first needs a valid configuration for the remote service `sap.capire.flights.data`:
```
renejeglinsky: I'd argue that this cannot be called `xtravels-srv` yet. We should only use it when it is important to differentiate between xtravels and the deployed server.

guides/security/remote-authentication.md
renejeglinsky: I just included it in my changes. Please have a look if you have any objections against one/some of my edits.

vkozyura: @renejeglinsky: the changes are fine for me

guides/security/remote-authentication.md
vkozyura: @renejeglinsky: i was unsure if we should include cds up here. There exists a detailed deployment guide. Finally i decided to keep cds up, but I would not go deeper into the deployment and the deployed artifacts. Maybe link deployment guide once? - leave it up to you.

guides/security/remote-authentication.md
vkozyura: @renejeglinsky: agree! just like "deployed `xtravels` app first needs a valid configuration ..." ?

guides/security/remote-authentication.md
PDT42: IMO repeating cds up for each step is somewhat redundant ... People should know how to deploy at this point - and if they don't they should go back to learn about it specifically - not be able to copy&paste from here and "stay ignorant" 🙈

renejeglinsky: The guide as it is written has deployment as an integral part. So I think it would feel unfinished if we don't add those small boxes there. While I agree that those boxes occur very often in that guide, I think we'd put a burden on the consumer, that very much influences the experience in a negative way, if we just say: Deploy now and see the result. If you know how to do it, you'll skip the boxes and I don't think it's annoying. If you don't know it by heart and we make you switch guides, that's a bad experience.
Would it only be optional to deploy, that's a different story.

vkozyura: @renejeglinsky: The docs are now fine for me. Feel free to merge it.

PDT42 (approved):
IMO - when both of you are happy with the result: 🚀?

renejeglinsky (approved):

PR #2597, opened 2026-05-28
Remove vector dimension from examples

MattSchur:
- `Vector` (`length`) -> `Vector` (`dimension`)
- Remove vector dimension from examples


stewsk (dismissed):

guides/databases/vector-embeddings.md
```diff
@@ -12,11 +12,11 @@ Choose an embedding model that fits your use case and data (for example English
 Use the [SAP Generative AI Hub](https://www.sap.com/products/artificial-intelligence/generative-ai-hub.html) for unified consumption of embedding models and LLMs across different vendors and open-source models. Check for available models on the [SAP AI Launchpad](https://help.sap.com/docs/ai-launchpad/sap-ai-launchpad-user-guide/models-and-scenarios-in-generative-ai-hub-fef463b24bff4f44a33e98bb1e4f3148#models).
 
 ## Add Embeddings to Your CDS Model
-Use the built-in CDL [Vector type](../../cds/types) in your CDS model to store embeddings. Set the vector dimensions to match the embedding model (for example, 768 for *SAP_GXY.20250407*).
+Use the built-in CDL [Vector type](../../cds/types) to store embeddings. Use `Vector` without specifying a dimension to simplify changing the embedding model, or make sure the vector dimension matches the embedding model (for example, 768 for *SAP_GXY.20250407*).
```
renejeglinsky: From the sample later on, the vector dimension (prev. length) has been removed. According to this sentence,  the dimension should be somewhere in it, or?

MattSchur: The dimension is not necessary and removing it makes it easier to change the embedding model in a later step to one providing vectors with a different dimension.

renejeglinsky: Ok. And how can I make sure that the "vector dimension matches the embedding model"?

MattSchur: that's only necessary if a vector dimension is specified (or).
```suggestion
Use the built-in CDL [Vector type](../../cds/types) to store embeddings. Use `Vector` without specifying a dimension to simplify changing the embedding model. If you specify a vector dimension, make sure it matches the embedding model (for example, 768 for *SAP_GXY.20250407*).
```

renejeglinsky: Thanks. I really haven't understood/read that before. So if I specify it, it would be clear how to do it, right? I guess like this:

`embedding : Vector(768) = vector_embedding(`

MattSchur: Yes, that should be clear from the type definition

renejeglinsky (dismissed):

PR #2604, opened 2026-06-01
CDS to DDL: clarify treatment of expressions

stewsk:


guides/databases/cdl-to-ddl.md
```diff
@@ -60,6 +60,11 @@ code --diff _out/c/sqlite.sql _out/c/h2.sql
 > [!tip] CDS models are database-agnostic
 > CDS models are designed to be database-agnostic, allowing you to switch between different databases with minimal changes. The `--dialect` option helps you see how your models translate to different database-specific DDLs. \
 
+Note that expressions in CDS models are _not_ treated in a database-agnostic way. In other words, they are not rewritten
+into database-specific equivalents; instead, they are carried over one-to-one into the DDL.
+You therefore need to ensure that any expression you write in your CDS model works across all databases you use.
+See also [Database-Specific Models](#database-specific-models).
```
renejeglinsky:
```suggestion
Expressions in CDS models are _not_ treated in a database-agnostic way. The compiler does not rewrite them into database-specific equivalents. Instead, it carries them over directly into the DDL. Ensure that any expression you write in your CDS model works across all databases you use. For more information, see [Database-Specific Models](#database-specific-models).
```

guides/databases/cdl-to-ddl.md
```diff
@@ -60,6 +60,8 @@ code --diff _out/c/sqlite.sql _out/c/h2.sql
 > [!tip] CDS models are database-agnostic
 > CDS models are designed to be database-agnostic, allowing you to switch between different databases with minimal changes. The `--dialect` option helps you see how your models translate to different database-specific DDLs. \
```
danjoa:
```suggestion
> CDS models are designed to be database-agnostic, allowing you to switch between different databases with minimal changes. The `--dialect` option helps you see how your models translate to different database-specific DDLs. 
```

danjoa (changes requested):
Why did we have to add this? → should make it more specific as stated generically like that it is also wrong. In particular: Expressions in `@assert` annotations are database-agnostic, which is done by the respective expression evaluator. Here: the CAP runtimes. 
In general: expressions in annotations are rarely evaluated / executed by the compiler. So stating expressions are not agnostic from a compiler-only perspective is misleading.

stewsk:
> Why did we have to add this?

because the statement "CDS models are designed to be database-agnostic, " by itself is also misleading. With respect to expressions that makes the impression that one could write any expression and CAP ensures that it translated correctly to the respective database - like in ABAP. This is not true for the compiler, where all expressions are sent to the DB one-to-one, and even the runtimes only provide this service for certain expressions.

(reason for this addendum: there are cap/issues rooted in the expectation that one can just write any expression and that would work)

renejeglinsky:
@danjoa @stewsk If this should be included in the release please let me know.

PR #2606, opened 2026-06-01
Document only one value help annotation shortcut

stewsk:
Don't document `@ValueList.entity`, but only `@cds.odata.valuelist`


stewsk:
@Akatuoro for https://github.tools.sap/cap/issues/issues/20571

PR #2614, opened 2026-06-03
Add concrete logger example for application.yaml

Akatuoro:
Currently we don't show any cds specific loggers in the `application.yaml` example. Users reading capire are expecting to find CAP specific information. Adding an example for some of the most used settings.


renejeglinsky:
Should we involve someone from the Java team to even include more examples?

Akatuoro:
The available loggers are already listed further down in [predefined loggers](https://cap.cloud.sap/docs/java/operating-applications/observability#predefined-loggers), so I think this is enough. Other loggers simply follow the java package + class names.

rjayasinghe (approved):

renejeglinsky (approved):

PR #2617, opened 2026-06-07
Update compiler error for redirection target in cdl.md

qmacro:
The error emitted in the context of an ambiguous redirection is different now, so I changed it in the "danger" highlight section. In order to "contextualise" this part of the error:

```text
... as redirection target for “<original-scope-prefix>.Books”
```

I also added the `using` statement to explain the scoped target reference.

### DIFF
--- a/cds/cdl.md
+++ b/cds/cdl.md
@@ -1993,10 +1993,12 @@ service AdminService {
 Auto-redirection fails if a target can't be resolved unambiguously, that is, when there is more than one projection with the same minimal 'distance' to the source. For example, compiling the following model with two projections on `my.Books` would produce this error:
 
 ::: danger
-Target "Books" is exposed in service "AdminService" by multiple projections "AdminService.ListOfBooks", "AdminService.Books" - no implicit redirection.
+Add “@cds.redirection.target” to either “AdminService.Books” or “AdminService.ListOfBooks” to select the entity as redirection target for “bookshop.Books” in this service; can't auto-redirect “AdminService.Authors:books” otherwise (in entity:“AdminService.Books”)
 :::
 
 ```cds
+using bookshop as my from '../db/schema';
+
 service AdminService {
   entity ListOfBooks as projection on my.Books;
   entity Books as projection on my.Books;


stewsk (approved):

cds/cdl.md
```diff
@@ -1993,10 +1993,12 @@ service AdminService {
 Auto-redirection fails if a target can't be resolved unambiguously, that is, when there is more than one projection with the same minimal 'distance' to the source. For example, compiling the following model with two projections on `my.Books` would produce this error:
 
 ::: danger
-Target "Books" is exposed in service "AdminService" by multiple projections "AdminService.ListOfBooks", "AdminService.Books" - no implicit redirection.
+Add “@cds.redirection.target” to either “AdminService.Books” or “AdminService.ListOfBooks” to select the entity as redirection target for “bookshop.Books” in this service; can't auto-redirect “AdminService.Authors:books” otherwise (in entity:“AdminService.Books”)
```
renejeglinsky, line 1996: In the following we show two possible options: `redirected to` and `@cds.redirection.target`. Why is the detailed note about the annotation here at the top? Shouldn't it rather be located in the section about the annotation? Or is the annotation the recommendation and `redirected to` just another option?

Anyway, here are two option wrt styling:

```suggestion
Add "@cds.redirection.target" to either "AdminService.Books" or "AdminService.ListOfBooks" to select the entity as redirection target for "bookshop.Books" in this service; can't auto-redirect "AdminService.Authors:books" otherwise (in entity: "AdminService.Books")
```
I'm not sure why we're not using code here:


```suggestion
Add `@cds.redirection.target` to either `AdminService.Books` or `AdminService.ListOfBooks` to select the entity as redirection target for `bookshop.Books` in this service; can't auto-redirect `AdminService.Authors:books` otherwise (in entity: `AdminService.Books`)
```

qmacro: I was trying to respect / reflect what was there before, i.e. the order in which things occur, and the presentation of the error message in plain text:

Image

I think either approach (complete plain text or formatted as if it were Markdown, with backticks), is fine.

On the "smart" double quotes, I've no idea how they snuck in, I hate them with a passion! :-)

stewsk: > Why is the detailed note about the annotation here at the top?

Because this is the error message you get when automatic redirection fails. In this case, using the annotation is the simplest solution.

On the various quotes: DJ's snippet exactly reflects the quotes as they are in the error message from the compiler. I think using different style here in capire would rather confuse people.

renejeglinsky: Alright, I got it now. Then it's fine.
And: The error message is much better now 👍

renejeglinsky (approved):

PR #2634, opened 2026-06-17
fix: support redirects ending with `/`

chgeo:
So that http://cap.cloud.sap/docs/about/ works again.
This one is referenced from https://me.sap.com/notes/3464370.


danjoa (approved):

renejeglinsky:
Very nice! Looking at analytics, this was one of the biggest dead ends we created and weren't able to solve until now. Thanks!

PR #2635, opened 2026-06-17
fix: drop Node.js 20 references, recommend 22/24

DanSchlachter:
Node.js 20 reached end of life in April 2026. Update the get-help guide to no longer list it as a supported LTS version and bump the Maven pom.xml example to v24.14.1.


renejeglinsky (approved):
👍

PR #2641, opened 2026-06-19
fiori: add warning to protect partial updates

johannes-vogel:


guides/uis/fiori.md
```diff
@@ -359,6 +359,9 @@ While this was always possible in CAP Java before, it's available for CAP Node.j
 > Directly updating an active entity does **not** bypass [draft locks](#draft-locks).
 > If an existing draft locks the entity, direct updates are blocked to prevent lost update situations.
 
+> [!warning] Ensure additional entry points are imlemented
```
renejeglinsky: Should we mention validation in the title?

```suggestion
> [!warning] Ensure validation for all entry points
```

guides/uis/fiori.md
```diff
@@ -359,6 +359,9 @@ While this was always possible in CAP Java before, it's available for CAP Node.j
...
+> Through this feature also partial CREATE/UPDATE requests to the root entity and its composition children have to be handled. Ensure that the validations and determinations are run in all situations, not only on the root.
```
renejeglinsky:
```suggestion
> Requests to active data also features partial CREATE/UPDATE requests to the root entity and its composition children. Ensure that the validations and determinations are run in all situations, not only on the root.
```

renejeglinsky (approved):

PR #2648, opened 2026-06-22
Node.js: Rework Generic Handlers

sjvans:


renejeglinsky:
@danjoa Would you review this already for the upcoming release?

PR #2664, opened 2026-06-24
Performance degradations in H2 `2.4`

MattSchur:


java/developing-applications/testing.md
```diff
@@ -267,6 +267,10 @@ Check out the version in our [CAP Java bookshop sample project](https://github.c
 
   H2 includes a built-in web console application, providing a user-friendly interface for database administration, query execution, and data inspection without requiring external tools. CAP Java applications configured with the H2 database expose the administration console under `http://localhost:8080/h2-console` (assuming default port `8080`).
 
+::: warning Use the latest `2.3.x` version of H2
+H2 `2.4` has a bug causing performance degradations in certain subqueries, which are used by cascade delete.
```
agoerler:
```suggestion
Avoid using the version `2.4.240` of H2. It has a bug causing performance degradations in certain subqueries, which are used by cascade delete.
```

agoerler (dismissed):

java/developing-applications/testing.md
```diff
@@ -267,6 +267,10 @@ Check out the version in our [CAP Java bookshop sample project](https://github.c
 
   H2 includes a built-in web console application, providing a user-friendly interface for database administration, query execution, and data inspection without requiring external tools. CAP Java applications configured with the H2 database expose the administration console under `http://localhost:8080/h2-console` (assuming default port `8080`).
 
+::: warning Use the latest `2.3.x` version of H2
+Avoid using the version `2.4.240` of H2. It has a bug causing performance degradations in certain subqueries, which are used by cascade delete.
+:::
```
renejeglinsky, line 272: When would we delete this note? When the bug is fixed? Or are they not planning on doing so?

agoerler: When a release with a bugfix is available.

I am a bot uneasy wether we can (legally) highlight this bug. Alternatively we could phrase a positive message and state that the max supported minor version is 2.3.x.

renejeglinsky: Is there a public bug report? Or did we just identify that during testing?
I agree to a positive formulation, but if that is a publicly known bug, I don't see an issue mentioning that.

```suggestion
::: warning Use the latest `2.3.x` version of H2
Even though there are newer versions available, we cannot recommend using them for now. Therefore we only support until version 2.3.x of H2. This recommendation is subject to change. If our test results allow to recommend newer versions we will do so.
:::
```

MattSchur: There is one that looks related: https://github.com/h2database/h2database/issues/4353

renejeglinsky: Thanks @MattSchur ! What do we do with that now?
@agoerler @MattSchur A more positive formulation or stay as it is now and add a link to the bug report?

agoerler (approved):

PR #2671, opened 2026-06-25
use concrete version number for open rewrite call

rjayasinghe:
the migration name has a version, too but we can't use the placeholder here. Thus, I will also remove the other placeholder.


agoerler (approved):

rjayasinghe:
@renejeglinsky please take over :)

renejeglinsky (approved):

PR #2673, opened 2026-06-25
Minor fixes to constraints.md

qmacro:
missing slash after scheme in URL causes the target URL to be conflated, leading to e.g. `https://cap.cloud.sap/github.com/capire/xtravels/tree/main/srv/travel-constraints.cds` and a 404.

Also fixed a typo (anoter -> another).


renejeglinsky (approved):
Thank you!

PR #2678, opened 2026-06-26
Clarification on using directive

stewsk:
Clarify that the names provided in a `using´ directive don't influence what is imported into the model.

This change was triggered by stakeholder questions.

### DIFF
--- a/cds/cdl.md
+++ b/cds/cdl.md
@@ -146,11 +146,10 @@ Within those strings, escape sequences from JavaScript, such as `\t` or `\u0020`
 
 #### The `using` Directive {#using}
 
-Using directives allows to import definitions from other CDS models. As shown in line three below, you optionally can specify local aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix.
+Using directives allows to import definitions from other CDS models. As shown in line 3 below, you optionally can specify local aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix.
 
 ::: code-group
 
-```cds [using-from.cds]
 using foo.bar.scoped.Bar from './contexts';
 using foo.bar.scoped.nested from './contexts';
 using foo.bar.scoped.nested as specified from './contexts';
@@ -172,8 +171,9 @@ entity Car : Bar { /*...*/ }
 
 > Also in the deconstructor variant of `using` shown in the previous example, specify fully qualified names.
 
-All definitions of the model proivided after `from` are imported, no matter which names are specified before `from`.
-The purpose of these names is only to make global names accessible locally.
+> [!important] Names do not restrict the import scope
+> All definitions of the model provided after `from` are imported, no matter which names are specified before `from`.
+> The purpose of these names is only to make global names accessible locally.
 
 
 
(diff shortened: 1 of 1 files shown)


cds/cdl.md
```diff
@@ -172,6 +172,8 @@ entity Car : Bar { /*...*/ }
 
 > Also in the deconstructor variant of `using` shown in the previous example, specify fully qualified names.
 
+All definitions of the model proivided after `from` are imported, no matter which names are specified before `from`.
+The purpose of these names is only to make global names accessible locally.
```
chgeo: Shouldn't we give this a note heading?

```suggestion
> [!important] Names do not restrict the import scope
> All definitions of the model provided after `from` are imported, no matter which names are specified before `from`.
> The purpose of these names is only to make global names accessible locally.
```

cds/cdl.md
chgeo: Because line 3 is mentioned above, add line numbers:
```suggestion
```cds :line-numbers [using-from.cds]
```

cds/cdl.md
```diff
@@ -146,7 +146,7 @@ Within those strings, escape sequences from JavaScript, such as `\t` or `\u0020`
 
 #### The `using` Directive {#using}
 
-Using directives allows to import definitions from other CDS models. As shown in line three below you can specify aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix. Optional: Choose a local alias.
+Using directives allows to import definitions from other CDS models. As shown in line three below, you optionally can specify local aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix.
```
chgeo:
```suggestion
Using directives allows to import definitions from other CDS models. As shown in line 3 below, you optionally can specify local aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix.
```

chgeo (commented):
Just came across this...

cds/cdl.md
renejeglinsky: Very interesting. I also was under a different assumption.

cds/cdl.md
renejeglinsky: I was tricked into thinking that "as specified from" would have a meaning. Only when I read that this was showing how to create an alias I got it. Would it be helpful, for those skipping the text, to have a different alias that does not make sense? Do you know what I mean?


```suggestion
using foo.bar.scoped.nested as animal from './contexts';
```

Would also need to change the next part then to `animal.Zoo` instead of `specified.Zoo` then

cds/cdl.md
stewsk: Note heading is fine with me

cds/cdl.md
stewsk: yes, that may indeed be helpful

renejeglinsky (dismissed):

cds/cdl.md
```diff
@@ -146,18 +146,17 @@ Within those strings, escape sequences from JavaScript, such as `\t` or `\u0020`
 
 #### The `using` Directive {#using}
 
-Using directives allows to import definitions from other CDS models. As shown in line three below you can specify aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix. Optional: Choose a local alias.
+Using directives allows to import definitions from other CDS models. As shown in line 3 below, you optionally can specify local aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix.
```
stewsk:
```suggestion
Using directives allow to import definitions from other CDS models. As shown in line 3 below, you optionally can specify local aliases to be used subsequently. You can import single definitions as well as several ones with a common namespace prefix.
```

renejeglinsky (approved):

PR #2659, opened 2026-06-23
Add further reading section for CAP application deployment automation

KoblerS:

PR #2684, opened 2026-06-26
CAP Operator (Deploy to Kyma) section updated

Pavan-SAP:
- Link to the correct landing page
- Update description

### DIFF
--- a/guides/deploy/to-kyma.md
+++ b/guides/deploy/to-kyma.md
@@ -677,4 +677,4 @@ Additionally Cloud Native Buildpacks can be easily plugged together to fulfill m
 
 ### CAP Operator
 
-For further deployment automation, the [CAP Operator](https://github.com/SAP/cap-operator) provides Kubernetes-native lifecycle management for multitenant CAP applications. It introduces Custom Resource Definitions (CRDs) to declaratively describe your application and manages tenant provisioning, upgrades, and decommissioning. This keeps operations consistent across Kyma and Gardener-managed clusters.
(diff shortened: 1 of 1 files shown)


Pavan-SAP:
@KoblerS , @renejeglinsky , @swaldmann : 
At least the landing page from my comment in: https://github.com/capire/docs/pull/2659 was not used.

IMO, this as this actually fits better while not getting into mentioning CROs. The old text was also a bit too specific to tenants.
Can you check if this text is fine and update as needed?

guides/deploy/to-kyma.md
```diff
@@ -677,4 +677,4 @@ Additionally Cloud Native Buildpacks can be easily plugged together to fulfill m
 
 ### CAP Operator
 
-For further deployment automation, the [CAP Operator](https://github.com/SAP/cap-operator) provides Kubernetes-native lifecycle management for multitenant CAP applications. It introduces Custom Resource Definitions (CRDs) to declaratively describe your application and manages tenant provisioning, upgrades, and decommissioning. This keeps operations consistent across Kyma and Gardener-managed clusters.
+The [CAP Operator]((https://sap.github.io/cap-operator/) brings enterprise-grade lifecycle automation to CAP applications. It is a Kubernetes-native Operator that declaratively deploys and manages application versions, automates tenant operations, and manages domains and networking across Kyma and Gardener-managed clusters.
```
renejeglinsky: Is CRD not a concept we want to mention? Do users not need to know about it?
In the other version I liked that consistency has been emphasized. How important do you think that is?
Other than that, I like the more concise version.

Pavan-SAP: CRD is a concept, but is being covered by mentioning Kubernetes-native Operator and declarative approach which is sort of a better fitting description for the CAP Operator IMO.

I now added consistency aspect to the 2nd sentence - is that ok? 

Let me know if you preferred the previous approach - so I can go back to having the 3rd sentence explicitly.

renejeglinsky (approved):

PR #2685, opened 2026-06-26
Update Notification details

eric-pSAP:
Update both CAP Plugins and Notifications in SAP BTP pages for updated information using the Node Notifications plug-in


maxieckert-sap (dismissed):
Content-wise looks good

plugins/index.md
```diff
@@ -313,29 +313,89 @@ Available for:
...
+
+    IncidentCreatedContext ctx = IncidentCreatedContext.create();
+    ctx.setData(data);
+    notificationService.emit(ctx);
+}
+```
+
+:::
+
+Alternatively, for Java you can use declarative `@notifications` on entities to trigger notifications automatically without writing handler code:
```
renejeglinsky, line 365: Does this as well need the `event IncidentCreated ()`?

plugins/index.md
```diff
@@ -313,29 +313,89 @@ Available for:
...
+  subtitle     : 'Created by {{customer}}',
+  groupedTitle : 'Incident Updates'
+}
+event IncidentCreated {
+  title      : String;
+  customer   : String;
+  recipients : String;
+}
+```
+
```
renejeglinsky, line 331: Would it make sense to introduce the following snippet? Maybe sth like this?


```suggestion

Then implement the notification handling:

```

plugins/index.md
eric-pSAP: No, shouldn't need it here.

renejeglinsky (approved):

PR #2690, opened 2026-06-29
Node.js: Clarify Consistent Return Values

sjvans:


node.js/app-services.md
```diff
@@ -159,27 +159,29 @@ cds.ApplicationService.handle_log_events = cds.service.impl (function(){
...
+When a write targets a single row by key (for example, `srv.update(Books, 201)` or `srv.delete(Books, '1')`) and no row matches, the handler throws a 404 error. A `where` clause that matches zero rows returns an array with `affected: 0` without throwing.
 
-::: tip Consistent results across local and remote services
-This return shape was introduced in **cds 10** to make results from local services, HCQL-proxied remote services, and database services consistent. To restore the previous behavior, set `{ "features": { "legacy_srv_results": true } }` in your project configuration.
-:::
+> [!tip] Consistent Results Across Local and Remote Services
+> This shape was introduced in cds 10 so that local services, HCQL-proxied remote services, and database services return the same thing. To restore the previous behavior, set <Config>cds.features.legacy_srv_results: true</Config>.
+
+[See the migration guide for opt-out options.](../migration/cds10#fixed-service-results){.learn-more}
```
renejeglinsky:
```suggestion
[See the migration guide for opt-out options.](../releases/migration/cds10#fixed-service-results){.learn-more}
```

BobdenOs (dismissed):

node.js/app-services.md
```diff
@@ -159,27 +159,34 @@ cds.ApplicationService.handle_log_events = cds.service.impl (function(){
...
 ```js
 const inserted = await srv.create(Books).entries({title:'Catweazle'})
-inserted[0]      // { ID: '...' }  — primary key of the inserted row
-inserted.affected  // 1
+inserted.affected            // 1
+const [row] = [...inserted]  // materializes — row holds the generated key
+inserted[0]                  // same row (materialized above)
+```
+
+For `UPDATE`, `UPSERT`, and `DELETE`, the array is reserved for rows returned by a SQL `RETURNING` clause — not yet supported, so currently always empty:
```
renejeglinsky: I'm note sure if I understand this sentence.
What is described in the first part is how it should be but it's not yet supported? Or what is not yet supported if the first part is how it works?

sjvans: we will add support for `returning`, which allows to specify which data i want back (bundle roundtrips to db). that will be the content of the array. since we don't yet support `returning`, the array is always empty. but the "shape" of the result is now established, namely an array with a property `affected: <number>`

renejeglinsky:
```suggestion
For `UPDATE`, `UPSERT`, and `DELETE`, the array is reserved for rows returned by a SQL `RETURNING` clause. But `RETURNING` is not yet supported, so the array currently is always empty:
```

node.js/app-services.md
```diff
@@ -159,27 +159,34 @@ cds.ApplicationService.handle_log_events = cds.service.impl (function(){
...
-| **UPDATE** / **UPSERT** / **DELETE** | An empty array with `.affected` set to the number of rows changed or deleted |
+| Operation             | Return value                                                                                  |
+|-----------------------|-----------------------------------------------------------------------------------------------|
+| `READ`                | Array of matching records, or a single record / `null` when read by key                       |
+| `INSERT` / `CREATE`   | Array with `.affected` (rows written); iterate to access the inserted rows' primary keys      |
+| `UPDATE` / `UPSERT`   | Array with `.affected` (rows changed); populated with rows from a `RETURNING` clause          |
+| `DELETE`              | Array with `.affected` (rows deleted); populated with rows from a `RETURNING` clause          |
 
-The `.affected` count is a property directly on the returned array:
+For `INSERT`s, the result is a lazy array: iterating it (`[...result]`, `for…of`, `JSON.stringify`) materializes the inserted rows' generated primary keys in place. Direct index access works after the first iteration.
```
renejeglinsky:
```suggestion
For `INSERT`s, the result is a lazy array: iterating it (`[...result]`, `for…of`, `JSON.stringify`) materializes the generated primary keys of the inserted rows. Direct index access works after the first iteration.
```

renejeglinsky (approved):

PR #2702, opened 2026-07-01
move project explorer docs from release notes

renejeglinsky:


chgeo (approved):
I have removed the settings screenshot (which is boring) in favor of the explorer screenshot.

PR #2704, opened 2026-07-02
Streamlining docs on cds upgrade

danjoa:
- [x] Is the former `cds upgrade` command still in place?


chgeo:
> Is the former `cds upgrade` command still in place?

Yes, as `cds upgrade-tenant`, with a compat mode if `cds upgrade <tenant>` is called (positional argument).

chgeo (approved):

renejeglinsky (approved):

PR #2708, opened 2026-07-03
Add warning about cds-services-archetype and JDK 26

mofterdinger:


java/migration.md
```diff
@@ -184,6 +184,11 @@ If your project uses Lombok, you must explicitly add its annotation processor to
 [Learn more about Maven setup with Lombok.](https://projectlombok.org/setup/maven){.learn-more}
 [Learn more about the change in the Java compiler.](https://bugs.java.com/bugdatabase/JDK-8321314/description){.learn-more}
 
+::: warning
+The `cds-services-archetype` and therefore also `cds init --java` is not working with JDK 26, because of an incompatibility of the generic `maven-archetype-plugin` with that Java version. For project creation use JDK 21 or 25 to avoid creation issues.
+Using Homebrew on MacOS to install latest Maven version, will cause also the installation of a JDK 26. Ensure that the environment variable JAVA_HOME points to a JDK 21 or 25.
```
renejeglinsky, line 189: Do we need to adapt this part in some way?
https://pages.github.tools.sap/cap/docs/get-started/#java-and-maven
For example, add a version to the command if possible?

renejeglinsky: Fir Windows that is already done:

Image

mofterdinger: > Do we need to adapt this part in some way? https://pages.github.tools.sap/cap/docs/get-started/#java-and-maven For example, add a version to the command if possible?

Regarding the homepage, SAPMachine provides different versions for installation: https://github.com/SAP/homebrew-SapMachine#released

So, maybe we should switch to the installation of `sapmachine25-jdk` instead of just `brew install sapmachine-jdk `:
```bash
brew tap sap/sapmachine
brew install sapmachine25-jdk
```

It looks like executing `brew instll sapmachine-jdk` will install SAPMachine 26: https://formulae.brew.sh/cask/sapmachine-jdk#default

renejeglinsky: done

renejeglinsky (approved):

PR #2710, opened 2026-07-03
MCP going public

danjoa:


marianfoo:
is intentional to still have the internal links in there?

PR #2727, opened 2026-07-07
Remove @data.product anno from xflights srv sample in cds-export section

qmacro:
The annotation [was removed in the source](https://github.com/capire/xflights/commit/52edcaa5a7b779ee734fb62942c8959dbd5b2e48) so perhaps this should be realigned.


swaldmann (approved):
makes sense to keep it aligned!

PR #2740, opened 2026-07-10
package managers for deployment tooling installation

swaldmann:
Similar to _Get Started_, using `homebrew` (macOS/Linux) and `winget/chocolatey` (Windows) where possible to install required tooling.


renejeglinsky:
I'm thinking if we can avoid having a mixture of package managers? Maybe it's fine and maybe I wouldn't nudge you if we'd use chocolatey throughout the document, but the mixture of winget and choco makes me worry about the user experience.
@joergmann What was the reason you used winget in the getting started guide? You also looked into choco, right?

PR #2782, opened 2026-07-29
docs: add RestTestClient integration testing example

rjayasinghe:
## What

Adds a `RestTestClient` example alongside the existing `MockMvc` example in the Java integration testing guide (`java/developing-applications/testing.md`).

## Why

Spring Framework 7.0 introduced `RestTestClient`. This documents it as an equivalent alternative for CAP Java integration tests.

## Details

- Splits the **Integration Testing** section into **Using MockMvc** (unchanged) and **Using RestTestClient**.
- The `RestTestClient` code is copied from the [CAP Java bookshop sample](https://github.com/SAP-samples/cloud-cap-samples-java/blob/main/srv/src/test/java/my/bookshop/CatalogServiceITest.java) (uses `@AutoConfigureRestTestClient` + `@Autowired RestTestClient` and the typed `.value(String.class, ...)` assertion).
- States both approaches are equivalent with no preference.
- Links the Spring [MockMvc](https://docs.spring.io/spring-framework/reference/testing/mockmvc.html) and [RestTestClient](https://docs.spring.io/spring-framework/reference/testing/resttestclient.html) reference docs.

### DIFF
--- a/java/developing-applications/testing.md
+++ b/java/developing-applications/testing.md
@@ -217,9 +217,11 @@ For a more extensive version of the previous `CatalogServiceTest` snippets, have
 
 Integration tests enable us to verify the behavior of a custom event handler execution doing a roundtrip starting at the protocol adapter layer and going through the whole CAP architecture until it reaches the service and event handler layer and then back again through the protocol adapter.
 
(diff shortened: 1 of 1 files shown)


agoerler (approved):
perfect

PR #2788, opened 2026-07-30
Extend embeddings docs

vkozyura:
Docs for https://github.com/cap-js/cds-dbs/pull/1676

### DIFF
--- a/guides/databases/vector-embeddings.md
+++ b/guides/databases/vector-embeddings.md
@@ -45,8 +45,8 @@ extend Incidents with {
 If the database calculates vector embeddings on write it automatically regenerates the embedding if the input data changes.
 :::
 
-::: info Local Testing with H2, SQLite, and PostgreSQL
-On H2, SQLite, and PostgreSQL the `CQL.vectorEmbedding` function is emulated using a hash-based algorithm to support local testing. The implementation generates deterministic 384-dimensional vectors suitable for testing, but does not capture semantic meaning. For production use, generate embeddings programmatically using real AI models.
+::: info Local Testing with SQLite and PostgreSQL
+SAP HANA supports all vector functions including the `vector_embedding` function with real AI models. PostgreSQL supports vector functions when the pgvector extension is created, but does not support the `vector_embedding` function natively. A hash-based `vector_embedding` function is provided for both SQLite and PostgreSQL to avoid issues and crashes during development, but it is strongly recommended to override this function with a reasonable custom or third-party implementation for production use.
 :::
 
 > [!warning] Java only and <Beta/>
@@ -136,6 +136,7 @@ Normalizes a vector to unit length.
 ### vector_embedding
 ```
 vector_embedding(text, text_type, model_name) → vector
+vector_embedding(text, text_type, model_name, remote_source) → vector
 ```
 Generates vector embeddings from text.
 
@@ -143,6 +144,7 @@ Generates vector embeddings from text.
 - `text` - Input text to embed
 - `text_type` - `'DOCUMENT'` (for storing content) or `'QUERY'` (for search queries)
 - `model_name` - Model identifier (database-specific)
+- `remote_source` (optional) - Remote source configuration for external embedding services (SAP HANA only)
 
 **Database Implementation:**
 - **HANA:** Uses real AI models (SAP built-in models or external remote sources)
--- a/guides/databases/vector-embeddings.md
+++ b/guides/databases/vector-embeddings.md
@@ -45,8 +45,8 @@ extend Incidents with {
 If the database calculates vector embeddings on write it automatically regenerates the embedding if the input data changes.
 :::
 
-::: info Local Testing with H2 and SQLite
-On H2 and SQLite the `CQL.vectorEmbedding` function is emulated to support local testing.
+::: info Local Testing with H2, SQLite, and PostgreSQL
+On H2, SQLite, and PostgreSQL the `CQL.vectorEmbedding` function is emulated using a hash-based algorithm to support local testing. The implementation generates deterministic 384-dimensional vectors suitable for testing, but does not capture semantic meaning. For production use, generate embeddings programmatically using real AI models.
 :::
 
 > [!warning] Java only and <Beta/>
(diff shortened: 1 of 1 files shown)


guides/databases/vector-embeddings.md
```diff
@@ -46,8 +46,8 @@ extend Incidents with {
 If the database calculates vector embeddings on write it automatically regenerates the embedding if the input data changes.
 :::
 
-::: info Local Testing with H2 and SQLite
-On H2 and SQLite the `CQL.vectorEmbedding` function is emulated to support local testing.
+::: info Local Testing with SQLite and PostgreSQL
+SAP HANA supports all vector functions including the `vector_embedding` function with real AI models. PostgreSQL supports vector functions when the pgvector extension is created, but does not support the `vector_embedding` function natively. A hash-based `vector_embedding` function is provided for both SQLite and PostgreSQL to avoid issues and crashes during development, but it is strongly recommended to override this function with a reasonable custom or third-party implementation for production use.
```
BobdenOs: Where did the `h2` reference go ?
What is the purpose of mentioning that SAP HANA works in an `info` section ?
This comment mentions that for Postgres a hash version is delivered, but this is not currently the case.

We should probably add the information that for Postgres production the customers have to define their own `vector_embedding` function.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+```
+l2distance(vector1, vector2) → number
+```
+Calculates Euclidean distance between two vectors. Returns 0 for identical vectors.
+
+### l2normalize
+```
+l2normalize(vector) → vector
+```
+Normalizes a vector to unit length.
```
BobdenOs: `unit length` being `-1` to `1` ?

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+
+**Parameters:**
+- `text` - Input text to embed
+- `text_type` - `'DOCUMENT'` (for storing content) or `'QUERY'` (for search queries)
+- `model_name` - Model identifier (database-specific)
+- `remote_source` (optional) - Remote source configuration for external embedding services (SAP HANA only)
+
+**Database Implementation:**
+- **HANA:** Uses real AI models (SAP built-in models or external remote sources)
+- **SQLite & PostgreSQL:** Hash-based deterministic implementation for testing. Can be overridden by application developers to use external embedding services.
```
BobdenOs: Postgres function not currently automatically deployed.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+
+## Database-Specific Considerations
+
+### SQLite
+- Vector functions implemented as JavaScript UDFs
```
BobdenOs: I don't think we should document this internal details.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+- Vectors stored as JSON strings
+- No external dependencies required
```
BobdenOs: Sounds a bit weird when the future state is not mentioned.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+- `vector_embedding()` uses hash-based algorithm for testing. Override the JavaScript function to use real embedding services in production.
```
BobdenOs: `@cap-js/sqlite` is not meant to be used for production.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+
+### PostgreSQL
+- Requires creating the [pgvector extension](https://github.com/pgvector/pgvector) in the database:
```
BobdenOs: not just `creating` it is also required that the Postgres instance that they use already has the plugin binaries included.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+  ```sql
+  CREATE EXTENSION IF NOT EXISTS vector;
+  ```
+- Vectors stored in native `vector` type
+- `vector_embedding()` SQL function uses hash-based algorithm for testing. Override the SQL function to use real embedding services in production.
+- For Node.js, optional `pgvector` npm package for type registration: `npm install pgvector`
```
BobdenOs: Are the `vector` types optional for the client ? I guess if people run queries like:

```sql
SELECT * FROM Books WHERE cosine_similarity(description, vector_embedding(?, 'QUERY')) > 0.8
```

Then the `?` will be a simple string. The client would not actually need to have the `vector` type support unless they explicitly query for the `vector` type.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
...
+
+### SAP HANA
+- Native vector engine with built-in support
+- Type mapping: `cds.Vector` → `REAL_VECTOR`
+- `vector_embedding()` supports built-in SAP models:
+  - `SAP_NEB.20240715` (German, English, Spanish, French, Portuguese - 768 dimensions)
+  - `SAP_GXY.20250407` (All above + Italian, Japanese, Chinese - 768 dimensions)
+- Can use external remote sources (Azure OpenAI, SAP AI Core) as optional 4th parameter
```
BobdenOs: I think it would be better to exclude the specifics. So that the official SAP HANA help docs can be used as their source of truth directly instead. Otherwise we have to somehow know what models HANA has released in all possible SAP HANA versions.

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,70 @@ let similarIncidents = await SELECT.from('Incidents')
 ```
 :::
 
+## Vector Functions
```
BobdenOs, line 115: We should consider simply referencing to the official SAP HANA documentation for these functions. As we are required to provide an equivalent implementation of those functions.

guides/databases/vector-embeddings.md
vkozyura: changed. my hope was that compiler will deliver hash function soon.

guides/databases/vector-embeddings.md
vkozyura: As we are providing only subset of HANA functions, I list them here, but for more details i added link to HANA page.

guides/databases/vector-embeddings.md
vkozyura: changed. again, I hoped for soon compiler change

guides/databases/vector-embeddings.md
vkozyura: removed sqlite section from here

guides/databases/vector-embeddings.md
vkozyura: explained

guides/databases/vector-embeddings.md
vkozyura: removed

guides/databases/vector-embeddings.md
vkozyura: removed

guides/databases/vector-embeddings.md
vkozyura: changed

guides/databases/vector-embeddings.md
vkozyura: explained

guides/databases/vector-embeddings.md
vkozyura: specifics removed

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,64 @@ let similarIncidents = await SELECT.from('Incidents')
...
+
+## Database-Specific Considerations
+
+### PostgreSQL
+- Requires that the [pgvector extension](https://github.com/pgvector/pgvector) is installed on your PostgreSQL instance. Then create the extension in your database:
+  ```sql
+  CREATE EXTENSION IF NOT EXISTS vector;
+  ```
+- Vectors stored in native `vector` type
+- `vector_embedding()` SQL function uses hash-based algorithm for testing. Override the SQL function to use real embedding services in production.
```
BobdenOs, line 165: This still says "uses hash based".

guides/databases/vector-embeddings.md
```diff
@@ -112,3 +112,64 @@ let similarIncidents = await SELECT.from('Incidents')
...
+- For Node.js, the `pgvector` npm package is only required when you need to read vector columns from query results. If you only pass vectors as parameters (for example, in WHERE clauses or function arguments), the package is not needed: `npm install pgvector`
```
BobdenOs, line 166: This example is not fully accurate and should be double checked. If the vector embedding function is used to create the ending on the fly inside the database the plug-in is not required, but if the client is sending the embedding through a placeholder then the plug-in is still required. The current way it is written is the second example which does require the plug-in.

guides/databases/vector-embeddings.md
BobdenOs: We could still consider simply replacing all the function descriptions to the HANA docs for each function not just linking to the vector functions overview page. I'm mostly trying to prevent duplicating documentation as it is more likely to go out of sync with reality.

vkozyura: simplified and pointed to HANA docs

guides/databases/vector-embeddings.md
vkozyura: fixed

guides/databases/vector-embeddings.md
vkozyura: fixed. hope it is correct now

BobdenOs (approved):

vkozyura:
Hi @smahati, from a content perspective, the docs have been reviewed and are ready. Please review them from your side.

PR #2790, opened 2026-07-31
docs: document cds.drafts.enforceReadonly and the @readonly bypass hint

rjayasinghe:
## What

Adds prose documentation for the `cds.drafts.enforceReadonly` property in **`java/fiori-drafts.md`** (new *Read-Only Fields in Drafts* section):

- Explains that `@readonly` / `@Core.Computed` are enforced only at activation by default, and that setting `cds.drafts.enforceReadonly: true` enforces them on the draft itself during `DRAFT_NEW` / `DRAFT_PATCH` (so read-only fields can't be changed via OData `PATCH` on the draft).
- Documents the statement hint `.hint("@readonly", false)` to intentionally write read-only / computed fields from trusted custom code (e.g. determinations), with a warning that managed (`@cds.on.insert`/`@cds.on.update`) and `@Core.Immutable` fields remain enforced regardless.

## Why

The property already appears in the generated properties table but had no prose explaining the custom-code implication or the bypass hint, which surfaced in a developer question.

## Verification

- `npm run docs:build` passes with no dead-link errors; the property cross-link anchor `#cds-drafts-enforcereadonly` matches the generated id.
- Hint and behavior verified against the CAP Java runtime source (`ReadOnlyHandler`).


smahati:
Hey @mofterdinger Could you provide your feedback for this PR? Thank you :)

java/fiori-drafts.md
```diff
@@ -116,6 +116,32 @@ public void validateOrderItem(CdsCreateEventContext context, OrderItems orderIte
 During activation the draft data is deleted from the database. This happens before the active entity is created or updated within the same transaction.
 In case the create or update operation raises an error, the transaction is rolled back and the draft data is restored.
 
+## Read-Only Fields in Drafts { #readonly-in-drafts }
+
+By default, `@readonly` and `@Core.Computed` fields are only enforced when a draft is activated, that means during the `CREATE` or `UPDATE` event on the active entity. Until then, such fields can still be changed on the draft, for example through an OData `PATCH` request.
+
+To enforce these annotations on the draft already, during the `DRAFT_NEW` and `DRAFT_PATCH` events, set the [`cds.drafts.enforceReadonly`](./developing-applications/properties#cds-drafts-enforcereadonly) property (default `false`):
```
smahati, line 123:
```suggestion
To enforce these annotations on the draft even before activation, during the `DRAFT_NEW` and `DRAFT_PATCH` events, set the [`cds.drafts.enforceReadonly`](./developing-applications/properties#cds-drafts-enforcereadonly) property (default `false`):
```

renejeglinsky: Maybe it doesn't even need the "even" :)

```suggestion
To enforce these annotations on the draft before activation, during the `DRAFT_NEW` and `DRAFT_PATCH` events, set the [`cds.drafts.enforceReadonly`](./developing-applications/properties#cds-drafts-enforcereadonly) property (default `false`):
```

mofterdinger:
Hi, sorry, I can't provide any helpful review feedback, I wasn't involved in implementing this feature. Will remove my assignment.
Thanks,
Markus

rjayasinghe:
I guess @beckermarc is the better candidate. :)

java/fiori-drafts.md
```diff
@@ -116,6 +116,32 @@ public void validateOrderItem(CdsCreateEventContext context, OrderItems orderIte
 During activation the draft data is deleted from the database. This happens before the active entity is created or updated within the same transaction.
 In case the create or update operation raises an error, the transaction is rolled back and the draft data is restored.
 
+## Read-Only Fields in Drafts { #readonly-in-drafts }
```
beckermarc, line 119: We should also explain the effects for readonly elements calculated in code (determinations). In the default scenario determinations need to happen during draft activation on CREATE or UPDATE. They can optionally happen at draft edit time as well (to show the determined field on the UI already in that mode). If the determination is not fully deterministic (e.g. creating a random UUID) there is no guarantee that values created in draft mode and during activation are the same.

This can be solved by setting the `enforceReadonly` flag to `false` as well, as on that mode `@readonly` is no longer evaluated during activation, thus elements calculated in draft mode are taken over with the exact same value.

java/fiori-drafts.md
```diff
@@ -116,6 +116,32 @@ public void validateOrderItem(CdsCreateEventContext context, OrderItems orderIte
...
+
+```yaml
+cds.drafts.enforceReadonly: true
+```
+
+With this property enabled, values sent for `@readonly` or `@Core.Computed` fields are ignored when a draft is created or patched, so these fields can no longer be modified through OData requests on the draft.
+
+### Writing Read-Only Fields from Custom Code { #readonly-hint }
+
+When `cds.drafts.enforceReadonly` is enabled, the read-only enforcement also applies to `Update` or `Insert` statements that your own event handlers run against the draft entity, for example in a determination. As a result, values for `@readonly` and `@Core.Computed` fields are removed from these statements as well.
```
beckermarc, line 133: We should mention here explicitly that this applies to statements run via the DraftService.

java/fiori-drafts.md
rjayasinghe: Good point — added a *Determined Read-Only Fields* subsection covering this. With the default (`enforceReadonly: false`) a non-deterministic determination (e.g. generating a random UUID) can produce different values at draft-edit time and at activation, so the draft shows one value and the active entity persists another. Setting `true` avoids this: read-only fields are no longer evaluated at activation, so the draft value is carried over verbatim. Verified against `ReadOnlyHandler` / `DraftActionsHandler.defaultSave` and the `testReadonlyOnDraft` unit test.

java/fiori-drafts.md
rjayasinghe: Done — also clarified the scope of the statement-level enforcement. It's actually registered on `ApplicationService` for all INSERT/UPDATE/UPSERT events (not exclusively `DraftService` statements), so any handler that writes to the draft entity through the application service is affected. The meaningful boundary is vs. the `PersistenceService`: statements run directly against the persistence service bypass the application service and are not stripped. Updated the paragraph to say 'through the application service' to make that accurate. Let me know if you'd prefer different wording.

beckermarc (dismissed):

rjayasinghe:
@beckermarc I applied one change from Rene, now I need a new approval. 🫣

beckermarc (approved):

PR #2846, opened 2026-08-28
Beta: HANA Cloud serverless

swaldmann:


ecklie (commented):
Does this work for external users?

ecklie (approved):

PR #2851, opened 2026-09-02
Update notifications plugin

lisajulia:
- Update link to notifications plugin: add alpha label for Java
- Add optional database storage for Java


plugins/index.md
maxieckert-sap: Does this still only work for Java? @eric-pSAP

eric-pSAP: @maxieckert-sap currently this is in review for Node

lisajulia: Then I'd say we leave this as it is and update the documentation once Node is ready.

eric-pSAP (approved):

renejeglinsky (approved):

PR #2886, opened 2026-09-14
Allow links to service bindings and Joule guides

chgeo:
Set pages to released as
- node.js/remote-services want to link to service bindings
- AI guides want to mention Joule page, even though it's under construction


guides/integration/_menu.md
```diff
@@ -1,6 +1,6 @@
 
 # [ CAP-level Service Integration ](calesi.md)
-# [ CAP-level Data Federation ](data-federation.md) 
-# [ Inner Loop Development ](inner-loops.md) 
+# [ CAP-level Data Federation ](data-federation.md)
+# [ Inner Loop Development ](inner-loops.md)
 # [ Reuse & Compose ](reuse-and-compose.md)
-# [ Service Bindings ](service-bindings.md) <!-- UNRELEASED -->
+# [ Service Bindings ](service-bindings.md)
```
renejeglinsky: This has quite some empty sections and back then we decided to not release it. If the link is the only reason, I'd vote for fixing the link and not release this here. If this will be filled/worked on, I'm fine with releasing it.

chgeo: > If this will be filled/worked on, I'm fine with releasing it.

It least in https://github.com/capire/docs/commit/02038d3cd12a21a87b2db2bdcb12be4ba6e2a4be#diff-bd2c13cdb01ff8572eed7b9985ef1368bb4abe124e02e86a91e3402649c9e603 this was recently worked on.

@schwma The issue here is that your links to the service-bindings.md fail in external builds as that page is not yet released.  Can you tell us, or e.g. @sjvans, if the page's content is good enough for an external release?
Alternatively, we'd need to remove the links to the page again, making your contributions there not appear externally.

sjvans: i agree with fixing the links. i will move some content from service bindings into node.js' remote service when i find the time.

chgeo: Ok, then I'll go ahead and comment out the links, so that we can get back to green.

renejeglinsky (approved):

PR #2890, opened 2026-09-15
Use mutual TLS route as webhookUrl for Event Broker binding

robinkuck:
I was trying to set up this configuration today and it failed. The incoming requests from the Event Broker to the webhookUrl failed with HTTP status 401 and following error message: `c.s.c.s.t.validation.ValidationResults   : Token signature can not be validated because JWKS could not be fetched: Client certificate for proof token validation could not be read from 'x-forwarded-client-cert' header.`

After some analysis of internal resources, I found out that the .cert. route needs to be chosen as host for the webhookUrl because it's the only endpoint that enables mutual TLS, which is the authentication mechanism the Event Broker uses (`authentication-type: X509_IAS` in mta.yaml) to prove its identity to the webhook. The default host only does server-side TLS + token auth, so certificate-based inbound calls wouldn't authenticate there.


renejeglinsky:
Thanks @robinkuck for your contribution! I'll hand it over to our experts for a review.

@sjvans Could you review the change or assign someone? Thanks.

sjvans (approved):

renejeglinsky (approved):

PR #2902, opened 2026-09-18
fix: hide left sidebar scrollbar on long menus

DanSchlachter:
## Problem

When the left navigation menu content is longer than the viewport, a grey vertical scrollbar appears between the sidebar and the main doc area. It looks bad.
Screenshot 2026-09-18 at 10 02 23

## Change

Hide the scrollbar on `.VPSidebar` across all browsers (Firefox, legacy Edge/IE, Chrome/Safari) while keeping the sidebar scrollable via mouse wheel / trackpad.


chgeo:
Why do you think the scrollbar is bad here?  It only appears when scrolling, so exactly when it's needed.
From an accessbility point I am not sure no scrollbars are desired.  Check what the bast practises say here.

DanSchlachter:
i think it looks really out of place there somehow, as it goes through the header area which isn't scrollable and is very dominant compared to the minimalistic style (especially on the internal docs page with changed layout).
There is also no scrollbar on the right page outline and I think this looks more modern.
But maybe that can be improved without removing it completely - especially if needed for accessibility

chgeo:
The scrollbar is standard even in vitepress.dev:

image

> it looks really out of place there somehow, as it goes through the header area

True for the header area, this is different compared to Vitepress.
So what about fixing this issue first?

chgeo:
> True for the header area, this is different compared to Vitepress.
> So what about fixing this issue first?

I have fixed this with https://github.com/capire/docs/pull/2910

chgeo:
The rest we should discuss in the UA sync.

chgeo:
Scrollbar behavior on macOS demystified:
 
> That's macOS's "Automatically based on mouse or trackpad" setting in action — even without changing System Settings manually. macOS detects which input device you're using and switches scrollbar mode on the fly:
> - Trackpad → overlay scrollbars (fade out when idle)
> - Mouse connected → persistent scrollbars (always visible)
> 
> So if you plug/unplug a mouse, or use Bluetooth that reconnects, the behavior switches mid-session without any explicit settings change. Chrome follows the OS setting for this — it's not a browser or CSS issue.
 
Can I change the Mouse settings to fade out?

>  Yes — System Settings → Mouse → uncheck "Show scroll bars: Always" isn't quite the right path; it's actually in:
System Settings → Appearance → Show scroll bars → When scrolling
That applies to both trackpad and mouse, so scrollbars will always fade out when idle regardless of input device.

PR #2910, opened 2026-09-21
Restore standard sidebar, fixing scrollbar

chgeo:
Restore the default Vitepress style that leaves room for the scrollbar.

Also, the default color is restored, so that the vertical separation between main and sidebar is there. This looks kind of broken atm.

### Now

image

### Previously

image

##### Weak visiual separation, cut-off divider line
image


DanSchlachter:
i still find it strange that the scrollbar starts at the header, as the header is sticky (doesn't move when you scroll down). 
With the grey background and space that is less obvious and looks more natural

chgeo:
W/ the white background, we'd get a strange gap in the navbar for the scrollbar:
image

So either we keep the current state of the overlapping separator/scrollbar, or change to the gray sidebar from this PR.

swaldmann (approved):

